PR #3, opened 2015-10-02
Add foundation for ListCommentsUserCase

JeroenDeDauw:


JeroenDeDauw:
@wmde-manicki @KaiNissen I just added a Comment and CommentRpeository now without knowing if that really makes sense or not. It's certainly a bit silly at the moment as this Comment class is the same as the one in the list. However it now does demonstrate the general pattern, and the stuff can be refactored as we get to understand better what sort of repository makes sense while implementing the other use cases that touch it.

src/Domain/CommentRepository.php
```diff
@@ -0,0 +1,14 @@
...
+namespace WMDE\Fundraising\Frontend\Domain;
+
+interface CommentRepository {
+
+	/**
+	 * @param int $limit
+	 *
+	 * @return Comment[]
+	 */
+	public function getComments( $limit );
```
gbirke: Should't there be an optional param for the start offset so we can get comment paging?

JeroenDeDauw: Eventually there might be. However at this point the use case does not provide pagination yet, so I've not (IMO prematurely) added this to the service is uses either.

PR #9, opened 2016-01-05
Use return type hints and scalar type hints

JeroenDeDauw:


gbirke:
So ... the use cases only return use case data types, presentation is left to the "framework layer" above? I'd like to discuss the pros and cons.

JeroenDeDauw:
Well sure, though that's not really relevant for this PR...

PR #11, opened 2016-01-05
Add Silex and routing+presentation for validate email 

JeroenDeDauw:
Includes commit of #10 which should be merged first


src/FFFactory.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace WMDE\Fundraising\Frontend;
+
+use WMDE\Fundraising\Frontend\UseCases\ValidateEmail\ValidateEmailUseCase;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class FFFactory {
```
JeroenDeDauw: pfft

:)

JeroenDeDauw:
@KaiNissen addressed your comments... think it's better if you comment on the PR itself rather than the individual commits, which makes it easier to track the comments

PR #13, opened 2016-01-07
Add handling for route not found

JeroenDeDauw:
/kittens

```json
{
    "message": "No route found for \"GET /kittens\"",
    "code": 404
}
```

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -20,4 +20,14 @@ $app->after( function( Request $request, Response $response ) {
 	return $response;
 } );
 
+$app->error( function ( \Exception $e, $code ) {
+	return new JsonResponse(
+		[
+			'message' => $e->getMessage(),
+			'code' => $code
+		],
+		$code
+	);
+} );
+
 return require __DIR__ . '/routes.php';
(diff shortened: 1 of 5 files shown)


gbirke:
Ok, as a first version I'd merge this, but then we'll have to create a ticket to remind us to do an HTML version. Or will you do the HTML version as part of https://phabricator.wikimedia.org/T123037 ?

JeroenDeDauw:
I am creating an HTML version for that use case at the very moment. However that so far will only apply to the not finding of the specified page, and not affect the global 404 handling of the app. As far as I can see the old one has no global handling that gives you HTML either.

* https://spenden.wikimedia.de/spenden/list.kittens 404 not found
* https://spenden.wikimedia.de/spenden/kittens.php File not found
* https://spenden.wikimedia.de/spenden/kittens/cats 404 not found
* https://spenden.wikimedia.de/weeee 404 not found

We can of course change that...

gbirke:
In our new setup all requests will be routed to the application if no file exists. So the application should be able to display a nice 404-page. But for the time being, a JSON error message is an improvement over a raw nginx 404-page.

PR #16, opened 2016-01-07
Work on display page usecase

JeroenDeDauw:
The real work in DisplayPageUseCase::getPageContent still needs to happen. The comments there indicate the relevant steps, as far as I can tell, that happen in the old code. I don't just want to copy WikiRip, as it seems to be doing things not needed here, and is also not structured nicely. Instead I want to copy the relevant parts of the relevant methods and reorganizing them. This is what I have in mind ATM:

* normalization - method in the UC itself
* white and blacklisting in a new PageFlufinessDetector class that returns a yay or nay for a page name (see #17)
* fetching of the content via existing PageRetriever stuff (action always seems to be "render", so we can probably drop this parameter and add support for its feature again when we need it (probably in a way different than using such a param))
* old getProcessedContent behaviour via a new service (to not have it in the UC)
* optional outputting of debug into via new service (maybe the same as for getProcessedContent) (maybe we don't need this right away)

(To clarify, these new services/classes would mainly contain code copied from the old codebase)

Thoughts? In particular, did I miss anything that needs to be done in this UC, or are any of these things not actually needed?

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -3,7 +3,7 @@
 /**
  * These variables need to be in scope when this file is included:
  *
- * @var \WMDE\Fundraising\Frontend\FFFactory $ffFactory
+ * @var \WMDE\Fundraising\Frontend\FunFunFactory $ffFactory
  */
 
 use Symfony\Component\HttpFoundation\JsonResponse;
--- a/app/routes.php
+++ b/app/routes.php
@@ -4,7 +4,7 @@
  * These variables need to be in scope when this file is included:
  *
  * @var \Silex\Application $app
- * @var \WMDE\Fundraising\Frontend\FFFactory $ffFactory
+ * @var \WMDE\Fundraising\Frontend\FunFunFactory $ffFactory
  */
 
 use Silex\Application;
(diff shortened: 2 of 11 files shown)


JeroenDeDauw:
Feel free to merge or add your own stuff on top (if it's part of getting this UC done that is).

JeroenDeDauw:
This also kinda depends on https://github.com/JeroenDeDauw/FileFetcher/pull/3

JeroenDeDauw:
Is it correct that the WikiRip stuff is only used by the page displaying stuff?

wmde-manicki:
WikiRip is used also to rip the content of emails (after donation, confirming the subscription, when your donation has been unlocked etc).
And, to make things worse, you might actually would like to know that the same code is also used in https://github.com/wmde/Wikipedia.de, at least to display banners that are being ripped from some other wiki. I suspect that other WikiRip is a bit different than the one from old fundraising app (meaning it is old version before @WMDE-Fisch did some changes to it) but I guess it used to simple copy.

JeroenDeDauw:
Thanks for explaining @wmde-manicki

JeroenDeDauw:
Would be good if this gets merged soon, cause else it'll become difficult to hold the changes this thing makes to common stuff into account

KaiNissen:
err, are we sure we want to call the factory class this way? or did someone just successfully sneak more fun into the code base?

JeroenDeDauw:
@KaiNissen was it not you that suggested this name? :D

PR #17, opened 2016-01-07
[DNM] Add StringValueGuard

JeroenDeDauw:
Originally created a PageNameGuard though then remebered how functional programming
advocates bash typical OOP programs for being overly specific and thus requiring to
reinvent the wheel the whole time. In this case a PageNameGuard would indeed be
overly specific as its logic would not be specific to page names at all.

For page name whitelisting and blacklisting in #16

### DIFF
--- a/src/StringValueGuard.php
+++ b/src/StringValueGuard.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class StringValueGuard {
+
+	private $whitelist = [];
+	private $blacklist = [];
+
+	public function isAllowed( string $value ): bool {
+		if ( $this->whitelist !== [] && !in_array( $value, $this->whitelist ) ) {
+			return false;
+		}
+
+		return !in_array( $value, $this->blacklist );
+	}
+
+	/**
+	 * @param string[] $strings
+	 */
+	public function setWhitelist( array $strings ) {
+		$this->assertAreStrings( $strings );
+		$this->whitelist = $strings;
+	}
+
+	/**
+	 * @param string[] $strings
+	 */
+	public function setBlacklist( array $strings ) {
+		$this->assertAreStrings( $strings );
+		$this->blacklist = $strings;
+	}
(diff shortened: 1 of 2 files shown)


gbirke:
I found this class very confusing, because I expected it to check against white- and blacklist and allow all values that are either on the whitelist or not in the blacklist. Instead, if a whitelist is given, all other values are disallowed. 

To make it more clear what the class does, maybe we could get rid of `setWhitelist` and `setBlacklist` and use two static factory functions `createWhitelist` and `createBlacklist`?

JeroenDeDauw:
I can't easily check now, but did I not retain the behavior that is currently in WikiRip? The goal was to have the same behavior.

gbirke:
Yes, you did retain the behavior in current wiki rip and we can leave it like this. I'm just suggesting another public interface that is less confusing.

If we decide to keep this class - see my (updated) last comment.

wmde-manicki:
As I've seen this discusson in github's notification: it is true that this has not been used but maybe it would be worth keeping it there?
I think there is one potential usage for that: At the end of campaign it turned out that there are some outdated CMS wiki pages (at least there has been one) being linked from outside, which is not intended. The particular page that fundraising people found was this thing: `https://spenden.wikimedia.de/spenden/?rip=Fragen_und_Antworten` (was accessed from outside world through `https://spenden.wikimedia.de/spenden/Fragen_und_Antworten`)

CMS wiki pages are normally not removed from the wiki, and there are probably more reasons to keep this behaviour than I could think of. So that particular page needed to be blocked from being displayed.
It is now "disabled" (meaning it redirects to the form start page) using some simple `nginx` redirect statement (similarly to thing that changes `spenden/Nyan` stuff to `spenden/rip?Nyan`). This could be probably achieved by using the black list but I wasn't aware of it existing :)

And I think in this particular case, it would actually make more sense to only allow certain pages being ripped ("stand-alone" membership forms etc) instead of blacklisting all obsolete stuff from CMS wiki, at least on production instance.
I wanted to create a general ticket about suggesting some whitelist in nginx config in late December but haven't done this yet as it was unclear where should it go (old Fun fun repo, new Fun Fun repo, or maybe the infrastructure?). Having such feature here would solve the problem, and it would do it far more elegant way than hard-coding it in web server config.

PR #18, opened 2016-01-09
Have display page UC use PageRetriever and PageContentModifier

JeroenDeDauw:
PageRetriever and implementations copied from old repo and somewhat
modified. Currently unclear to me when which implementation should
be used, and which ones we actually need. Did not move over the
ApiBasedPageRetriever yet, as it has additional dependencies.

Introduced PageContentModifier class, which holds modified
code from the old WikiRip::getProcessedContent.

For both of these I dropped the action parameter, as it appears
to always be "render" for this UC. We can re-add this later if
needed, probably in a different manner.


src/UseCases/DisplayPage/PageContentModifier.php
```diff
@@ -0,0 +1,51 @@
...
+ * @licence GNU GPL v2+
+ */
+class PageContentModifier {
+
+	private $logger;
+	private $substitutions;
+
+	public function __construct( LoggerInterface $logger, array $substitutions = [] ) {
+		$this->logger = $logger;
+		$this->substitutions = $substitutions;
```
gbirke: Please add
`$this->page_cache_list = [];`

src/UseCases/DisplayPage/PageContentModifier.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Psr\Log\LoggerInterface;
+
+/**
+ * @licence GNU GPL v2+
+ */
+class PageContentModifier {
+
+	private $logger;
+	private $substitutions;
```
gbirke: Please add
`private $page_cache_list;`

src/UseCases/DisplayPage/PageContentModifier.php
```diff
@@ -0,0 +1,51 @@
...
+
+	public function getProcessedContent( string $content, string $pageName ): string {
+		# full HTML document usually indicates an error (e.g. access denied)
+		if ( preg_match( '/^<!DOCTYPE html/', $content ) ) {
+			$this->logger->debug( __METHOD__ . ': fail, got error page', [ $content ] );
+			return '';
+		}
+
+		# NOTE: don't strip comments, they may contain javascript, etc!
+		# $content = preg_replace('/<!--.*?-->/s', '', $content);
```
gbirke: This is probably copied over from the original code. Should't we delete these two lines?

KaiNissen:
The wiki it is supposed to retrieve content from is a closed one, we actually need the API based page retriever to be able to authenticate.

src/UseCases/DisplayPage/PageContentModifier.php
JeroenDeDauw: They are copied yes. I've not really looked at them yet and do not know what their deal is.

src/UseCases/DisplayPage/PageContentModifier.php
JeroenDeDauw: Not done yet as its not clear if we need this here. There is a TODO about it further down. The empty check does not appear to cause a notice even on strict mode. Given that, do you still think the field should be defined and initialized already?

JeroenDeDauw:
> The wiki it is supposed to retrieve content from is a closed one, we actually need the API based page retriever to be able to authenticate.

What are the other page retrievers for then?

src/UseCases/DisplayPage/PageContentModifier.php
JeroenDeDauw: Actually this thing is of type `FreshCacheList`, not `array`. I've added a commit on top of this PR commenting the code out.

KaiNissen:
When we introduced the page retrievers we had [Extension:IPAuth](https://www.mediawiki.org/wiki/Extension:IPAuth) running, but we later removed that and only use API authentication now to be able to fetch content from our dev setups.

JeroenDeDauw:
> When we introduced the page retrievers we had Extension:IPAuth running, but we later removed that and only use API authentication now to be able to fetch content from our dev setups.

Ok, I'll migrate the API one then. Is there still a purpose to the other derivatives? For instance usage in testing? If not, we can just remove them.

JeroenDeDauw:
#43

PR #19, opened 2016-01-09
Add persistance initialization to system test setup

JeroenDeDauw:
@gbirke you wanted to start with some UC that interacts with the db, so will presumably need this


tests/TestEnvironment.php
```diff
@@ -13,6 +13,8 @@ class TestEnvironment {
 	public static function newInstance() {
 		$instance = new self();
 
+		$instance->factory->newInstaller()->install();
```
JeroenDeDauw: @KaiNissen here is the usage of `$instance` :)

PR #20, opened 2016-01-11
Make application read config from json files

JeroenDeDauw:
https://phabricator.wikimedia.org/T123065


phpcs.xml
```diff
@@ -52,7 +52,6 @@
     <rule ref="Generic.PHP.ForbiddenFunctions" />
     <rule ref="Generic.PHP.LowerCaseConstant" />
     <rule ref="Generic.PHP.LowerCaseKeyword" />
-    <rule ref="Generic.PHP.NoSilencedErrors" />
```
gbirke: I'm a dogmatic "Never use the error suppression operator of PHP. Ever." person. Please get back to me on this, I need some covincing before I'm comfortable with this.

src/ConfigReader.php
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		return array_replace_recursive(
+			$this->getFileConfig( $this->baseConfigFilePath ),
+			$this->getFileConfig( $this->instanceConfigFilePath )
+		);
+	}
+
+	private function getFileConfig( string $filePath ): array {
+		$config = @json_decode( $this->getFileContents( $filePath ), true );
```
gbirke: Are you sure you need error suppression here? I've tested several cases of invalid JSON and never got any message.

JeroenDeDauw: Oh right... I was thinking about file_get_contents here...

Feel free to add a commit that removes this. I'll get to this during the evening if you have not already.

src/ConfigReader.php
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	private function getFileConfig( string $filePath ): array {
+		$config = json_decode( $this->getFileContents( $filePath ), true );
+
+		if ( is_array( $config ) ) {
+			return $config;
+		}
+
+		throw new RuntimeException( 'Cannot read config file at path "' . $filePath . '"' );
```
gbirke: This can be confusing, since an empty file with the correct permissions will still generate "Cannot read config file". Maybe change to:
```php
throw new RuntimeException( 'No valid config data found in config file at path "' . $filePath . '"' );
```

JeroenDeDauw: Good point, I've amended the change now

JeroenDeDauw:
Resubmitted against master as #22. There does not appear to be a way to change the target of a PR after its been created or to close the PR and create a new one for the same branch... that kinda sucks

PR #32, opened 2016-01-11
Add dev web entry point

JeroenDeDauw:
Includes addition of profiler (see README)

PR #34, opened 2016-01-12
Dev entry point

gbirke:
Please merge again, previous PR was for the wrong branch


JeroenDeDauw:
It'd be nice if you could mention the other PR here like `resubmission of #32`, so a backlink will show up there, informing people looking at that that this exists

PR #35, opened 2016-01-12
Do not use sudo in travis before script

wmde-manicki:
As I got bunch of notifications about adding the bank data library etc, I had a look and noticed that `sudo` being used while build the lib.
As Travis installs konto check extension in `phpenv` subdir which is in build user's home directory, `sudo` is not needed for it. Actually I might have noticed that earlier in the old repo and get rid of this redundant sudo there as well.
I reckon it is good not to use sudo, so the new code wouldn't be bound to the old Travis infrastructure where sudo is available (so you could use composer cache, for instance).
I am submitting the change as a PR into bank branch but you might as well prefer to simply amend existing commits there or so.


KaiNissen:
Thank you for fixing this. I rebased my PR to resolve a conflict and squashed some other commits, so I just cherry-picked f5e8246 and close this one.

PR #41, opened 2016-01-13
Subscription validation

gbirke:
Make validation of scalar values and instances more generic and reusable. 
Move validation into its own namespace.


gbirke:
The `RequestValidator::validate` method is very generic and should probably extracted into an abstract base class or trait. 

Please review and tell me what you think of this validation pattern in general. I've mimicked the good aspects of the Symfony Validation component, without importing it. If you agree, I can change the code to use the Symfony Component instead of my "self-knitted" classes.

src/Validation/ConstraintViolation.php
```diff
@@ -0,0 +1,63 @@
...
+	 * @return mixed
+	 */
+	public function getValue() {
+		return $this->value;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getMessage() {
```
JeroenDeDauw: Is this new code? If so, y u no use php7 ;p

src/Validation/ConstraintViolation.php
```diff
@@ -0,0 +1,63 @@
...
+	private $source;
+	private $constraint;
+
+	/**
+	 * @param mixed $value The value that caused this violation
+	 * @param string $message Error message
+	 * @param mixed $constraint The validator instance that created this
+	 * @param string $source Class name or Class.Field name
+	 */
+	public function __construct( $value, string $message, $constraint = null, string $source = '' ) {
```
JeroenDeDauw: What's the deal with this highlighting of the empty string?

confused

src/Validation/RequestValidator.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+
+namespace WMDE\Fundraising\Frontend\Validation;
+
+use WMDE\Fundraising\Entities\Request;
```
JeroenDeDauw: The name of this entity is really unfortunate. Request already has a meaning in web apps, so now things such as this RequestValidator require a double take.

src/Validation/ValidationConstraint.php
```diff
@@ -0,0 +1,34 @@
...
+		$this->fieldName = $fieldName;
+		$this->validator = $validator;
+	}
+
+	/**
+	 * @param mixed $instance
+	 * @return null|ConstraintViolation
+	 */
+	public function validate( $instance ) {
+		$accessor = 'get' . ucfirst( $this->fieldName );
```
JeroenDeDauw: :(

src/Validation/ValidationConstraint.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+
+namespace WMDE\Fundraising\Frontend\Validation;
+
+/**
+ * Validate an instance property and add the source information to the ConstraintViolation
```
JeroenDeDauw: The grammar here is off - this is an instruction, and not a description of what something does

src/Validation/RequestValidator.php
```diff
@@ -0,0 +1,41 @@
...
+					return $constraint->validate( $request );
+				},
+			$this->constraints
+		) );
+		return count( $this->constraintViolations ) == 0;
+	}
+
+	public function getConstraintViolations(): array {
+		return $this->constraintViolations;
+	}
```
JeroenDeDauw: Looks like the only thing in this class specific to `Request` is what is in the constructor, with the rest being a "composite constraint" or something like that. I figure that if you put these two together like this, that you'll end up repeating the later elsewhere.

src/Validation/ValueValidator.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace WMDE\Fundraising\Frontend\Validation;
+
+/**
+ * Validate a single, scalar value
+ *
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+interface ValueValidator {
```
JeroenDeDauw: Note that we already have an interface with this name at WMDE (in a library used by the Wikibase software). If there is a good alternative for the name, this might be better to avoid potential confusion. No real objection to using `ValueValidator` though.

tests/Unit/Validation/ValidationConstraintTest.php
```diff
@@ -0,0 +1,53 @@
...
+		$this->validator->method( 'validate' )->willReturn( true );
+		$constraint = new ValidationConstraint( 'test', $this->validator );
+		$this->assertNull( $constraint->validate( $this->valueObject ) );
+	}
+
+	public function testGivenAnInvalidValue_constraintReturnsConstraintViolation() {
+
+		$this->validator->method( 'validate' )->willReturn( false );
+		$this->validator->method( 'getLastViolation' )->willReturn(
+			new ConstraintViolation( 'nyan', 'Too many cats!', $this->validator )
```
JeroenDeDauw: This can't be right. Too many cats? That's just ridiculous!

tests/Unit/Validation/ValidationConstraintTest.php
```diff
@@ -0,0 +1,53 @@
...
+		);
+		$constraint = new ValidationConstraint( 'test', $this->validator );
+		$this->assertInstanceOf( ConstraintViolation::class, $constraint->validate( $this->valueObject ) );
+	}
+
+	public function testGivenAnInvalidValue_ConstraintViolationSourceConatinsClassAndFieldName() {
```
JeroenDeDauw: _c

src/Validation/RequestValidator.php
gbirke: Since one of the things I'll be doing today is integrating the 2.x dev branch of the fundraising store, I'll rename the Entity there to something less confusing.

src/Validation/ValidationConstraint.php
gbirke: ?

gbirke: Any better suggestion or a short explanation why this is bad?

JeroenDeDauw: Same as what I wrote on #42

src/Validation/MailValidator.php
```diff
@@ -1,14 +1,18 @@
...
+
+use WMDE\Fundraising\Frontend\MailAddress;
 
 /**
  * @licence GNU GPL v2+
  * @author Christoph Fischer < christoph.fischer@wikimedia.de >
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
  */
-class MailValidator {
+class MailValidator implements ScalarValueValidator {
```
KaiNissen: The way this validator works it is actually also a required field validator. There is no way to make the email address an optional field, because passing an invalid value like an empty string, or null, always makes the validation fail.

src/Validation/RequestValidator.php
```diff
@@ -0,0 +1,23 @@
...
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class RequestValidator implements InstanceValidator {
+
+	use CanValidate;
+
+	public function __construct( MailValidator $mailValidator ) {
+		$this->setConstraints( [
+			new ValidationConstraint( 'email', $mailValidator ),
+			new ValidationConstraint( 'vorname', new RequiredFieldValidator() ),
```
KaiNissen: Shouldn't the `RequiredFieldValidator` also be injected? Or maybe the whole `ValidationConstraint[]`

tests/Unit/Validation/ValidationConstraintTest.php
```diff
@@ -0,0 +1,53 @@
...
+
+		$this->validator->method( 'validate' )->willReturn( false );
+		$this->validator->method( 'getLastViolation' )->willReturn(
+			new ConstraintViolation( 'nyan', 'Too many missing cats!', $this->validator )
+		);
+		$constraint = new ValidationConstraint( 'test', $this->validator );
+		$this->assertInstanceOf( ConstraintViolation::class, $constraint->validate( $this->valueObject ) );
+	}
+
+	public function testGivenAnInvalidValue_constraintViolationSourceConatinsClassAndFieldName() {
```
KaiNissen: conatins is an anagram of contains ;)

src/Validation/MailValidator.php
KaiNissen: I'd suggest that we either allow empty values and add the `ValidationConstraint` `RequiredFieldValidator` to the `RequestValidator` (or any other validator we might want to implement), or provide both an `OptionalMailValidator` and a `RequiredMailValidator`.

tests/Unit/Validation/ValidationConstraintTest.php
JeroenDeDauw: It conatins an anagram of contains? :)

src/Validation/RequestValidator.php
gbirke: Since RequiredFieldValidator does not need any external service or configuration value I'd say it's better to initialize it in RequestValidator. What do you think, @JeroenDeDauw ?

JeroenDeDauw: Looks fine to me, not sure what would be gained by injecting this.

tests/Integration/Validation/RequestValidatorTest.php
```diff
@@ -0,0 +1,71 @@
...
+		$this->assertConstraintWasViolated( $requestValidator->getConstraintViolations(), 'anrede' );
+	}
+
+	/**
+	 * @param ConstraintViolation[] $violations
+	 * @param string $fieldName
+	 */
+	private function assertConstraintWasViolated( array $violations, string $fieldName ) {
+		$this->assertCount( 1, $violations );
+		$this->assertInstanceOf( ConstraintViolation::class, current( $violations ) );
```
JeroenDeDauw: `assertContainsOnlyInstanceOf`

PR #42, opened 2016-01-13
Create SubscriptionRequest

gbirke:
Create SubscriptionRequest from Silex HTTP request data (GET, POST,
Cookies).

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -48,8 +48,11 @@ $app->post(
 	'contact/subscribe',
 	function( Request $request ) use ( $ffFactory ) {
 		$useCase = $ffFactory->newAddSubscriptionUseCase();
-		// TODO: Create SubscriptionRequest from $request params
-		$subscriptionRequest = new SubscriptionRequest();
+		$subscriptionRequest = SubscriptionRequest::createFromArray( $request->request->all() );
+		$subscriptionRequest->setWikiloginFromValues( [
+			$request->request->get( 'wikilogin' ),
+			$request->cookies->get( 'spenden_wikilogin' ),
+		] );
 		$responseModel = $useCase->addSubscription( $subscriptionRequest );
 		// TODO forward/dispatch to matching 'page/name' route, depending on $responseModel->getType();
 		return 'TODO';
--- a/app/routes.php
+++ b/app/routes.php
@@ -48,11 +48,25 @@ $app->post(
 	'contact/subscribe',
 	function( Request $request ) use ( $ffFactory ) {
 		$useCase = $ffFactory->newAddSubscriptionUseCase();
-		$subscriptionRequest = SubscriptionRequest::createFromArray( $request->request->all() );
+
+		$subscriptionRequest = new SubscriptionRequest();
+		$subscriptionRequest->setAddress( $request->get( 'address' ) );
+		$subscriptionRequest->setCity( $request->get( 'city' ) );
+		$subscriptionRequest->setPostcode( $request->get( 'postcode' ) );
+
+		$subscriptionRequest->setFirstName( $request->get( 'firstName' ) );
+		$subscriptionRequest->setLastName( $request->get( 'lastName' ) );
+		$subscriptionRequest->setSalutation( $request->get( 'salutation' ) );
+		$subscriptionRequest->setTitle( $request->get( 'title' ) );
+
+		$subscriptionRequest->setEmail( $request->get( 'email' ) );
+		$subscriptionRequest->setWikilogin( $request->get( 'wikilogin' ) );
+
 		$subscriptionRequest->setWikiloginFromValues( [
 			$request->request->get( 'wikilogin' ),
 			$request->cookies->get( 'spenden_wikilogin' ),
 		] );
+
 		$responseModel = $useCase->addSubscription( $subscriptionRequest );
 		// TODO forward/dispatch to matching 'page/name' route, depending on $responseModel->getType();
 		return 'TODO';
--- a/src/UseCases/AddSubscription/SubscriptionRequest.php
+++ b/src/UseCases/AddSubscription/SubscriptionRequest.php
@@ -91,4 +91,27 @@ class SubscriptionRequest {
 		$this->wikilogin = $wikilogin;
 	}
 
+	/**
+	 * Set the wikilogin value from the first value that matches /^(1|0|yes|no)$/
+	 *
+	 * @param array $values
+	 */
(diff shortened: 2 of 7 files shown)


src/UseCases/AddSubscription/SubscriptionRequest.php
```diff
@@ -91,4 +91,27 @@ public function setWikilogin( bool $wikilogin ) {
...
+	 */
+	public function setWikiloginFromValues( array $values ) {
+		$trueValues = ['yes', '1'];
+		$falseValues = ['no', '0'];
+		$matchingValues = array_intersect( $values, array_merge( $trueValues, $falseValues ) );
+		$wikilogin = in_array( array_shift( $matchingValues ), $trueValues );
+		$this->setWikilogin( $wikilogin );
+	}
+
+	public static function createFromArray( array $values ): SubscriptionRequest {
```
JeroenDeDauw: This is presentation related code on the wrong side of the boundary.

src/UseCases/AddSubscription/SubscriptionRequest.php
```diff
@@ -91,4 +91,27 @@ public function setWikilogin( bool $wikilogin ) {
...
+	 * @param array $values
+	 */
+	public function setWikiloginFromValues( array $values ) {
+		$trueValues = ['yes', '1'];
+		$falseValues = ['no', '0'];
+		$matchingValues = array_intersect( $values, array_merge( $trueValues, $falseValues ) );
+		$wikilogin = in_array( array_shift( $matchingValues ), $trueValues );
+		$this->setWikilogin( $wikilogin );
+	}
+
```
JeroenDeDauw: In such a case you can use `self` as type hint

src/UseCases/AddSubscription/SubscriptionRequest.php
```diff
@@ -91,4 +91,27 @@ public function setWikilogin( bool $wikilogin ) {
...
+	public static function createFromArray( array $values ): SubscriptionRequest {
+		$instance = new self();
+		foreach ( $values as $key => $value ) {
+			$accessor = 'set' . ucfirst( $key );
+			if ( is_callable( [$instance, $accessor ] ) ) {
+				call_user_func( [$instance, $accessor ], $value );
```
JeroenDeDauw: I really do not like this kind of magic. I understand that doing

```php
$object->setFoo( $params['foo'] );
$object->setBar( $params['bar'] );
// ...
```

is boring. However it's very easy to understand and to maintain. It does not mess up static code analysis, and it does not make me wonder if there is some way this code could be exploited by a malicious agent.

JeroenDeDauw: This pattern also prevents using composition without adding shortcut methods. Looks like `SubscriptionRequest` could contain a `PesonName` value object with first and last name and an `Address` value object with street address, postcode and city.

app/routes.php
```diff
@@ -48,8 +48,11 @@ function( Application $app, $pageName ) use ( $ffFactory ) {
 	'contact/subscribe',
 	function( Request $request ) use ( $ffFactory ) {
 		$useCase = $ffFactory->newAddSubscriptionUseCase();
-		// TODO: Create SubscriptionRequest from $request params
-		$subscriptionRequest = new SubscriptionRequest();
+		$subscriptionRequest = SubscriptionRequest::createFromArray( $request->request->all() );
```
JeroenDeDauw: The new code here appears to be untested

tests/Unit/UseCases/AddSubscription/SubscriptionRequestTest.php
```diff
@@ -0,0 +1,53 @@
...
+		] );
+		$this->assertEquals( $subscriptionRequest->getFirstName(), 'Nyan' );
+		$this->assertEquals( $subscriptionRequest->getLastName(), 'Cat' );
+		$this->assertEquals( $subscriptionRequest->getSalutation(), 'Herr' );
+		$this->assertEquals( $subscriptionRequest->getTitle(), 'Prof. Dr. Dr.' );
+		$this->assertEquals( $subscriptionRequest->getAddress(), 'Awesome Way 1' );
+		$this->assertEquals( $subscriptionRequest->getCity(), 'Berlin' );
+		$this->assertEquals( $subscriptionRequest->getPostcode(), '12345' );
+		$this->assertEquals( $subscriptionRequest->getEmail(), 'me@nyancat.wtf' );
+		$this->assertEquals( $subscriptionRequest->getWikilogin(), true );
```
JeroenDeDauw: assertSame would be more appropriate for all of these

tests/Unit/UseCases/AddSubscription/SubscriptionRequestTest.php
```diff
@@ -0,0 +1,53 @@
...
+ * @licence GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class SubscriptionRequestTest extends \PHPUnit_Framework_TestCase {
+	public function testSubscriptionRequestCanBeInitializedFromArray() {
+		$subscriptionRequest = SubscriptionRequest::createFromArray( [
+			'firstName' => 'Nyan',
+			'lastName' => 'Cat',
+			'salutation' => 'Herr',
+			'title' => 'Prof. Dr. Dr.',
```
JeroenDeDauw: Took me till seeing this to understand what title meant. In the request TDO this was certainly not clear to me. This is another reason to not use the dynamic mapping magic: it forces the setter names to match those of the presentation interface.

src/UseCases/AddSubscription/SubscriptionRequest.php
JeroenDeDauw: Fixed by #46

src/UseCases/AddSubscription/SubscriptionRequest.php
JeroenDeDauw: Replaced in #46

app/routes.php
JeroenDeDauw: Partially addressed by #46. Even found a bug when writing the test :)

tests/Unit/UseCases/AddSubscription/SubscriptionRequestTest.php
JeroenDeDauw: Made obsolete by #46

gbirke:
Since I've accepted the edits by Jeroen, this can now be merged.

PR #43, opened 2016-01-13
Work on the display page usecase

JeroenDeDauw:
Resubmission of #18

PR #48, opened 2016-01-14
Add API based PageRetriever

JeroenDeDauw:
Submitted against #43


composer.json
```diff
@@ -10,6 +10,7 @@
 
 		"jeroen/file-fetcher": "~3.1",
 		"wmde/fundraising-store": "~1.0",
+		"addwiki/mediawiki-api-base": "~2.1",
```
JeroenDeDauw: @addshore ALL your codes are belong to us!

PR #52, opened 2016-01-15
Use Pimple in the FunFunFactory

JeroenDeDauw:


JeroenDeDauw:
So far all access to stuff in the pimple container happens via wrapper functions, so even internally static code analysis is happy.

JeroenDeDauw:
Fun fact: Pimple 1.x series only has 1.0.0 from 2012, which is what Silex uses. Latest Pimple version is 3.0.2.

PR #58, opened 2016-01-15
configurable base path for asset reference in twig templates

KaiNissen:
This PR takes into account that the application may be running in a subdirectory of the webroot.


KaiNissen:
Builds are currently failing due to a malfunction on sourceforge.

JeroenDeDauw:
Could have written a little system test for this one

PR #68, opened 2016-01-19
Use PHP strict types in tests

JeroenDeDauw:


KaiNissen:
Shouldn't `declare(strict_types = 1)` also work? Or ignoring the rule for these specific lines?

JeroenDeDauw:
Indeed Kai, that works. I did not actually really see what the issue was and just assumed a PHP7 compat problem...

PR #70, opened 2016-01-19
Add presenter to list comments UC

JeroenDeDauw:


src/DataAccess/DbalCommentRepository.php
```diff
@@ -33,9 +33,20 @@ function( Spenden $spenden ) {
...
+					->setDonationId( $spenden->getId() )
+					->freeze()
+					->assertNoNullFields();
 			},
-			$this->entityRepository->findBy( [ 'isPublic' => true ], null, $limit )
+			$this->getSpenden( $limit )
+		);
+	}
+
+	private function getSpenden( int $limit ): array {
```
gbirke: Should'nt that be called getDonations, even when the entity is called Spenden (until we use the 2.x branch)?

JeroenDeDauw: I've stuck with Spenden for with the intention of this being changed when we switch to the renamed entities. For this method it really does not matter much to me, as it is private. Feel free to change it already if you prefer that.

PR #71, opened 2016-01-19
Load wiki pages as Twig templates

gbirke:
This PR allows templates to come from the wiki.

Since the `{{ variableName }}` placeholder is already used by wiki text, the Twig templates must use `{$ variableName $}` instead.


tests/Unit/TwigPageLoaderTest.php
```diff
@@ -0,0 +1,61 @@
...
+		$pageRetriever = $this->getMock( PageRetriever::class );
+		$pageRetriever->method( 'fetchPage' )->willReturn( '' );
+		$loader = new TwigPageLoader( $pageRetriever );
+		$loader->getCacheKey( 'Felis silvestris' );
+	}
+
+	/**
+	 * @expectedException \Twig_Error_Loader
+	 */
+	public function testGivenMissingPage_isFreshKeyThrowsException() {
```
JeroenDeDauw: I think it's better to use setExpectedException, and do so right before the line that is expected to cause the exception. Like this the test will fail if the setup code unexpectedly throws the exception

src/TwigFactory.php
```diff
@@ -0,0 +1,44 @@
...
+use Twig_Loader_Filesystem;
+use WMDE\Fundraising\Frontend\PageRetriever\PageRetriever;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class TwigFactory {
+
+	public static function newFromConfig( array $config, PageRetriever $pageRetriever ): Twig_Environment {
```
JeroenDeDauw: Could have been non-static!

src/TwigPageLoader.php
```diff
@@ -0,0 +1,61 @@
...
+		return $this->retrievePage( $name );
+	}
+
+	public function getCacheKey( $name ) {
+		// retrieve page to generate loader exception if page does not exist
+		$this->retrievePage( $name );
+		return $name;
+	}
+
+	public function isFresh( $name, $time ) {
```
JeroenDeDauw: <darth vader voice>I find your lack of type hints disturbing</darth vader voice>

gbirke: `<squeaky voice>b-b-but it's just implementing the interface!</squeaky voice>`

PR #76, opened 2016-01-21
Modify content

gbirke:
Remove wiki-specific markup from templates, allow rendering variables as templates

### DIFF
--- a/app/templates/DisplayPageLayout.twig
+++ b/app/templates/DisplayPageLayout.twig
@@ -56,7 +56,7 @@
     <!-- Header -->
     <header id="header">
         <div class="container clearfix">
-            {$ header | raw $}
+            {% include 'header' %}
         </div>
     </header>
 
@@ -64,14 +64,14 @@
     <!-- Main -->
     <main id="main">
         <div class="container">
-            {$ main | raw $}
+            {% include 'main' %}
         </div>
     </main>
 
     <!-- Footer -->
     <footer id="footer">
         <div class="container">
-            {$ footer | raw $}
+            {%  include 'footer' %}
         </div>
     </footer>
 </div>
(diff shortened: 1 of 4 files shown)


app/templates/DisplayPageLayout.twig
```diff
@@ -56,7 +56,7 @@
     <!-- Header -->
     <header id="header">
         <div class="container clearfix">
-            {$ header | raw $}
+            {% include 'header' %}
```
JeroenDeDauw: Is this just some other syntax, or does it do something else?

gbirke: `{$ header $}` is replacing the placeholder with the contents of the context variable `header`.
`{% include 'header' %}` looks for a template named `header`.

JeroenDeDauw: > `{% include 'header' %}` looks for a template named header.

... and falls back to the context variable?

gbirke: No. When the `header` template is not found, there is an error message. But that's a good thing since that will point to a mismatch between the includes and the arguments that were passed to `renderArgumentsAsTemplates`. The unit tests will catch this.

src/TwigTemplate.php
```diff
@@ -23,4 +26,20 @@ public function render( array $arguments ): string {
...
+	 *
+	 * @param array $arguments
+	 * @return string
+	 */
+	public function renderArgumentsAsTemplates( array $arguments ): string {
+		$loader = new Twig_Loader_Chain( [
+			new Twig_Loader_Array( $arguments ),
+			$this->twig->getLoader()
+		] );
+		$this->twig->setLoader( $loader );
```
JeroenDeDauw: Like this a dependency that was constructor injected gets modified on calling this method. That's unexpected state change. Is there no way to avoid this?

app/templates/DisplayPageLayout.twig
JeroenDeDauw: Right. I had misread the other code. Makes sense now

gbirke:
This will be implemented in a different way since all use cases that output HTML need header, footer and nonjs-areas.

PR #77, opened 2016-01-21
Strip twig suffix in PageLoader

gbirke:
Pages in the CMS have no suffix. This patch makes it easier to prototype
new templates as files before putting them in the wiki.


JeroenDeDauw:
Perhaps its good to add a small doc somewhere explaining how this works?

PR #78, opened 2016-01-21
Add comment-list.rss route

JeroenDeDauw:


app/templates/CommentList.rss.twig
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<rss version="2.0">
+    <channel>
+        <title>Spenden an Wikimedia</title>
+        <link>{$ rssLink |e $}</link>
```
gbirke: All variables are [auto-escaped for HTML output](http://twig.sensiolabs.org/doc/api.html#environment-options). Adding `|e` to all placeholders is unneccessarily explicit and verbose IMHO.

JeroenDeDauw: You sure this is on for this template? In case of security, I much prefer explicitness over smart magic to reduce verbosity.

In any case, should also add a test to verify stuff is being escaped

gbirke: I'm sure. I've added a test in #80 and removed the `|e` and it still passed. But keep them if you like.

PR #79, opened 2016-01-21
Inject app into Twig

gbirke:
Now templates can use the app container for things like UrlGenerator plugin.


JeroenDeDauw:
I'm still ambivalent about this: like this we end up with binding of presenters to the framework, while at present nothing in `src/`depends on it. Not the end of the world, so let's see how this goes. It's not clear to me how much value this service will provide. If it turns out to not warrant its cost, we can always change it later.

PR #80, opened 2016-01-21
Test escaping in RSS

gbirke:


JeroenDeDauw:
Cherry-picked comit into 78

PR #81, opened 2016-01-22
Put internets using validation behind an interface

JeroenDeDauw:
Still working on removing the internets usage from the tests


JeroenDeDauw:
Featuring: anonymous class

anonymous

gbirke:
+1

src/Validation/MailValidator.php
```diff
@@ -13,14 +14,11 @@
...
-	private $testWithMX;
+	private $domainValidator;
 	private $lastViolation;
 
-	const TEST_WITH_MX = true;
-	const TEST_WITHOUT_MX = false;
-
-	public function __construct( bool $testWithMX ) {
-		$this->testWithMX = $testWithMX;
+	public function __construct( DomainNameValidator $tldValidator ) {
```
KaiNissen: Not really an issue, but it's actually not a top level validator.

KaiNissen:
Apart from my picky comment, also +1

src/Validation/MailValidator.php
JeroenDeDauw: Yeah, I forgot to rename this var. Originally I thought this stuff only did TLD validation.

PR #82, opened 2016-01-22
changed domain name to be tested

KaiNissen:
This is a quick fix for `MailValidatorTest` to pass. It just changes the domain name to be tested to that of the tld registry of the UAE (which is problably more fail-safe).


JeroenDeDauw:
Guess I should have pinged you on IRC. #81

PR #83, opened 2016-01-22
Contact form

KaiNissen:
This PR adds the route for the contact form to be submitted to. It still lacks to display the proper content for failed and successful requests.


JeroenDeDauw:
I've rebased the code and added a commit on top that fixes failures caused by changes that where made on master in the meanwhile.

phpcs.xml
```diff
@@ -6,6 +6,9 @@
 <ruleset name="MediaWiki">
     <rule ref="Generic.Classes" />
     <rule ref="Generic.CodeAnalysis" />
+    <rule ref="Generic.CodeAnalysis.UnusedFunctionParameter.Found">
+        <severity>0</severity>
+    </rule>
```
JeroenDeDauw: @gbirke I'm disabling this again, as this rule just has too many false positives. PhpStorm has a sniff to find this kind of stuff which is somewhat more intelligent - it will not shout at you in places where an interface or base class forces a param to be present.

src/FunFunFactory.php
```diff
@@ -266,4 +280,32 @@ public function setRequestValidator( RequestValidator $requestValidator ) {
...
+
+	public function setContactValidator( GetInTouchValidator $validator ) {
+		$this->pimple['contact_validator'] = $validator;
+	}
+
+	private function getContactValidator(): GetInTouchValidator {
+		return $this->pimple['contact_validator'];
+	}
+
+	public function getMessenger(): Messenger {
```
JeroenDeDauw: private?

src/FunFunFactory.php
```diff
@@ -266,4 +280,32 @@ public function setRequestValidator( RequestValidator $requestValidator ) {
 	public function setPageTitlePrefix( string $prefix ) {
 		$this->config['cms-wiki-title-prefix'] = $prefix;
 	}
+
+	public function newGetInTouchUseCase() {
+		return new GetInTouchUseCase( $this->getContactValidator(), $this->getMessenger() );
+	}
+
+	public function setContactValidator( GetInTouchValidator $validator ) {
```
JeroenDeDauw: not used?

tests/Integration/Validation/ValidatorTestCase.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class ValidatorTestCase extends \PHPUnit_Framework_TestCase {
+
+	/**
+	 * @param ConstraintViolation[] $violations
+	 * @param string $fieldName
+	 */
+	protected function assertConstraintWasViolated( array $violations, string $fieldName ) {
+		$this->assertCount( 1, $violations );
+		$this->assertInstanceOf( ConstraintViolation::class, current( $violations ) );
```
JeroenDeDauw: ```$this->assertContainsOnlyInstanceOf( ConstraintViolation::class, $violations );```

tests/System/MessengerTest.php
```diff
@@ -0,0 +1,64 @@
...
+		$message = new Swift_Message( $subject, $body );
+		$message->setFrom( $operatorAddress->getNormalizedAddress(), $operatorAddress->getDisplayName() );
+		$message->setTo( $receiver->getNormalizedAddress(), $receiver->getDisplayName() );
+		$message->setReplyTo( $sender->getNormalizedAddress(), $sender->getDisplayName() );
+		return $message;
+	}
+
+	private function createActualMessage( MailAddress $sender, MailAddress $receiver,
+										  string $subject, string $body ) {
+		return $this->messenger->constructMessage( $sender, $receiver, $subject, $body );
```
JeroenDeDauw: Not sure what you gain by having this method. If you get rid of it, you also can drop the `$this->messenger` field.

src/UseCases/GetInTouch/GetInTouchUseCase.php
```diff
@@ -0,0 +1,70 @@
...
+					$this->request->getEmailAddress(),
+					implode( ' ', [ $this->request->getFirstName(), $this->request->getLastName() ] )
+				),
+				new MailAddress( 'kai.nissen@wikimedia.de' ),
+				$this->request->getSubject(),
+				$this->request->getMessageBody()
+			)
+		);
+
+		return $this->messenger->getFailedRecipients() === [];
```
JeroenDeDauw: This code looks like it deals with multiple recipients but it does not correctly I think. As soon as there is one recipient to which the message was not send, this code will behave as if sending altogether failed, while stuff was still send to the other recipients.

`Messenger` currently only sends to a single one, so perhaps its interface should just be changed accordingly. Rather than having this `getFailedRecipients`, the `sendMessage` method could throw an exception on failure.

src/MailAddress.php
```diff
@@ -7,21 +7,46 @@
...
-	public $userName;
-	public $domain;
+	private $userName;
+	private $domain;
+	private $displayName;
 
-	public function __construct( string $userName, string $domain ) {
-		$this->userName = $userName;
-		$this->domain = $domain;
+	public function __construct( string $emailAddress, string $displayName = '' ) {
```
JeroenDeDauw: I like you moved a bunch of stuff here. Not sure about the constructor change though. Might be better to have a `static newFromString` function. `$displayName` is not required by the existing users of this value object, which is an indication it does not belong here. How about creating a new value object, perhaps one that contains a MailAddress?

src/Messenger.php
```diff
@@ -0,0 +1,45 @@
...
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class Messenger {
+
+	private $mailTransport;
+	private $failedRecipients = [];
+	private $operatorAddress;
+
+	public function __construct( Swift_Transport $mailTransport, MailAddress $operatorAddress ) {
```
JeroenDeDauw: The code in this class is not cohesive. `$operatorAddress` is only used by `constructMessage` and the other two fields are only used by `sendMessage` (and `getFailedRecipients`).

I'm a bit confused by what this ought to be. You can split the two things, and perhaps drop the sendMessage part as its rather simple (even though that means injecting a service from an external library). Alternatively I see how putting them together could make sense if you do something like `sendMessage( MailAddress $sender, MailAddress $receiver, string $subject, string $body )`. When doing that, you decouple the user from Swift stuff altogether.

src/Validation/GetInTouchValidator.php
```diff
@@ -0,0 +1,44 @@
...
+namespace WMDE\Fundraising\Frontend\Validation;
+
+use WMDE\Fundraising\Frontend\UseCases\GetInTouch\GetInTouchRequest;
+
+/**
+ * @license GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class GetInTouchValidator implements InstanceValidator {
+
```
JeroenDeDauw: no newline here please

JeroenDeDauw:
Minor follow up at #85

src/MailAddress.php
KaiNissen: Initially I added a static function like that, but then I figured that it is actually quite unlikely to construct a MailAddress by passing a split email address. You're right about the $displayName, though. The only user would still be the 'operator', so maybe having a value object for it is unnecessary.

src/Messenger.php
KaiNissen: The idea of introducing this was to have some kind of wrapper for the SwiftMailer lib. I changed it to what you proposed alternatively.

PR #85, opened 2016-01-24
Add type hints to Messenger

JeroenDeDauw:
Follow up to #83


KaiNissen:
Messenger was changed in bfc65db.

PR #86, opened 2016-01-25
Default HTML presentation layout

gbirke:
Define a common, extendable layout with default content for header and footer, where each use case can define its own content and reuse the common HTML layout.

Trimmed down the DisplayPage use case so it doesn't need to concern itself with the default layout areas.

When this PR gets accepted I'll write a short documentation on how to use the base template in a use case.


src/Presenters/Content/WikiContentProvider.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Presenters\Content;
```
JeroenDeDauw: Think `WMDE\Fundraising\Frontend\DataAccess` is more appropriate

gbirke: Will do that. Does PageContentModifier belong there also?

2016-01-25 14:29 GMT+01:00 Jeroen De Dauw <notifications@github.com>:

> In src/Presenters/Content/WikiContentProvider.php
> <https://github.com/wmde/FundraisingFrontend/pull/86#discussion_r50692784>
> :
>
> > @@ -0,0 +1,40 @@
> > +<?php
> > +
> > +namespace WMDE\Fundraising\Frontend\Presenters\Content;
>
> Think WMDE\Fundraising\Frontend\DataAccess is more appropriate
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/wmde/FundraisingFrontend/pull/86/files#r50692784>.
>



-- 
Gabriel Birke
Software Developer

Wikimedia Deutschland e. V. | Tempelhofer Ufer 23-24 | 10963 Berlin
Phone: +49 (0)30 219 158 26-0
http://wikimedia.de

Imagine a world, in which every single human being can freely share in the
sum of all knowledge. That‘s our commitment.

Wikimedia Deutschland - Gesellschaft zur Förderung Freien Wissens e. V.
Eingetragen im Vereinsregister des Amtsgerichts Berlin-Charlottenburg unter
der Nummer 23855 B. Als gemeinnützig anerkannt durch das Finanzamt für
Körperschaften I Berlin, Steuernummer 27/681/51985.

JeroenDeDauw: I actually wanted to remove/edit my comment but my phones connection failed.

Currently the code in the Presenters folder only has presenters that all only depend on response model objects, value objects from the domain and presentation services (ie the TwigTemple). There are no dependencies on domain services or anything data access related. That was my original idea of this folder, just such presenters, and perhaps parts of them if some get split into multiple classes.

Perhaps we need a more general `Presentation` folder, in which `Presenters` is a sub folder? This can then hold presentation services and whatever `WikiContentProvider` and `PageContentModifier` are.

JeroenDeDauw: I'm fine with this being merged as-is btw. We can move this stuff around further afterwards.

JeroenDeDauw:
+1

PR #87, opened 2016-01-25
Subscription presentation

gbirke:
HTML Presenter for AddSubscription error state (success state is handled by redirect to `/page/SubscriptionSuccess`)
JSON Presenter for AddSubscription success and error state.

This PR contains the changes made in #86. Those changes should be merged before merging this.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -74,7 +74,7 @@ $app->get(
 
 $app->post(
 	'contact/subscribe',
-	function( Request $request ) use ( $ffFactory ) {
+	function( Application $app, Request $request ) use ( $ffFactory ) {
 		$useCase = $ffFactory->newAddSubscriptionUseCase();
 
 		$subscriptionRequest = new SubscriptionRequest();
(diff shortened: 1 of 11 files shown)


app/routes.php
```diff
@@ -95,10 +97,16 @@ function( Request $request ) use ( $ffFactory ) {
 		] );
 
 		$responseModel = $useCase->addSubscription( $subscriptionRequest );
-		// TODO forward/dispatch to matching 'page/name' route, depending on $responseModel->getType();
-		return 'TODO';
+		if ( in_array( 'application/json', $request->getAcceptableContentTypes() ) ) {
+			return $app->json( $ffFactory->newAddSubscriptionJSONPresenter()->present( $responseModel ) );
+		}
+		if ( $responseModel->isSuccessful() ) {
+			return $app->redirect( $app['url_generator']->generate('page', [ 'pageName' => 'SubscriptionSuccess' ] ) );
```
KaiNissen: Maybe we should forward the request using `$app->handle()`? Doing it like this will send 302 to the browser, which then re-requests.

gbirke: This is intentional: When the user is forwarded after submitting a form, he can hit the reload button without accidentally resubmitting it. See https://en.wikipedia.org/wiki/Post/Redirect/Get
The redirect is only for the success state, the error state will present the form again.

KaiNissen: Good point. Forget, that I ever mentioned it.

KaiNissen:
rebased

PR #88, opened 2016-01-26
changed AddSubscriptionResponse to generic ValidationResponse

KaiNissen:
The response model is quite generic for validation requests, and could be reused by other use case implementations.


JeroenDeDauw:
When doing this, we should be careful to not add usecase specific things to the generic thing. If there is need for usecase specific stuff, it's time to move away from the generic solution.

KaiNissen:
Sure, but we can still derive from it, I guess.

JeroenDeDauw:
+1

gbirke:
I'm also not sure if this is a case of premature generalization, but will merge it none the less.

PR #90, opened 2016-01-27
GetInTouch response

KaiNissen:
This PR introduces the actual response model for the use case. I also made some changes to styles and changed the on-wiki templates to use twig (and added some layout stuff to make the [contact confirmation](https://backend.wikimedia.de/wiki/index.php?title=Web%3ASpendenseite-HK2013%2Frewrite%2FKontaktBestaetigung&diff=7035&oldid=6922) and the [contact form](https://backend.wikimedia.de/wiki/index.php?title=Web%3ASpendenseite-HK2013%2Frewrite%2FKontaktformular&diff=7034&oldid=6925) look prettier).

### DIFF
--- a/src/UseCases/GetInTouch/GetInTouchUseCase.php
+++ b/src/UseCases/GetInTouch/GetInTouchUseCase.php
@@ -6,6 +6,7 @@ namespace WMDE\Fundraising\Frontend\UseCases\GetInTouch;
 
 use WMDE\Fundraising\Frontend\MailAddress;
 use WMDE\Fundraising\Frontend\Messenger;
+use WMDE\Fundraising\Frontend\ResponseModel\ValidationResponse;
 use WMDE\Fundraising\Frontend\Validation\GetInTouchValidator;
 
 /**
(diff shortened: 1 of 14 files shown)


src/UseCases/GetInTouch/GetInTouchUseCase.php
```diff
@@ -24,17 +25,17 @@ public function __construct( GetInTouchValidator $validator, Messenger $messenge
...
+			return ValidationResponse::newFailureResponse( $this->validator->getConstraintViolations() );
 		}
 
 		try {
 			$this->forwardContactRequest();
 			$this->confirmToUser();
-			return 'request successful';
+			return ValidationResponse::newSuccessResponse();
 		} catch ( \RuntimeException $e ) {
 			return 'mail transmission failed';
```
gbirke: This will create a TypeError because a string is returned. Either don't catch the error (it will be handled by the default error handler) or create a new response.
Since there is nothing much the user can do in this case (and we don't need to inform him what failed specifically), I'd recommend not handling the exception and displaying the general error response.

JeroenDeDauw: You're right, the way the code got changed now will cause it to error out on error. Not sure about not catching the exception. Generally not nice to have such service specific stuff bubble up. Isn't returning a failure response model here very easy?

JeroenDeDauw:
Generally looks good to me

src/UseCases/GetInTouch/GetInTouchUseCase.php
gbirke: Sure, but to be very explicit, this would have to be a different response than `ValidationResponse` (because it's no validation error) and the route (or Presenter) would have to check for the different response types.

KaiNissen: Yes, you're both right. I let the route handle exceptions and display an error page on those cases.

PR #92, opened 2016-01-28
Remove MediaWiki API dependency for test templates

gbirke:
All tests that were using templates were dependent on the MediaWiki API
and needed complicated setup of templates wrapped in JSON. This patch
introduces the ability to turn the MediaWiki page loading off and
replace the JSON responses with simple Twig templates.

The following changes were made:
* Put all twig configuration in one place in JSON config
* Refactor TwigFactory to use the new twig configuration array to
  create various twig classes.
* Add relative path loading for file system based templates.
* Remove MediaWiki API stubs and their JSON data files from tests.
* Add .twig suffix to included wiki page templates. The Wiki page
  loader will automatically remove the suffix and the tests can use
  files with the correct highlighting.
* Remove Twig injection into $app - this created a Twig instance
  even when none was needed.

### DIFF
--- a/src/TwigFactory.php
+++ b/src/TwigFactory.php
@@ -16,6 +16,8 @@ use WMDE\Fundraising\Frontend\Presenters\Content\WikiContentProvider;
  */
 class TwigFactory {
 
+	const DEFAULT_TEMPLATE_DIR = 'app/templates';
+
 	private $config;
 
 	public function __construct( array $config ) {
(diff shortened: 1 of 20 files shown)


src/TwigFactory.php
```diff
@@ -44,4 +47,42 @@ public static function newFromConfig( array $config, WikiContentProvider $provid
...
+			return null;
+		}
+		return new TwigPageLoader( $provider );
+	}
+
+	public function newFileSystemLoader() {
+		if ( empty( $this->config['loaders']['filesystem'] ) ) {
+			return null;
+		}
+		if ( empty( $this->config['loaders']['filesystem']['template-dir'] ) ) {
```
JeroenDeDauw: $templateDir = $this->getTemplateDir()

src/TwigFactory.php
```diff
@@ -44,4 +47,42 @@ public static function newFromConfig( array $config, WikiContentProvider $provid
...
+			if ( strlen( $dir ) == 0 || $dir{0} != '/' ) {
+				return $appRoot . $dir;
+			}
+			return $dir;
+		}, $templateDir );
+		return new Twig_Loader_Filesystem( $templateDir );
+	}
+
+	public function newArrayLoader() {
+		$templates = $this->config['loaders']['array'] ?? [];
```
JeroenDeDauw: This is valid PHP ?? :)

gbirke: Yea, it avoids errors when any of the array elements is missing. One of the great features of PHP 7.

http://www.lornajane.net/posts/2015/new-in-php-7-null-coalesce-operator

JeroenDeDauw: Right. Btw, do we have a list somewhere of PHP7 features we still have to use in this project? Gotta use 'em all!

PR #93, opened 2016-01-29
Send Subscription email

gbirke:
Added a new class, TemplatedMessenger, that that is prepared with a mail
template and only receives data for sending messages. This avoids having
to inject four additional  dependencies (mail template, translatable
subject line, Messenger, recipient) into the use cases just to send an email.

Made mail recipients configurable.

Did some cleanup in the add subscription use case class.

This is based on #92, please merge that first.

### DIFF
--- a/app/config/config.test.json
+++ b/app/config/config.test.json
@@ -21,5 +21,9 @@
 		}
 	},
 	"operator-email": "email@operatorsownmailserver.com",
-	"operator-displayname": "Friendly Operator"
+	"operator-displayname": "Friendly Operator",
+	"mail-recipients": {
+		"add-subscription": "email@operatorsownmailserver.com",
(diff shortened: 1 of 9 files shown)


JeroenDeDauw:
Ugh, why does GitHub not auto-rebase this? Now we need to do it manually each time for a proper diff

app/config/config.test.json
```diff
@@ -21,5 +21,9 @@
 		}
 	},
 	"operator-email": "email@operatorsownmailserver.com",
-	"operator-displayname": "Friendly Operator"
+	"operator-displayname": "Friendly Operator",
+	"mail-recipients": {
+		"add-subscription": "email@operatorsownmailserver.com",
```
JeroenDeDauw: should these not be different from `operator-email` so you can catch them getting mixed up in the code?

PR #94, opened 2016-02-02
Templated messenger

KaiNissen:
This adds a message interface as well as its implementations `SimpleMessage` and `TemplatedMessage`. In favor of this the `TemplatedMessenger` is dropped and `Messenger::sendMessage()` accepts `Message` instead of `$subject` and `$body`.

`Messenger` is also slightly improved, as there now is a method to send emails to the operator which does not expect a recipient parameter.

Lastly, the recipient configuration was removed, since all the mails need to be sent to a recipient determined at runtime; only in the use case GetInTouch a message will be sent to a static email address.


gbirke:
:+1:

PR #95, opened 2016-02-02
retrieve raw wikitext

KaiNissen:
This changes the TwigLoader for wiki content to be able to retrieve raw wikitext, which is needed for plain text emails.

### DIFF
--- a/src/DataAccess/ApiBasedPageRetriever.php
+++ b/src/DataAccess/ApiBasedPageRetriever.php
@@ -41,18 +41,7 @@ class ApiBasedPageRetriever implements PageRetriever {
 			$this->doLogin();
 		}
 
-		$content = false;
-		switch ( $action ) {
-			case 'raw':
-				$content = $this->retrieveWikiText( $pageTitle );
-				break;
-			case 'render':
-				$content = $this->retrieveRenderedPage( $pageTitle );
-				break;
-			default:
-				throw new \RuntimeException( 'Action "' . $action . '" not supported' );
-				break;
-		}
+		$content = $this->retrieveContent( $pageTitle, $action );
 
 		if ( $content === false || $content === null ) {
 			$this->logger->debug( __METHOD__ . ': fail, got non-value', [ $content ] );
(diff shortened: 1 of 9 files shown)


src/DataAccess/ApiBasedPageRetriever.php
```diff
@@ -28,14 +28,24 @@ public function __construct( MediawikiApi $api, ApiUser $apiUser, LoggerInterfac
...
+		$content = false;
+		switch ( $action ) {
+			case 'raw':
+				$content = $this->retrieveWikiText( $pageTitle );
+				break;
+			case 'render':
+				$content = $this->retrieveRenderedPage( $pageTitle );
+				break;
+			default:
+				break;
```
gbirke: Should't this fail loudly with a RuntimeException?

src/DataAccess/ApiBasedPageRetriever.php
```diff
@@ -64,4 +81,23 @@ private function retrieveRenderedPage( $pageTitle ) {
...
+	private function retrieveWikiText( $pageTitle ) {
+		$params = [
+			'titles' => $pageTitle,
+			'prop' => 'revisions',
+			'rvprop' => 'content'
+		];
+
+		try {
+			$response = $this->api->postRequest( new SimpleRequest( 'query', $params ) );
+			if ( !is_array( $response['query']['pages'] ) ) {
```
JeroenDeDauw: Code starting at this if block can be moved out of the try-catch

src/DataAccess/ApiBasedPageRetriever.php
```diff
@@ -28,14 +28,31 @@ public function __construct( MediawikiApi $api, ApiUser $apiUser, LoggerInterfac
...
+	 */
+	public function fetchPage( string $pageTitle, string $action = 'render' ): string {
 		$this->logger->debug( __METHOD__ . ': pageTitle', [ $pageTitle ] );
 
 		if ( !$this->api->isLoggedin() ) {
 			$this->doLogin();
 		}
 
-		$content = $this->retrieveRenderedPage( $pageTitle );
+		$content = false;
```
JeroenDeDauw: Looks like this assignment has no effect

Perhaps its worthwhile to factor out the switch into a `retrieveContent` method?

src/TwigFactory.php
```diff
@@ -51,10 +51,10 @@ public function create( array $loaders ): Twig_Environment {
 	}
 
 	public function newWikiPageLoader( WikiContentProvider $provider ) {
-		if ( empty( $this->config['loaders']['wiki'] ) ) {
+		if ( !$this->config['loaders']['wiki']['enabled'] ) {
 			return null;
 		}
-		return new TwigPageLoader( $provider );
+		return new TwigPageLoader( $provider, $this->config['loaders']['wiki']['rawpages'] ?? [] );
```
JeroenDeDauw: should this config not always exist? If so, the defaulting is not needed

JeroenDeDauw:
Generally looks good

src/TwigFactory.php
KaiNissen: Hm, can we be sure this field always exists? One might overwrite the settings in `config.prod.json` and break this, because "I don't want no raw pages here".

JeroenDeDauw: If you omit the field, the recursive replace will not replace it, and it will end up being the default.

PR #96, opened 2016-02-02
Add ConfirmationCodeConverter

gbirke:
The `ConfirmationCodeConverter` converts the binary data in the
`confirmationCode` field of the `Subscription` entity into readable codes
that can be safely inserted in URLs and templates.

### DIFF
--- a/src/ConfirmationCodeConverter.php
+++ b/src/ConfirmationCodeConverter.php
@@ -1,0 +1,21 @@
+<?php
+
+
+namespace WMDE\Fundraising\Frontend;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ConfirmationCodeConverter {
+
+	public function fromBinaryToReadable( string $binaryString ): string {
+		$data = unpack( 'H*', $binaryString );
+		return base_convert( array_shift( $data ), 16, 36 );
+	}
+
+	public function fromReadableToBinary( string $readableString ): string {
+		$hexData = base_convert( $readableString, 36, 16 );
+		return pack( 'H*', $hexData );
+	}
+}
--- a/src/UseCases/AddSubscription/AddSubscriptionUseCase.php
+++ b/src/UseCases/AddSubscription/AddSubscriptionUseCase.php
@@ -5,6 +5,7 @@ namespace WMDE\Fundraising\Frontend\UseCases\AddSubscription;
 
 use WMDE\Fundraising\Entities\Address;
 use WMDE\Fundraising\Entities\Subscription;
+use WMDE\Fundraising\Frontend\ConfirmationCodeConverter;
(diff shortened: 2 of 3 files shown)


src/ConfirmationCodeConverter.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+
```
JeroenDeDauw: Double spacing OMG

doublespace

src/ConfirmationCodeConverter.php
```diff
@@ -0,0 +1,21 @@
...
+namespace WMDE\Fundraising\Frontend;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ConfirmationCodeConverter {
```
JeroenDeDauw: Definitely belongs in DataAccess

src/UseCases/AddSubscription/AddSubscriptionUseCase.php
```diff
@@ -45,7 +46,12 @@ public function addSubscription( SubscriptionRequest $subscriptionRequest ) {
 		$this->subscriptionRepository->storeSubscription( $subscription );
 
 		$postalAddress = $subscription->getAddress();
-		$this->message->setTemplateParams( [ 'subscription' => $subscription ] );
+		$confirmationCodeConverter = new ConfirmationCodeConverter();
+		$confirmationCode = $confirmationCodeConverter->fromBinaryToReadable( $subscription->getConfirmationCode() );
```
JeroenDeDauw: Like this details from the database schema leak out from the data access layer, skip the domain layer and end up in the application layer. It'd be much nicer if this was taken care of by the persistence service.

src/ConfirmationCodeConverter.php
addshore: Oh these comments... :P

KaiNissen:
`$converter->fromBinaryToReadable( $converter->fromReadableToBinary( $string ) )` fails for long strings.

Actually the confirmation code was meant to be the hex representation of a 16 byte binary (as implemented in https://github.com/wmde/FundraisingFrontend/commit/fc4ac2ede15805270cf4b8140721bbffcacb6a59#diff-66bc72bd2239f66f5767ab1580a52756R56), so wouldn't `hex2bin()` and `bin2hex()` be sufficient?

wmde-manicki:
Jeroen's cat brought me here.
Just wanted to say that it is cool that PHP 7 now has this `random_bytes` function so there is no longer needed to use third-party library for those confirmation codes/uuids that old code uses.

gbirke:
Ok, if one of you merges https://github.com/wmde/FundraisingStore/pull/44, this PR can be dropped.

KaiNissen:
dropped in favor of wmde/FundraisingStore/pull/44

PR #97, opened 2016-02-03
Add GreetingGenerator

gbirke:
The class generates greetings based on lastname, salutation and title.
This is needed for all mails that are sent to users.

### DIFF
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -352,7 +352,7 @@ class FunFunFactory {
 			$this->getMessenger(),
 			new TwigTemplate(
 				$this->getTwig(),
-				'AddSubscriptionMailExternal.twig',
+				'Mail_Subscription_Request.twig',
 				[ 'greeting_generator' => $this->getGreetingGenerator() ]
 			),
 			$this->getTranslator()->trans( 'Your membership with Wikimedia Germany' )
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -33,6 +33,7 @@ use WMDE\Fundraising\Frontend\Presentation\Presenters\CommentListHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\CommentListJsonPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\CommentListRssPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Content\WikiContentProvider;
+use WMDE\Fundraising\Frontend\Presentation\GreetingGenerator;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\AddSubscriptionHTMLPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\AddSubscriptionJSONPresenter;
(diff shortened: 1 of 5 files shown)


wmde-manicki:
I don't know if this is relevant but it might be worth keeping in mind that there might be different types of greeting text. See for instance: https://github.com/wmde/fundraising/issues/776 where at least in some cases "Liebe" was used instead of "Sehr geehrte".
That particular case got apparently solved by adjusting the CMS wiki pages but I don't know is this the way it will/should be continued etc. So I am just leaving this as a random comment. Have no idea how this could be covered here.

wmde-manicki:
Oh, and also there is this who-knows-when-coming-true idea of having donation website in English. That would I believe include translatable email contents.

src/Presenters/GreetingGenerator.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ *
+ * @todo Make this translatable
+ */
+class GreetingGenerator {
+
+	public function createGreeting( string $lastName, string $salutation, string $title ): string {
+		if ( empty( $lastName ) ) {
```
JeroenDeDauw: #98

gbirke:
@wmde-manicki For the formal/informal greeting texts we may create another class when needed.

The possible translatability and making it unit-testable of the greetings was the reason I put this in a class and not in a template. However, since this is so much presentation-related, it might move in the future. 

What's your take in this, @JeroenDeDauw, regular PHP class (with the need for unit tests and injection) or template?

wmde-manicki:
> For the formal/informal greeting texts we may create another class when needed.

Sure.
What I wanted to to point out that given there are quite some different emails sent from the app (I didn't really count but I guess that might even be around 10? different emails, per language). Currently they share greetings part, as far as I remember, ie. they either use the same or simply say nothing more than "Howdy" (which might be explicitly included in the template).
But this case with some membership-related email needing different greeting shows that things might not always be that easy. But the generator still should be capable to handle other cases.

To sum up: I am not against the change you suggested. And good that you already have thought about possible "internationalization" of those things. Old codebase would probably require a series of nasty hacks to have it done.

JeroenDeDauw:
> But this case with some membership-related email needing different greeting shows that things might not always be that easy. But the generator still should be capable to handle other cases.

I don't know what is or will be needed here, however two things come to mind when I read that:

* If there is an easy case and a difficult case, then keeping them apart, so that the easy case does not depend on the difficult one, might be good.
* Doing the simplest thing that works for the features currently being created and refactoring later if additional stuff is needed rarely hurts. (YAGNI basically)

wmde-manicki:
> - If there is an easy case and a difficult case, then keeping them apart, so that the easy case does not depend on the difficult one, might be good.
> - Doing the simplest thing that works for the features currently being created and refactoring later if additional stuff is needed rarely hurts. (YAGNI basically)

I agree with both points, there is nothing to be changed in the proposed solution from my perspective. Having thought about all those greetings again I realized that there might not really be that much of hassle when some changes are required. English version(s) might not even need a class doing some magic, template would probably all cases.
So +1

PR #101, opened 2016-02-03
Improve error handling code

JeroenDeDauw:


app/bootstrap.php
```diff
@@ -31,11 +31,9 @@
 		'line' => $e->getLine(),
 		'stack_trace' => $e->getTraceAsString()
 	] );
-	return new JsonResponse(
-		[
-			'message' => $e->getMessage(),
-			'code' => $code
-		],
+
```
JeroenDeDauw: @KaiNissen this thing works for all routes, so no need to do anything route specific

app/bootstrap.php
```diff
@@ -31,11 +31,9 @@
 		'line' => $e->getLine(),
 		'stack_trace' => $e->getTraceAsString()
 	] );
-	return new JsonResponse(
```
gbirke: I'll be merging this but will be providing a patch that checks the request for acceptable content types and returns either HTML or JSON.

PR #107, opened 2016-02-04
Introduce TemplateBasedMailer

JeroenDeDauw:
PR includes  #104, please merge that one first and then review only the second commit here.

--

This is an interface designed based on the needs of the existing users
of TemplatedMessage and Messeger. They now only require a single service,
no longer need to mutate state part of a field and are exposed only
to what they need.

Only refactored the AddSubscriptionUseCase so far. Same should happen for
GetInTouchUseCase, after which TemplatedMessage can presumably be removed.

How to create weird interfaces:

* my code needs some Foo service
* Let's create a Foo service object
* What makes sense for Foo to have, and how does it work for all users I can think of
* Foo created. Now let's see how I can use it in my code

How to create good interfaces:

* my code needs a Foo service
* it has these variables that it would give to the service
* lets create Foo with a method that takes those parameters

Note how you can finish your code in test driven manner without actually implementing Foo.

### DIFF
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -316,15 +316,15 @@ class FunFunFactory {
 		return new AddSubscriptionUseCase(
 			$this->newRequestRepository(),
 			$this->getRequestValidator(),
-			$this->getMessenger(),
-			$this->newAddSubscriptionMessage()
+			$this->newAddSubscriptionMailer()
 		);
 	}
 
-	private function newAddSubscriptionMessage(): TemplatedMessage {
-		return new TemplatedMessage(
-			'Ihre Mitgliedschaft bei Wikimedia Deutschland', // TODO make this translatable
-			new TwigTemplate( $this->getTwig(), 'AddSubscriptionMailExternal.twig' )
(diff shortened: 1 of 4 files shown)


gbirke:
Interesting. In #93, I implemented the mailing exactly like this. Whenl @KaiNissen proposed the `Message`interface to have consistent interfaces in #94, I liked his idea of having a consistent interface for messages instead of having a specialized mailer. I'm a bit confused on the interface part, maybe you can look into those PRs and enlighten us?

KaiNissen:
I came up with #94, because the GetInTouch use case would send two messages: one templated confirmation message and one that forwards the content of the form fields to the operator. My thought was to decouple the message and the messenger to be able to pass in one messenger that handles both the templated and the forwarding messages.

JeroenDeDauw:
Follow up at #116

@KaiNissen I see what you mean. I can't quite articulate how this went wrong though. The result of these changes is better interface segregation and no need to mutate state. The decoupling of Messeger from the templating stuff remained, and decoupling in general increased due to the better interface segregation.

PR #109, opened 2016-02-04
Fix subscription validation

gbirke:
Set default emopty string values for fields that might not be set in the POST data.
Remove duplicate setting of wikilogin.
Changes names of validated fields from English to German


JeroenDeDauw:
> from English to German

Correct to someone travelling back through time!

PR #110, opened 2016-02-04
Add translations

gbirke:
Add Symfony Translator class to factory and integrate it into templates.
Use translator for mail subjects.
Add German translation files for messages and validations.
Use translator in AddSubscriptionJSONPresenter and test if tranlsation
works in System-Level test.

### DIFF
--- a/app/translations/messages.de_DE.json
+++ b/app/translations/messages.de_DE.json
@@ -1,0 +1,4 @@
+{
+  "Your inquiry at Wikimedia": "Ihre Anfrage an Wikimedia",
+  "Your membership with Wikimedia Germany": "Ihre Mitgliedschaft bei Wikimedia Deutschland"
+}
--- a/app/translations/validations.de_DE.json
+++ b/app/translations/validations.de_DE.json
@@ -1,0 +1,6 @@
+{
+  "This field is required": "Dieses Feld ist ein Pflichtfeld",
+  "Address has wrong format": "Kein gültiges Email-Adressformat",
+  "Address is no valid email": "Keine gültige E-Mail-Adresse",
+  "MX record not found": "Domain der Mailadresse nicht gefunden"
+}
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -22,6 +22,7 @@ use Monolog\Logger;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Swift_MailTransport;
+use Symfony\Component\Translation\TranslatorInterface;
 use Twig_Environment;
 use WMDE\Fundraising\Entities\Donation;
 use WMDE\Fundraising\Frontend\DataAccess\DbalCommentRepository;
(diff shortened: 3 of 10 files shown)


src/FunFunFactory.php
```diff
@@ -135,17 +136,35 @@ private function newPimple(): \Pimple {
...
+			$translator->addResource( 'json', __DIR__ . '/../app/translations/validations.de_DE.json', $locale, 'validations' );
+			return $translator;
+		} );
+
 		$pimple['twig_factory'] = $pimple->share( function () {
 			return new TwigFactory( $this->config['twig'] );
 		} );
 
 		$pimple['twig'] = $pimple->share( function( $pimple ) {
+			/** @var TwigFactory $twigFactory */
```
JeroenDeDauw: The rest of the code avoids the issue you fixed here by creating a getter, such as the `getTranslator` you introduced here.

src/TranslationFactory.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Symfony\Component\Translation\Loader\JsonFileLoader;
+use Symfony\Component\Translation\Translator;
+use Symfony\Component\Translation\MessageSelector;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class TranslationFactory {
```
JeroenDeDauw: What motivated you to create this factory? (Wondering the same about `TwigFactory`)

src/TwigFactory.php
```diff
@@ -38,7 +40,9 @@ public function create( array $loaders ): Twig_Environment {
 			$options
 		);
 
-		$twig->addExtension( new Twig_Extension_StringLoader() );
+		foreach ($extensions as $ext ) {
```
JeroenDeDauw: missing space

app/translations/validations.de_DE.json
```diff
@@ -0,0 +1,6 @@
+{
+  "This field is required": "Dieses Feld ist ein Pflichtfeld",
```
JeroenDeDauw: Looks like these are not used yet

JeroenDeDauw:
General approach looks good

app/translations/validations.de_DE.json
gbirke: It's the message from `RequiredFieldValidator`

src/TranslationFactory.php
gbirke: I wanted the factories to be testable in cases where they do more than just create a new instance. 
Also, I did not want to make `FunFunFactory` bigger.

app/translations/messages.de_DE.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Your inquiry at Wikimdia": "Ihre Anfrage an Wikimedia",
```
wmde-manicki: I generally like the idea and I am not familiar with this Symfony Translator but two minor remarks (just for you to possibly consider, I do not have a strong opinion about those things):
 - this "dictionary format" of `source_message: translated_message` makes sense but my only concern would be that when there is a need to change a subject of an email, you'd probably need to change the source part in all message files not only the translation part (plus in places where the particular translation is called), so you don't end up with things like `"Your inquiry at Wikimedia": "Your email to Wikimedia"`. Mediawiki stuff uses some more abstract things as keys, e.g. it'd be `"inquiry-email-subject": "Your inquiry at Wikimedia"`. This approach has this drawback that you have to look up what is the actual message (ie. what `trans("inquiry-email-subject")` stands for). But changing translation then only means changing "values" in the message files. I don't know which approach would fit better this use case. There shouldn't not be that much translatable messages here, and changes haven't been often.
 - I wouldn't expect it makes much of difference in this fundraising app but why having locale-specific message files and not language-specific? I doubt there will be different transations for Swiss German locale, British English etc.

Oh, and there is an "e" missing in "Wikimedia".

gbirke: 1. I agree that when the real translation feature is implemented, we should exchange the english strings in the code against translation keys.
2. Just for consistency. The translator needs to be initialized with a full locale string, so I was using the same locale for the message file.

The missing e is consistent in both code and message file, but I'll correct it.

src/TranslationFactory.php
JeroenDeDauw: I don't see a real issue with making the main factory bigger. Splitting it forces you to make more things public, and I think it often makes things harder to find rather than easier.

Such tests can be nice though.

PR #115, opened 2016-02-05
Validate against word list

gbirke:
Validate subscription requests against a configurable white- and blacklist from the wiki. This code can be reused for checking donations and membership requests.

This is for https://phabricator.wikimedia.org/T125958 and https://phabricator.wikimedia.org/T123227

### DIFF
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -52,6 +52,7 @@ use WMDE\Fundraising\Frontend\UseCases\ListComments\ListCommentsUseCase;
 use WMDE\Fundraising\Frontend\UseCases\CheckIban\CheckIbanUseCase;
 use WMDE\Fundraising\Frontend\UseCases\GenerateIban\GenerateIbanUseCase;
 use WMDE\Fundraising\Frontend\UseCases\ValidateEmail\ValidateEmailUseCase;
+use WMDE\Fundraising\Frontend\Validation\TextPolicyValidator;
 use WMDE\Fundraising\Store\Factory as StoreFactory;
 use WMDE\Fundraising\Store\Installer;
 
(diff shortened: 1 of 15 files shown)


src/FunFunFactory.php
```diff
@@ -417,4 +418,24 @@ private function getTwigFactory(): TwigFactory {
...
+		$contentProvider = $this->newWikiContentProvider();
+		$textPolicyConfig = $this->config['text-policies'][$policyName];
+		$badWords = $this->loadWordsFromWiki( $contentProvider, $textPolicyConfig['badwords'] ?? '' );
+		$whiteWords = $this->loadWordsFromWiki( $contentProvider, $textPolicyConfig['whitewords'] ?? '' );
+		$policyValidator->addBadWordsFromArray( $badWords );
+		$policyValidator->addWhiteWordsFromArray( $whiteWords );
+		return $policyValidator;
+	}
+
+	private function loadWordsFromWiki( WikiContentProvider $contentProvider, string $pageName ): array {
```
JeroenDeDauw: This does not belong here

src/Validation/SubscriptionValidator.php
```diff
@@ -26,20 +30,52 @@ public function __construct( MailValidator $mailValidator ) {
 	 * @return bool
 	 */
 	public function validate( $instance ): bool {
-		$violations = [];
-		$requiredFieldValidator = new RequiredFieldValidator();
+		$violations = $this->getRequiredFieldViolations( $instance );
+		$violations[] = $this->validateField( $this->mailValidator, $instance->getEmail(), 'email');
+		$this->constraintViolations = array_filter( $violations );
+		return count( $this->constraintViolations ) == 0;
```
JeroenDeDauw: If this is an array for sure, then using `empty` is nicer

PR #123, opened 2016-02-08
Ensure string values for SubscriptionRequest

gbirke:
Added defaults when getting values from the HTTP request to ensure we're
always getting string values even when a field is missing or browsers
stop sending empty strings for empty fields.


JeroenDeDauw:
What, no TDD? :)

PR #125, opened 2016-02-08
Add subscription design

gbirke:
Some presentation design changes


JeroenDeDauw:
+1

btw, better rebase before you push to the remote. Now the diff here includes stuff already merged, which the reviewer might not realize

gbirke:
bildschirmfoto 2016-02-08 um 15 32 34

Somehow these old changes got into this branch, rebase doesn't work. What can I do?

wmde-manicki:
Maybe try `rebase -i HEAD~3` and dropping two commits that sort of "are already in master"?

gbirke:
Thanks!

KaiNissen:
The submit buttons should be on the right side and the labels should have a fixed width to avoid the large margin on the right. Also the design of the select element is different in other forms. Did one of our UXperts provide something for this form? If not, we can just merge this, I guess.

gbirke:
Adressed all the issues raised by @KaiNissen

KaiNissen:
moved the submit button to the right in the [CMwiki](https://backend.wikimedia.de/wiki/index.php?title=Web%3ASpendenseite-HK2013%2Frewrite%2FSubscription_Form&diff=7045&oldid=7043)

KaiNissen:
btw: 2c685fc removed the usage of the variable `isMobile`, which me and @JeroenDeDauw were looking for, didn't find and thus removed its assignment in 6f972f1.

PR #127, opened 2016-02-08
Confirm subscription

gbirke:
First steps towards the ["Confirm Subscription" use case](https://phabricator.wikimedia.org/T123045).

### DIFF
--- a/tests/Integration/Presentation/Presenters/ConfirmSubscriptionHtmlPresenterTest.php
+++ b/tests/Integration/Presentation/Presenters/ConfirmSubscriptionHtmlPresenterTest.php
@@ -1,0 +1,37 @@
+<?php
+
+
+namespace WMDE\Fundraising\Frontend\Tests\Integration\Presentation\Presenters;
+
+use WMDE\Fundraising\Frontend\Presentation\Presenters\ConfirmSubscriptionHtmlPresenter;
+use WMDE\Fundraising\Frontend\ResponseModel\ValidationResponse;
+use WMDE\Fundraising\Frontend\TwigTemplate;
+use WMDE\Fundraising\Frontend\Validation\ConstraintViolation;
+
+/**
+ * @covers WMDE\Fundraising\Frontend\Presentation\Presenters\ConfirmSubscriptionHtmlPresenter
+ *
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ConfirmSubscriptionHtmlPresenterTest extends \PHPUnit_Framework_TestCase {
+
+	public function testGivenSuccessResponse_templateIsRenderedWithoutMessages() {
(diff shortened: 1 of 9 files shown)


tests/Integration/Presentation/Presenters/ConfirmSubscriptionHtmlPresenterTest.php
```diff
@@ -0,0 +1,37 @@
...
+class ConfirmSubscriptionHtmlPresenterTest extends \PHPUnit_Framework_TestCase {
+
+	public function testGivenSuccessResponse_templateIsRenderedWithoutMessages() {
+		$twig = $this->getMockBuilder( TwigTemplate::class )->disableOriginalConstructor()->getMock();
+		$twig->expects( $this->once() )
+			->method( 'render' )
+			->with( [] );
+		$presenter = new ConfirmSubscriptionHtmlPresenter( $twig );
+		$presenter->present( ValidationResponse::newSuccessResponse() );
+	}
```
JeroenDeDauw: The way you wrote those tests, they are unit tests no?

What I did so far is use real twig and see certain HTML is present in the output. Not saying that approach is better, just pointing out a potential interesting difference.

tests/System/Routes/ConfirmSubscriptionRouteTest.php
```diff
@@ -0,0 +1,63 @@
...
+use WMDE\Fundraising\Frontend\Messenger;
+use Swift_NullTransport;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ConfirmSubscriptionRouteTest extends WebRouteTestCase {
+
+	// @codingStandardsIgnoreStart
```
JeroenDeDauw: I think we should just disable this thing. Too many false positives. And our IDEs highlight those occurrences anyway (with less false positives than PHPCS).

tests/System/Routes/ConfirmSubscriptionRouteTest.php
```diff
@@ -0,0 +1,63 @@
...
+	protected function onTestEnvironmentCreated( FunFunFactory $factory, array $config ) {
+		// @codingStandardsIgnoreEnd
+		$factory->setMessenger( new Messenger(
+			Swift_NullTransport::newInstance(),
+			$factory->getOperatorAddress()
+		) );
+	}
+
+	public function testGivenAnUnconfirmedSubscriptionRequest_successPageIsDisplayed() {
```
JeroenDeDauw: If you have this, then I'd also expect to have a `testGivenAConfirmedSubscriptionRequest_stuffHappens`. Or does that not make sense?

tests/System/Routes/ConfirmSubscriptionRouteTest.php
gbirke: Go ahead.

PR #131, opened 2016-02-09
Minor simplifications

JeroenDeDauw:


src/Validation/ConstraintViolation.php
```diff
@@ -45,10 +45,7 @@ public function setSource( string $source ) {
 		$this->source = $source;
 	}
 
-	/**
-	 * @return mixed|null
-	 */
-	public function getConstraint() {
+	public function getConstraint(): string {
```
KaiNissen: This method actually returns a validator object.

JeroenDeDauw: Ugh... so this means we do not have a single test that executes this code path?!

JeroenDeDauw:
Panic!!!!!!!!!!!!!!

KaiNissen:
This method is also never used.

JeroenDeDauw:
I'm killing it now

JeroenDeDauw:
See #132

PR #132, opened 2016-02-09
Remove unused method

JeroenDeDauw:


gbirke:
I put that method in to be able to filter the violations by validation type in the future (e.g. to be able to display different messages in the presenter), but so far there was no need. We'll probably think of something else when the need arises.

PR #134, opened 2016-02-09
More work on subscription confirmation

gbirke:
Create actual use case code.
Add method to SubscriptionRepository to fetch a Subscription by
confirmation code.

Checking if the subscription has the right status is not implemented
yet.

### DIFF
--- a/src/FunFunFactory.php
+++ b/src/FunFunFactory.php
@@ -41,6 +41,7 @@ use WMDE\Fundraising\Frontend\Presentation\Presenters\GetInTouchHTMLPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\IbanPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\InternalErrorHTMLPresenter;
 use WMDE\Fundraising\Frontend\UseCases\CancelDonation\CancelDonationUseCase;
+use WMDE\Fundraising\Frontend\UseCases\ConfirmSubscription\ConfirmSubscriptionUseCase;
 use WMDE\Fundraising\Frontend\Validation\GetInTouchValidator;
 use WMDE\Fundraising\Frontend\Validation\MailValidator;
 use WMDE\Fundraising\Frontend\Validation\SubscriptionDuplicateValidator;
@@ -244,7 +245,7 @@ class FunFunFactory {
 		return $this->pimple['comment_repository'];
 	}
 
-	private function getSubscriptionRepository(): SubscriptionRepository {
+	public function getSubscriptionRepository(): SubscriptionRepository {
 		return $this->pimple['subscription_repository'];
 	}
 
@@ -359,6 +360,13 @@ class FunFunFactory {
 		);
 	}
 
+	public function newConfirmSubscriptionUseCase(): ConfirmSubscriptionUseCase {
+		return new ConfirmSubscriptionUseCase(
+			$this->getSubscriptionRepository(),
+			$this->newConfirmSubscriptionMailer()
+		);
+	}
+
 	private function newAddSubscriptionMailer(): TemplateBasedMailer {
 		return new TemplateBasedMailer(
 			$this->getMessenger(),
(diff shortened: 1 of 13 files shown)


src/FunFunFactory.php
```diff
@@ -366,6 +374,18 @@ private function newAddSubscriptionMailer(): TemplateBasedMailer {
 		);
 	}
 
+	private function newConfirmSubscriptionMailer(): TemplateBasedMailer {
+		return new TemplateBasedMailer(
+				$this->getMessenger(),
```
JeroenDeDauw: Double indent OMG

hqdefault

src/DataAccess/DoctrineSubscriptionRepository.php
```diff
@@ -25,4 +25,10 @@ public function storeSubscription( Subscription $subscription ) {
 		$this->entityManager->persist( $subscription );
 	}
 
+	public function findByConfirmationCode( string $confirmationCode ) {
```
JeroenDeDauw: Return type hint (in doc form) please

src/UseCases/ConfirmSubscription/ConfirmSubscriptionUseCase.php
```diff
@@ -14,13 +18,27 @@ class ConfirmSubscriptionUseCase {
...
 	}
 
-	public function confirmSubscription( string $confirmationCode ) {
-		// TODO Look for subscription confirmation code
-		// TODO if found && state == unconfirmed -> change state and show success page
-		// TODO else show failure page
+	public function confirmSubscription( string $confirmationCode ): ValidationResponse {
+		$binaryConfirmationCode = hex2bin( $confirmationCode );
+		$subscription = $this->subscriptionRepository->findByConfirmationCode( $binaryConfirmationCode );
+		if ( ! $subscription ) {
```
JeroenDeDauw:
```php
if ( $subscription === null )
```

src/UseCases/ConfirmSubscription/ConfirmSubscriptionUseCase.php
```diff
@@ -14,13 +18,27 @@ class ConfirmSubscriptionUseCase {
...
 		$this->subscriptionRepository = $subscriptionRepository;
+		$this->mailer = $mailer;
 	}
 
-	public function confirmSubscription( string $confirmationCode ) {
-		// TODO Look for subscription confirmation code
-		// TODO if found && state == unconfirmed -> change state and show success page
-		// TODO else show failure page
+	public function confirmSubscription( string $confirmationCode ): ValidationResponse {
+		$binaryConfirmationCode = hex2bin( $confirmationCode );
```
JeroenDeDauw: Does this binary-ness not belong in the repo?

tests/Fixtures/SubscriptionRepositorySpy.php
```diff
@@ -26,4 +26,14 @@ public function getSubscriptions(): array {
 		return $this->subscriptions;
 	}
 
+	public function findByConfirmationCode( string $confirmationCode )
+	{
```
JeroenDeDauw: `) {`

Kinda lame we had to disable this sniff due to PHP7

tests/Fixtures/SubscriptionRepositorySpy.php
```diff
@@ -26,4 +26,14 @@ public function getSubscriptions(): array {
...
+		foreach( $this->subscriptions as $subscription ) {
+			if ( $subscription->getConfirmationCode() === $confirmationCode ) {
+				return $subscription;
+			}
+		}
```
JeroenDeDauw: You are creating a "fake" here, something like an `InMemorySubscriptionRepository`. Spies just record the calls made.

gbirke: Actually, I accidentally created a new "InMemorySubscriptionRepository" here, because i really need the result. Will see if I can use the DBAL repo instead or if InMemorySubscriptionRepository has to be revived.

JeroenDeDauw: Nothing against having a `InMemorySubscriptionRepository` (which does not need full implementation, but would need to throw exceptions for stuff not implemented), it should just not be done in the spy.

JeroenDeDauw:
Follow up: #135

PR #135, opened 2016-02-10
Refcator code in ConfirmSubscriptionUseCaseTest

JeroenDeDauw:
Reduced state on both class and function level and extracted
string into constant

Follow up to #134


gbirke:
So state in test classes is a bad thing? I always thought, setting up mocks in the setup methods was a good thing to reduce clutter and avoid the initialization (=code duplication) in each method.

PR #136, opened 2016-02-10
donation validator

KaiNissen:


src/Validation/CanValidateValueObject.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Validation;
+
+trait CanValidateValueObject {
```
JeroenDeDauw: What's the reasoning behind introducing this trait?

JeroenDeDauw:
Follow up: #137

src/Validation/CanValidateValueObject.php
KaiNissen: Uh, might as well just be a private method.

JeroenDeDauw:
+2 when follow up is merged

PR #145, opened 2016-02-11
Amount validation

KaiNissen:
This adds an amount validator and an amount policy validator (the latter marking a donation as moderated).

### DIFF
--- a/src/Validation/DonationValidator.php
+++ b/src/Validation/DonationValidator.php
@@ -14,15 +14,23 @@ class DonationValidator implements InstanceValidator {
 	private $nameValidator;
 	private $addressValidator;
 	private $mailValidator;
+	private $amountValidator;
+	private $amountPolicyValidator;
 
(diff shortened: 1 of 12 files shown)


src/Validation/DonationValidator.php
```diff
@@ -49,6 +64,12 @@ public function validate( $donation ): bool {
 		return empty( $this->constraintViolations );
 	}
 
+	public function needsModeration( Donation $donation ): bool {
+		// TODO: add TextPolicyValidator
+		$this->policyViolations[] = $this->amountPolicyValidator->validate( $donation->getAmount() );
```
JeroenDeDauw: I'm thinking that the interval parameter should probably be required. However this call does not have access to that info, as it is not in Donation. Any plan on how to solve that yet?

KaiNissen: Right, this is not in donation, but should be. See #147

PR #147, opened 2016-02-11
considering interval value when validating amount policy

KaiNissen:
add field `Donation::interval`, pass it to `AmountPolicyValidation`, and fix return value of `DonationValidator::needsModeration()`


JeroenDeDauw:
Can you also make the parameter required then?

KaiNissen:
see #148

PR #124, opened 2016-02-08
AddDonation use case

KaiNissen:

PR #149, opened 2016-02-11
Initial foundation for AddDonation usecase

JeroenDeDauw:
Based on, and replacement for, #124


JeroenDeDauw:
So far I've moved over the code from the old PR and based it on the new objects and validator on master. There still is behaviour without tests though, so probably good to address that before merging.

Feel free to pick this up, I will double check with everyone before doing that myself tomorrow.

gbirke:
+1

PR #152, opened 2016-02-12
Finishing touches for the subscription use cases

gbirke:
After testing in the browser some behaviors needed to be added: setting the "moderated" state and the "confirmed" state when confirming.

Added additional variables needed by the templates

### DIFF
--- a/src/Domain/SubscriptionRepository.php
+++ b/src/Domain/SubscriptionRepository.php
@@ -17,7 +17,6 @@ interface SubscriptionRepository {
 	 * If the subscription exists in the repository, update it.
 	 *
 	 * @param Subscription $subscription
-	 * @return mixed
 	 */
 	public function storeSubscription( Subscription $subscription );
 
--- a/src/Domain/SubscriptionRepository.php
+++ b/src/Domain/SubscriptionRepository.php
@@ -11,6 +11,14 @@ use WMDE\Fundraising\Entities\Subscription;
  */
 interface SubscriptionRepository {
 
+	/**
+	 * Add or update a subscription.
+	 *
+	 * If the subscription exists in the repository, update it.
+	 *
+	 * @param Subscription $subscription
+	 * @return mixed
+	 */
 	public function storeSubscription( Subscription $subscription );
 
 	/**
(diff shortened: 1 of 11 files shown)


src/Domain/SubscriptionRepository.php
```diff
@@ -11,6 +11,14 @@
  */
 interface SubscriptionRepository {
 
+	/**
+	 * Add or update a subscription.
+	 *
+	 * If the subscription exists in the repository, update it.
+	 *
+	 * @param Subscription $subscription
+	 * @return mixed
```
JeroenDeDauw: #153

We should also get to documenting exceptions for the repos. For these it's real important that we don't end up catching Doctrine ones in the domain or usecases.

tests/Fixtures/InMemorySubscriptionRepository.php
```diff
@@ -19,7 +19,11 @@ class InMemorySubscriptionRepository implements SubscriptionRepository {
 	private $subscriptions = [];
 
 	public function storeSubscription( Subscription $subscription ) {
-		$this->subscriptions[] = $subscription;
+		$subscriptionKey = array_search( $subscription, $this->subscriptions, true );
+		if ( $subscriptionKey === false ) {
+			$subscriptionKey = count( $this->subscriptions );
+		}
+		$this->subscriptions[$subscriptionKey] = $subscription;
```
JeroenDeDauw: Dear PHP... *sigh*

JeroenDeDauw:
+1

src/Domain/SubscriptionRepository.php
gbirke: That means the repositories should catch the implementation-specific exceptions and throw their own exceptions, right? Shall I do a PR that does that?

JeroenDeDauw: > That means the repositories should catch the implementation-specific exceptions and throw their own exceptions, right?

Yes. Since the implementations are on the far side of the data access boundary, and the repository interfaces are on the domain side.

JeroenDeDauw:
All good from my side. Would like @KaiNissen to have a look before merge

KaiNissen:
+1, please rebase

PR #155, opened 2016-02-12
determine tracking variables

KaiNissen:
This reproduces what the "old" application does when receiving tracking parameters. The pull request also contains a cascading mechanism to get the preferred value from different sources (in this case: cookie, post, get).

### DIFF
--- a/src/UseCases/AddDonation/AddDonationRequest.php
+++ b/src/UseCases/AddDonation/AddDonationRequest.php
@@ -29,7 +29,7 @@ class AddDonationRequest {
 	private $bankName = '';
 
 	# tracking
-	private $tracking = ''; # TODO: generated from piwik parameters
+	private $tracking = '';
 	private $source = ''; # TODO: generated from referer
 	private $totalImpressionCount = 0;
 	private $singleBannerImpressionCount = 0;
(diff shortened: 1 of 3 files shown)


src/UseCases/AddDonation/AddDonationRequest.php
```diff
@@ -174,4 +174,22 @@ public function setAmountFromString( string $value, string $locale = 'de_DE' ) {
 		);
 	}
 
+	public function getPreferredValue( array $values ) {
```
JeroenDeDauw: These methods don't belong in this value object. You can tell by them not accessing anything in here. They could just as well be static. I'll merge now and submit a follow up to make them static, so it' at least more clear they are not really part of this class.

src/UseCases/AddDonation/AddDonationRequest.php
```diff
@@ -174,4 +174,22 @@ public function setAmountFromString( string $value, string $locale = 'de_DE' ) {
...
+		foreach ( $values as $value ) {
+			if ( $value ) {
```
JeroenDeDauw: Explicit checks please

PR #156, opened 2016-02-12
Add RepositoryException

gbirke:
Add Exception class that converts low-level storage layer exceptions to
the domain.

The methods that just select data should not throw exceptions.

### DIFF
--- a/src/DataAccess/DbalSubscriptionRepository.php
+++ b/src/DataAccess/DbalSubscriptionRepository.php
@@ -3,8 +3,11 @@
 
 namespace WMDE\Fundraising\Frontend\DataAccess;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\ORMException;
 use WMDE\Fundraising\Entities\Subscription;
+use WMDE\Fundraising\Frontend\Domain\SubscriptionRepositoryException;
 use WMDE\Fundraising\Frontend\Domain\SubscriptionRepository;
 
 /**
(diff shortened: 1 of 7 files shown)


src/Domain/RepositoryException.php
```diff
@@ -0,0 +1,13 @@
...
+namespace WMDE\Fundraising\Frontend\Domain;
+
+/**
+ * Exception class that converts low-level storage layer exceptions to the domain.
+ *
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class RepositoryException extends \RuntimeException {
+	const CODE_DEFAULT = 0;
```
JeroenDeDauw: The default already is 0. And if you want to have a new explicit default for this class, then I think it's better to do that in this class (ie by means of constructor).

src/Domain/SubscriptionRepository.php
```diff
@@ -11,6 +11,11 @@
  */
 interface SubscriptionRepository {
 
+	/**
+	 * @param Subscription $subscription
+	 * @return mixed
+	 * @throws RepositoryException
```
JeroenDeDauw: This is certainly an improvement. However I think it's odd to have a single RepositoryException. This prevents any domain specific information from being added. Example: https://github.com/wmde/WikibaseDatabase/blob/master/src/QueryInterface.php#L18

I'd have not gone that far right off the bat, though still have added a single Exception per repository (or more generally speaking, per service).

src/Domain/RepositoryException.php
gbirke: ok, I'll create a constructor instead.

src/DataAccess/DbalSubscriptionRepository.php
```diff
@@ -19,22 +22,37 @@ public function __construct( EntityManager $entityManager ) {
...
+	 * @param Subscription $subscription
+	 * @throws SubscriptionRepositoryException
+	 */
 	public function storeSubscription( Subscription $subscription ) {
-		$this->entityManager->persist( $subscription );
-		$this->entityManager->persist( $subscription->getAddress() );
-		$this->entityManager->flush();
+		try {
+			$this->entityManager->persist( $subscription );
+			$this->entityManager->persist( $subscription->getAddress() );
```
JeroenDeDauw: As far as I can tell, the `persist` methods never throw an `ORMException`. They do throw some thing driving from `InvalidArgumentException`. .... ..... ok, it can throw this, the Doctine docs are just fail. Gonna submit a PR there.

tests/Integration/DataAccess/DbalSubscriptionRepositoryTest.php
```diff
@@ -60,6 +61,19 @@ public function testMultipleSubscriptions_onlySimilarAreCounted() {
...
+		$this->expectException( SubscriptionRepositoryException::class );
+		$entityManager = $this->getMock(
+			EntityManager::class, array('getRepository', 'getClassMetadata', 'persist', 'flush'), array(), '', false
+		);
+		$entityManager->expects( $this->once() )->method( 'persist' )->willThrowException( new ORMException() );
+		$repository = new DbalSubscriptionRepository(  $entityManager );
+		$subscription = new Subscription();
+		$subscription->setEmail( 'nyan@awesomecats.com' );
+		$subscription->setAddress( new Address() );
+		$repository->storeSubscription( $subscription );
```
JeroenDeDauw: #160

JeroenDeDauw:
Needs rebase

PR #164, opened 2016-02-15
add text policy validation

KaiNissen:


src/Validation/DonationValidator.php
```diff
@@ -54,13 +57,61 @@ public function validate( Donation $donation ): ValidationResult {
...
+				$fieldTextValidator->validate( $personalInfo->getPhysicalAddress()->getPostalCode() ),
+				'plz'
+			);
+			$violations[] = $this->getFieldViolation(
+				$fieldTextValidator->validate( $personalInfo->getPhysicalAddress()->getCity() ),
+				'ort'
+			);
+		}
+
+		return $violations;
```
JeroenDeDauw: If you do `return [ ... ]`, you can avoid having the local var and all the assignments to it.

src/Validation/DonationValidator.php
```diff
@@ -54,13 +57,61 @@ public function validate( Donation $donation ): ValidationResult {
...
+			$violations,
+			$this->getBadWordViolations( $donation )
+		);
+
 		$this->policyViolations = array_filter( $violations );
+
 		return !empty( $this->policyViolations );
 	}
 
+	public function getPolicyViolations() {
```
JeroenDeDauw: A type hint here would be nice (in the docs `@return ConstraintViolation[]`, and `:array` in the code)

PR #165, opened 2016-02-15
Simplify DisplayPageUseCase

gbirke:
Instead of pre-rendering the template, use the normal Twig template
loading mechanism and validate via TemplateNameValidator.

This enables testing complex pages by creating them temporarily in the
`app/templates` folder  before uploading them into the CMS wiki.

It'll improve caching, since the previously used extension interfered
with the Twig caching mechanism.

### DIFF
--- a/src/UseCases/DisplayPage/PageDisplayResponse.php
+++ b/src/UseCases/DisplayPage/PageDisplayResponse.php
@@ -13,15 +13,29 @@ use WMDE\Fundraising\Frontend\FreezableValueObject;
 class PageDisplayResponse {
 	use FreezableValueObject;
 
-	private $mainContent;
+	private $mainContentTemplate;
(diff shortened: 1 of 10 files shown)


src/UseCases/DisplayPage/PageDisplayResponse.php
```diff
@@ -13,15 +13,29 @@
...
+	private $templateExists;
 
-	public function getMainContent(): string {
-		return $this->mainContent;
+	public function getMainContentTemplate(): string {
+		return $this->mainContentTemplate;
 	}
 
-	public function setMainContent( string $mainContent ) {
+	public function setMainContentTemplate( string $mainContentTemplate ): PageDisplayResponse {
```
JeroenDeDauw: `: self`

PR #167, opened 2016-02-16
Cleanup JavaScript

gbirke:
Move Membership- and Lightbox-specific JavaScript to their own files.
Remove JS code which is not used in current form.

membership.js is not included at the moment, we'll look at it when
implementing the membership use case.

### DIFF
--- a/app/templates/BaseLayout.twig
+++ b/app/templates/BaseLayout.twig
@@ -32,6 +32,7 @@
     <script type="text/javascript" src="{$ basepath $}/res/js/lib/jquery.ui.touch-punch-0.2.3.min.js"></script>
     <script type="text/javascript" src="{$ basepath $}/res/js/lib/wlightbox.jquery.js"></script>
     <script type="text/javascript" src="{$ basepath $}/skin/10h16/_js/spenden.wikimedia.js"></script>
+    <script type="text/javascript" src="{$ basepath $}/skin/10h16/_js/lightbox_init.js"></script>
 
     <script type="text/javascript" src="{$ basepath $}/res/js/lib/jquery.als-1.7.min.js"></script>
     <script type="text/javascript" src="{$ basepath $}/res/js/wmde.js"></script>
--- a/web/skin/10h16/_js/lightbox_init.js
+++ b/web/skin/10h16/_js/lightbox_init.js
@@ -1,0 +1,76 @@
+$(initWLightbox);
+
+/* wlightbox inline */
+function initWLightbox() {
+    $('.rtcol, footer').each(function () {
+        var $container = $(this);
+
+        if ($container.find('a.wlightbox').length < 1) return true;
+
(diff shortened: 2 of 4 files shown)


JeroenDeDauw:
+1

JeroenDeDauw:
Would be good to have some smoke tests for the JS. (Guess this is covered by having browser tests)

And if we write any non totally tirivial JS ourselves, adding QUnit (or other) tests for that would also be nice.

wmde-manicki:
Generally I consider cleaning up and splitting the old JS code a very good idea. But as far as I remember my relationship with that thing we (using "we" here not because I am too afraid to sign below this statement but I think some other(s) shared it) had an impression that the JS code is so tangled, mixing handling different steps of donation and membership forms in a single place, and all other nasty things, that it seemed reasonable to consider throwing the old code away and rewriting the client-side part of form handling from scratch. At least I remember finding it difficult to separate something from the code, so all other things wouldn't suddenly break. The structure of the code (bunch of random functions spread around two big JS files) also affected (at least to some extent) the possibilities of adding QUnit test which is obviously a good idea!

I am not saying that there is something wrong in this pull request, it is definitely good. But it might be that it saves some hair on year heads if you forget about the existing JavaScript mess and add new shining and clean things as you progress with adding use cases to the new app etc.

But that said, I have no idea how having no client-side code would affect your work on the rewrite of the server part (how to reasonably test something in the browser when there is all JS part missing, etc). So it might be that my comment is entirely foobar.

KaiNissen:
I agree to Leszek's comment. The code is quite tangled and contains some stuff we actually want to get rid of (e. g. HTML5 form validation). In my opinion, separating the code by use case might not be the right approach, since the forms have quite a lot in common. You (or was it someone else) earlier proposed that we should separate it by functionality (form validation, DOM manipulation, client-server communication).

KaiNissen:
Oh, and while the change fixes quite a lot of the code style issues, it introduces others.

gbirke:
I totally agree with all of you. This is just a very small step on the road to better JavaScript. Mostly, I was trying to figure out what the code does and to make it a little bit more manageable, I broke it out into separate files.

Originally, Kai me and agreed to leave the this hairball as it is for the rewrite (since our progress looks horrible at the moment). New JS code will be written in a more modular, tested manner.

I could make the old code conform more to our JSCS checks, but then you wouldn't have seen what was actually changed. We can just close this for the time being instead.

wmde-manicki:
slightly cleaned mess > mess, I'd say, so I don't think this PR should be dropped by now.
Just wanted to mention that it might not be worth it to feel to attached to the "old" JS code, as the new app is going to do lot of things different. But as you two say, you already have it mind, so I guess enough talking for now.
Clean up started here surely helps!

gbirke:
Rebased

PR #168, opened 2016-02-16
Bank data validation

KaiNissen:


src/Validation/DonationValidator.php
```diff
@@ -27,13 +28,16 @@ public function __construct( AmountValidator $amountValidator,
 								 PersonNameValidator $nameValidator,
 								 PhysicalAddressValidator $addressValidator,
 								 AllowedValuesValidator $paymentTypeValidator,
+								 BankDataValidator $bankDataValidator,
 								 MailValidator $mailValidator ) {
+		// TODO: inject a ValidatorObject to reduce number of parameters
```
JeroenDeDauw: What do you mean by this?

tests/Unit/Validation/BankDataValidatorTest.php
```diff
@@ -0,0 +1,112 @@
...
+ * @license GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class BankDataValidatorTest extends ValidatorTestCase {
+
+	/**
+	 * @dataProvider invalidBankDataProvider
+	 * @param string $iban
+	 * @param string $bic
+	 * @param string $bankName
```
JeroenDeDauw: Please omit param docs for test methods with data providers. I know that recent PHPStorm bitches about this (it used to not do that, I already filed an issue about it). The docs are out of date here btw :)

JeroenDeDauw:
+1

src/Validation/DonationValidator.php
KaiNissen: I'm unhappy about how the list of expected parameters to create the use case object increases and thought about injecting a single container for all those validators.

PR #169, opened 2016-02-16
generalizing referring domains

KaiNissen:
This generalizes the referring URLs to contain the host name at most. The result of it needs to be stored as a tracking parameter when users enter the site.

### DIFF
--- a/tests/Unit/GeneralizedReferrerTest.php
+++ b/tests/Unit/GeneralizedReferrerTest.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace WMDE\Fundraising\Tests\Unit;
+
+use WMDE\Fundraising\Frontend\GeneralizedReferrer;
+use WMDE\Fundraising\Frontend\Tests\Unit\Validation\ValidatorTestCase;
+
+/**
+ * @covers WMDE\Fundraising\Frontend\GeneralizedReferrer
+ *
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class GeneralizedReferrerTest extends ValidatorTestCase {
+
+	private $domainMap = [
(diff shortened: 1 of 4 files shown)


tests/Unit/GeneralizedReferrerTest.php
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	/**
+	 * @dataProvider urlProvider
+	 * @param string $expected
+	 * @param string $url
+	 */
+	public function testGeneralization( string $url, string $expected ) {
+		$generalizer = new GeneralizedReferrer( 'web', $this->domainMap );
+		$this->assertEquals( $expected, $generalizer->generalize( $url ) );
```
JeroenDeDauw: assertSame is better when it are two scalars

tests/Unit/GeneralizedReferrerTest.php
```diff
@@ -0,0 +1,50 @@
...
+		'www.wikimedia.de' => 'wikimedia.de',
+		'wikipedia.de' => 'wikipedia.de',
+		'www.wikipedia.de' => 'wikipedia.de',
+		'de.wikipedia.org' => 'de.wikipedia.org',
+		'en.wikipedia.org' => 'en.wikipedia.org',
+		'ru.wikivoyage.org' => 'ru.wikivoyage.org',
+	];
+
+	public function urlProvider() {
+		return [
```
JeroenDeDauw: Personally I think it makes more sense to put the provider after the test method that uses it

JeroenDeDauw:
Looks good so far

KaiNissen:
Although there is no caller/user of this, it can already be merged, I guess.

JeroenDeDauw:
+1. @gbirke presumably knows more about this stuff than me

PR #170, opened 2016-02-16
One-page form

gbirke:
First steps towards a validated one-page form.

Using modern JavaScript modules and tests.

### DIFF
--- a/app/js/lib/form_pages.js
+++ b/app/js/lib/form_pages.js
@@ -1,4 +1,6 @@
-var FormPage = {
+var Promise = require( 'promise' ),
+
+	FormPage = {
 		show: function () {
 			jQuery( this.sections ).show();
 		},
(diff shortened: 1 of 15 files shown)


app/js/lib/form_pages.js
```diff
@@ -0,0 +1,89 @@
...
+	FormPage = {
+		show: function () {
+			jQuery( this.sections ).show();
+		},
+
+		hide: function () {
+			jQuery( this.sections ).hide();
+		},
+		validate: function () {
+			'use strict';
```
JeroenDeDauw: Why not do this at the top level?

JeroenDeDauw:
Had a look at all the things. +1 (not +2, as I think the @KaiNissen should merge)

app/js/lib/form_pages.js
gbirke: There was a point where one of the code quality tools did complain and I found some some explanation that putting it at the beginning of the file is not good style because when the files are concatenated the there are all these "use strict" statements in the middle of the code. However, now that I checked, the tools don't complain any more. and Browserify does something more intelligent than just concatenating. I'll rework this.

app/js/tests/test_form_pages.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+var test = require( 'tape' ),
+	formPages = require( '../lib/form_pages' ),
+	Promise = require( 'promise' ),
+	pageSpy = {
+		display: false,
```
KaiNissen: Just a minor thing: I find this variable name misleading and personally would prefer something like `displayed` or `visible`. Will merge this now, though.

JeroenDeDauw: I had the same though - I first saw it being used somewhere, and `display` seems like a method name. `isDisplayed` or `isVisible` are much nicer

PR #176, opened 2016-02-17
Make unit tests finish faster

JeroenDeDauw:
With this config PHPUnit will first run the unit tests, then the integration
tests and then the system tests. This makes so that you likely need to wait
a lot less long before the test you are interested in has run.


addshore:
+1

PR #177, opened 2016-02-17
Use short array syntax everywhere

JeroenDeDauw:
Used https://github.com/FriendsOfPHP/PHP-CS-Fixer and ran

    php php-cs-fixer.phar fix src/ --fixers=short_array_syntax
    php php-cs-fixer.phar fix tests/ --fixers=short_array_syntax


addshore:
+1

PR #179, opened 2016-02-18
disable xdebug extension on travis

KaiNissen:


JeroenDeDauw:
huh? why? don't we get less fancy errors now?

PR #175, opened 2016-02-17
Add payment type specific behaviour to AmountValidator

JeroenDeDauw:

PR #181, opened 2016-02-18
class PaymentType instead of interface

KaiNissen:
follow-up to #175

### DIFF
--- a/src/Domain/Model/PaymentType.php
+++ b/src/Domain/Model/PaymentType.php
@@ -7,12 +7,17 @@ namespace WMDE\Fundraising\Frontend\Domain\Model;
 /**
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
  */
-interface PaymentType {
+final class PaymentType {
 
 	const BANK_TRANSFER = 'UEB';
 	const CREDIT_CARD = 'MCP';
 	const DIRECT_DEBIT = 'BEZ';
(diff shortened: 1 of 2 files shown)


src/Domain/Model/PaymentType.php
```diff
@@ -17,4 +18,8 @@
 	const CASH = 'CASH';
 	const BITCOIN = 'BTC';
 
+	public static function getPaymentTypes() {
+		return ( new \ReflectionClass( self::class ) )->getConstants();
```
JeroenDeDauw: reflection

An actually non-evil looking usage of Reflection. Now I've seen it all.

Typical PHP... have `get_class_vars` and `get_class_methods`, but no `get_class_constants`

src/FunFunFactory.php
```diff
@@ -195,16 +194,6 @@ private function newPimple(): \Pimple {
 			] );
 		} );
 
-		// In the future, this could be locale-specific or filled from a DB table
-		$pimple['payment_types'] = $pimple->share( function() {
-			return new PaymentTypes( [
```
JeroenDeDauw: Is PaymentTypes not unused now?

src/Domain/Model/PaymentType.php
```diff
@@ -17,4 +18,8 @@
 	const CASH = 'CASH';
 	const BITCOIN = 'BTC';
 
+	public static function getPaymentTypes() {
```
JeroenDeDauw: : array

src/FunFunFactory.php
KaiNissen: Yes, did that on a different branch, rebased now.

PR #182, opened 2016-02-18
Persist donation

KaiNissen:


src/Domain/DoctrineDonationRepository.php
```diff
@@ -0,0 +1,26 @@
...
+
+	private $entityManager;
+
+	public function __construct( Connection $connection ) {
+		$factory = new FundraisingStoreFactory( $connection );
+		$this->entityManager = $factory->getEntityManager();
+	}
+
+	public function storeDonation( Donation $donation ) {
+		$this->entityManager->persist( $donation );
```
JeroenDeDauw: This is broken... nothing using it though

src/FunFunFactory.php
```diff
@@ -576,7 +576,8 @@ public function newAddDonationUseCase(): AddDonationUseCase {
 				$this->newPaymentTypeValidator(),
 				$this->newBankDataValidator(),
 				$this->getMailValidator()
-			)
+			),
+			$this->newGeneralizedReferrer()
```
JeroenDeDauw: This is a bad class name btw: adjective + noun. This would be better: `ReffererGeneralizer`

JeroenDeDauw:
Follow up: #183

src/Domain/Donation.php
```diff
@@ -69,4 +109,98 @@ public function setBankData( BankData $bankData ) {
...
+
+	public function determineFullName() {
+		if ( $this->getPersonalInfo() !== null ) {
+			return $this->getPersonalInfo()->getPersonName()->getFullName();
+		}
+
+		return 'Anonym';
+	}
+
+	public function generateTransferCode() {
```
JeroenDeDauw: Where do you intend to use this? This code appears to have no cohesion with the Donation class, which suggests it does not belong here. Can't say without knowing what this is intended for though.

src/Domain/Donation.php
```diff
@@ -13,11 +14,41 @@
...
 	private $personalInfo;
+	private $optIn;
 	private $bankData;
+	private $tracking;
+	private $source;
+	private $totalImpressionCount;
+	private $singleBannerImpressionCount;
+	private $color;
+	private $skin;
+	private $layout;
```
JeroenDeDauw: Would this all fit in a TrackingData object? It is all tracking stuff right, or is some of it not?

src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -59,4 +73,68 @@ private function newBankDataFromRequest( AddDonationRequest $request ): BankData
 		return $bankData->freeze()->assertNoNullFields();
 	}
 
+	private function newDonationEntityFromDonationDomain( Donation $donation ) {
+		$dbalDonation = new \WMDE\Fundraising\Entities\Donation();
```
JeroenDeDauw: This is misplaced. Fixed in #183

The idea is to not have anything that is a persistence implementation detail, such as the Store component or Doctine, outside of the DataAccess folder. (Well, also in the factory and tests possible, but certainly not in the Domain folder.) Not how the repository interfaces are in the Domain folder and the implementations are in DataAccess.

tests/System/Routes/AddDonationRouteTest.php
```diff
@@ -0,0 +1,101 @@
...
+		'layout' => 'Default',
+		'color' => 'blue',
+		'skin' => 'default',
+	];
+
+	public function testGivenValidRequest_donationGetsPersisted() {
+		$donationRepository = new DonationRepositorySpy();
+
+		$client = $this->createClient( [], function( FunFunFactory $factory ) use ( $donationRepository ) {
+			$factory->setDonationRepository( $donationRepository );
```
JeroenDeDauw: The responsibility of the system tests is to verify the whole stack works, so it's generally better to not mock out the database. Changed in #183

src/DataAccess/DoctrineDonationRepository.php
```diff
@@ -0,0 +1,99 @@
...
+use WMDE\Fundraising\Frontend\Domain\Model\PaymentType;
+use WMDE\Fundraising\Frontend\Domain\Repositories\DonationRepository;
+use WMDE\Fundraising\Entities\Donation as DoctrineDonation;
+use WMDE\Fundraising\Frontend\Domain\Donation;
+
+/**
+ * @license GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DoctrineDonationRepository implements DonationRepository {
```
gbirke: Could you rename it to DbalDonationRepository? No need to encode the library name in the class.

src/Domain/Donation.php
```diff
@@ -13,11 +14,41 @@
...
+
+	const STATUS_MODERATION = 'P';
+	const STATUS_DELETED = 'D';
+
+	private $status;
 	private $amount;
 	private $interval = 0;
 	private $paymentType;
 	private $personalInfo;
+	private $optIn;
```
gbirke: This is true when the user has opted in to receive more infos, right? or is it for othe opt-in processses? Maybe a more descriptive name?

gbirke:
+1, apart from some name nagging that could be ignored.

src/DataAccess/DoctrineDonationRepository.php
JeroenDeDauw: Dbal is also a library :) Actually I think our other ones should be renamed to Doctine, as they use Doctine ORM, not just DBAL

src/Domain/Donation.php
JeroenDeDauw: I'm going to split this off in a follow up, hoping I'm guessing correctly

src/Domain/Donation.php
JeroenDeDauw: Yeah, not very descriptive. This indeed is the `info` HTTP request parameter and `info` db field. So this should be a boolean, not a string

src/Domain/Donation.php
KaiNissen: moved it to FundraisingStore (see wmde/FundraisingStore#52)

src/Domain/Donation.php
KaiNissen: yes, you totally guessed right

PR #186, opened 2016-02-20
Work on Add Donation UC

JeroenDeDauw:


JeroenDeDauw:
Beware that this PR moves a bunch of stuff around. So if this sits here and you do new stuff based off from master, we'll have a lot of merging fun :)

PR #187, opened 2016-02-20
moved transfer code generator to FundraisingStore

KaiNissen:
depends on wmde/FundraisingStore#52


JeroenDeDauw:
Rebased

KaiNissen:
different approach in #193

PR #191, opened 2016-02-22
Fix mail transport error in add donation route test

gbirke:
Skip route test when kontocheck library is not installed.


JeroenDeDauw:
Thanks!

PR #194, opened 2016-02-22
Validate amount

gbirke:
Work on https://phabricator.wikimedia.org/T127001

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -25,6 +25,7 @@ use WMDE\Fundraising\Frontend\UseCases\GetInTouch\GetInTouchRequest;
 use WMDE\Fundraising\Frontend\UseCases\CancelDonation\CancelDonationRequest;
 use WMDE\Fundraising\Frontend\UseCases\ListComments\CommentListingRequest;
 use WMDE\Fundraising\Frontend\UseCases\AddSubscription\SubscriptionRequest;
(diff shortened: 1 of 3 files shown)


src/UseCases/ValidateAmount/ValidateAmountUseCase.php
```diff
@@ -0,0 +1,30 @@
...
+namespace WMDE\Fundraising\Frontend\UseCases\ValidateAmount;
+
+use WMDE\Fundraising\Frontend\Validation\AmountValidator;
+use WMDE\Fundraising\Frontend\ResponseModel\ValidationResponse;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ValidateAmountUseCase {
```
JeroenDeDauw: This usecase is somewhat of an empty partition. What do you think about omitting it in this case, by using the AmountValidator directly in the route?

gbirke: Deviation from the architecture? :open_mouth:
I can certainly do that.

JeroenDeDauw: I actually gave this exact case as an example of where to deviate to Kai... kinda facepalm now at the communication

PR #195, opened 2016-02-22
Donation mail

KaiNissen:


src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -91,11 +97,32 @@ private function newTrackingInfoFromRequest( AddDonationRequest $request ): Trac
...
+					'recipient' => [
+						'salutation' => ( new GreetingGenerator() )->createGreeting(
+							$donation->getPersonalInfo()->getPersonName()->getLastName(),
+							$donation->getPersonalInfo()->getPersonName()->getSalutation(),
+							$donation->getPersonalInfo()->getPersonName()->getTitle()
+						)
+					],
+					'donation' => [
+						'id' => $donationId,
+						'formattedAmount' => ( new \NumberFormatter( 'de_DE', \NumberFormatter::CURRENCY ) )->format(
```
JeroenDeDauw: I suspect the formatting is better done in the template, as this is presentation related. Have a look at `CommentList.rss.twig` for an example of number formatting in template.

src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -91,11 +97,32 @@ private function newTrackingInfoFromRequest( AddDonationRequest $request ): Trac
...
+							$donation->getAmount()
+						),
+						'needsModeration' => $needsModeration,
+						'paymentType' => [
+							'code' => $donation->getPaymentType(),
+							'text' => 'Überweisung' // TODO: use twig translator extension
+						],
+						'bankTransferCode' => 'W-Q-ABCDEF-G', // $donation->getBankTransferCode(), should this be fetched from entity?
```
JeroenDeDauw: With my follow up to your other PR, you can simply get it from the domain object.

src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -58,9 +59,14 @@ public function addDonation( AddDonationRequest $donationRequest ) {
...
-		$this->donationRepository->storeDonation( $donation );
+		$needsModeration = $this->donationValidator->needsModeration( $donation );
+		if ( $needsModeration ) {
+			$donation->setStatus( Donation::STATUS_MODERATION );
+		}
+
+		$donationId = $this->donationRepository->storeDonation( $donation );
 
-		$this->sendDonationConfirmationEmail( $donation );
+		$this->sendDonationConfirmationEmail( $donationId, $donation, $needsModeration );
```
JeroenDeDauw: Ah... this is silly. The Donation domain object should have the id. Which makes it clear the comment that was in the repository was not quite accurate, as it needs to update the passed donation object, not return the id. (That can be changed in another commit though (kinda unrelated to what you are doing here), and I'm happy to do this.)

JeroenDeDauw:
I'm rebasing this now

src/UseCases/AddDonation/AddDonationUseCase.php
KaiNissen: Since twig's `number_format()` does not support locales, I decided to add the intl extension and change the template to use that.

JeroenDeDauw:
+1 looks good to me, though someone with more knowledge of the domain and attention to detail should probably also look at it

app/templates/Mail_Donation_Confirmation.twig
```diff
@@ -1 +1,90 @@
...
+{% else %}
+{% if donation.paymentType == 'BEZ' %}
+In Kürze erhalten Sie eine Vorankündigung Ihres SEPA-Lastschriftmandats. Dort ist auch Ihre Mandatsreferenznummer enthalten.
+Ihre Spende in Höhe von {$ formattedAmount $} werden wir im Anschluss von dem von Ihnen angegebenen Konto abbuchen.
+{% if donation.interval > 0 %}
+{$ 'recurringDonation'|transchoice( donation.interval, { '%interval%': donation.interval } ) $}
+{% endif %}
+
+{% elseif donation.paymentType == 'PPL' or donation.paymentType == 'MCP' %}
+Ihre Spende in Höhe von {$ formattedAmount $} wurde durch unseren Partner {$ donation.paymentType|trans $} gebucht.
```
gbirke: This should be `{$ donation.paymentType|trans({}, 'paymentTypes') $}` otherwise the payment type will be looked up in the default messages file.

gbirke:
The new paymentTypes file must be added in the FunFunFactory.

KaiNissen:
You mean like [this](https://github.com/wmde/FundraisingFrontend/blob/donation_mail/src/FunFunFactory.php#L191)?

KaiNissen:
Ah, I see what you mean: I'm going to put this into a specific domain context.

JeroenDeDauw:
I'm rebasing this now...

PR #200, opened 2016-02-23
Refactor One-page form

gbirke:
Added all the missing fields to the second page HTML (not validated yet).
Use [Redux](http://redux.js.org/index.html) for storing form state.
Move HTML/CSS changes into testable viewHandler objects.
Getting values from one or more form elements with jquery is now encapsulated with a common interface and the only part that needs jsdom for testing. 
Display correct step in the form steps.


JeroenDeDauw:
Had a look at Redux. Interesting library. Found this also interesting https://github.com/reactjs/redux/issues/351

app/js/lib/reducers/form_pagination.js
```diff
@@ -0,0 +1,31 @@
...
+			break;
+		case 'NEXT_PAGE':
+			return pageLength > defaultPage + 1 ? defaultPage + 1 : defaultPage ;
+		default:
+			return defaultPage;
+	}
+}
+
+function formPagination( pageState, action ) {
+	var defaultPageState = pageState || {},
```
adrianheine: Misleading variable name

app/js/lib/reducers/form_pagination.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+var formPage = require( './form_pages' );
+
+function currentPage( page, pageLength, action ) {
+	var defaultPage = page || ( pageLength > 0 ? 0 : -1 );
```
adrianheine: Misleading variable name

adrianheine: Will do the wrong thing if `page` is `0`. Please add tests for `0`, `-1`, `1` cases when doing things like these.

JeroenDeDauw:
+1

app/js/lib/actions.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+function newAddPageAction( pageName ) {
+	return {
+		type: 'ADD_PAGE',
+		payload: { name: pageName }
```
inyono: Just a side note: you don't have to group the additional information into `payload`. This would also work perfectly fine:
```.js
function newAddPageAction( pageName ) {
  return {
    type: 'ADD_PAGE',
    name: pageName
  }
}
```
(with the appropriate changes in the reducer).

gbirke: It was like this at first but then I read about [Flux standard actions](https://github.com/acdlite/flux-standard-action) which also [exist for Reflux](https://github.com/acdlite/redux-actions).
Using the payload field was a conscious design decision.

app/js/lib/reducers/validity.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+function validity( state, action ) {
+	if ( typeof state === 'undefined' ) {
+		return {
```
inyono: You shouldn't `return` here usually. In this case, the reducer does not work if `state` is still undefined, but you want to handle some `action`. I think that this is no problem if you use Redux correctly, since the state will be initialised beforehand anyway. It is rather the usual practice to do so.

app/js/lib/view_handler/enabled_when_valid.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var objectAsssign = require( 'object-assign' ),
+	EnabledWhenValidHandler = {
+		elements: null,
+		update: function ( state ) {
+			if ( typeof state  !== 'undefined' && state.isValid && state.isValidated ) {
```
inyono: I don't think you need to check here whether the state is `undefined`. Because usually, you will invoke the root reducer at least once, and then all the subreducers are invoked and initialise the state (since you always handle `typeof state === 'undefined'` in the reducers)

inyono:
+1, my comments are basically side notes :)

JeroenDeDauw:
Ugh, one of the commits has [WIP] in its title :)

gbirke:
Fixed the WIP commit message. Merge plz?

PR #202, opened 2016-02-24
Add DonationRepository::getDonationById

JeroenDeDauw:


src/DataAccess/DoctrineDonationRepository.php
```diff
@@ -31,15 +33,17 @@ public function __construct( EntityManager $entityManager ) {
 	}
 
 	public function storeDonation( Donation $donation ) {
+		$doctrineDonation = $this->newDonationEntity( $donation );
+
 		try {
-			$this->entityManager->persist( $this->newDonationEntity( $donation ) );
+			$this->entityManager->persist($doctrineDonation  );
```
gbirke: missing space

JeroenDeDauw: lostinspacelogo

Seriously though, what happened with the CS rule that checked this...

src/DataAccess/DoctrineDonationRepository.php
```diff
@@ -136,4 +140,117 @@ private function getDataFieldsFromAddress( PhysicalAddress $address ) {
...
+	}
+
+	private function getPersonalInfoFromEntity( DoctrineDonation $dd ): PersonalInfo {
+		$personalInfo = new PersonalInfo();
+
+		$personalInfo->setEmailAddress( $dd->getEmail() );
+		$personalInfo->setPersonName( $this->getPersonNameFromEntity( $dd ) );
+		$personalInfo->setPhysicalAddress( $this->getPhysicalAddressFromEntity( $dd ) );
+		$personalInfo->setPhysicalAddress( $this->getPhysicalAddressFromEntity( $dd ) );
+		$personalInfo->setPhysicalAddress( $this->getPhysicalAddressFromEntity( $dd ) );
```
KaiNissen: You are really insisting on setting the physical address.

JeroenDeDauw: Yeah. Should I add a comment to clarify this is just in case the first call does not work properly?

KaiNissen: You could also extract it into a method. Something like `hitThePersonalInfoWithLeadPipeUntilItFinallySetsTheAddress()`.

PR #205, opened 2016-02-24
switch to key based message translation

KaiNissen:


tests/Integration/Presentation/Presenters/ConfirmSubscriptionHtmlPresenterTest.php
```diff
@@ -30,7 +31,7 @@ public function testGivenFailureResponse_templateIsRenderedWithoutMessages() {
 		$twig->expects( $this->once() )
 			->method( 'render' )
 			->with( [ 'error_message' => 'The confirmation code has expired.' ] );
-		$presenter = new ConfirmSubscriptionHtmlPresenter( $twig );
+		$presenter = new ConfirmSubscriptionHtmlPresenter( $twig, new Translator( 'en' ) );
```
gbirke: This case is a little bit misleading because the translator is not initialized with translations. It works none the less, because when it doesn't  find the message, it will return it unchanged. But to really have an "integration test" it could be initialized with an array like in the first code example of http://symfony.com/doc/current/components/translation/usage.html

app/routes.php
```diff
@@ -48,11 +48,11 @@ function( Request $request ) use ( $app, $ffFactory ) {
 
 		if ( $validationResult->isSuccessful() ) {
 			return $app->json( [ 'status' => 'OK' ] );
-		}
-		else {
-			$errors = array_map( function ( ConstraintViolation $constraintViolation ) {
-				return $constraintViolation->getMessage();
-			}, $validationResult->getViolations() );
+		} else {
```
JeroenDeDauw: Is the new spacing style check requiring you put it on one line?

KaiNissen: no, I just thought it was prettier

JeroenDeDauw:
Starting rebase...

JeroenDeDauw:
Since I'm not 100% I got it correct, I pushed the rebase to a new branch: https://github.com/wmde/FundraisingFrontend/pull/211

PR #212, opened 2016-02-25
Remove dead code

JeroenDeDauw:


src/DataAccess/DoctrineDonationRepository.php
```diff
@@ -60,7 +60,7 @@ private function newDonationEntity( Donation $donation ): DoctrineDonation {
 		} else {
 			$doctrineDonation->setCity( $donation->getPersonalInfo()->getPhysicalAddress()->getCity() );
 			$doctrineDonation->setEmail( $donation->getPersonalInfo()->getEmailAddress() );
-			$doctrineDonation->setName( $donation->determineFullName() );
+			$doctrineDonation->setName( $donation->getPersonalInfo()->getPersonName()->getFullName() );
```
JeroenDeDauw: Please note that line 59 is what got used before already, and that the capitlaizaion of the value there is different than the one in the removed `determineFullName` method. I do not know which capitalization is correct

KaiNissen: The capitalization is correct in "Anonym". This field is used as the author of a comment and should start with an uppercase letter.

PR #217, opened 2016-02-25
Added Donation::cancel

JeroenDeDauw:


src/Domain/Model/Donation.php
```diff
@@ -4,6 +4,8 @@
 
 namespace WMDE\Fundraising\Frontend\Domain\Model;
 
+use RuntimeException;
```
gbirke: Why are you using a generic RuntimeException?

PR #219, opened 2016-02-25
[WIP] refactor token generation

KaiNissen:
taking over token generation from old application


JeroenDeDauw:
So this is adding token generation code, which happens to be a refactored version of the old code?

Beware that @gbirke and me already discussed how to validate tokens and have some ideas here. So please double check with us before starting on that part.

KaiNissen:
After discussing with @gbirke, we came to the conclusion that we just should create a GUID and drop this approach.

PR #221, opened 2016-02-26
Work on CancelDonation UC

JeroenDeDauw:


src/UseCases/CancelDonation/CancelDonationUseCase.php
```diff
@@ -22,12 +23,30 @@ public function __construct( DonationRepository $donationRepository, TemplateBas
 	}
 
 	public function cancelDonation( CancelDonationRequest $cancellationRequest ): CancelDonationResponse {
+		try {
+			$donation = $this->donationRepository->getDonationById( $cancellationRequest->getDonationId() );
+		}
+		catch ( GetDonationException $ex ) {
+			return new CancelDonationResponse( $cancellationRequest->getDonationId(), false );
```
KaiNissen: maybe use class constants for the boolean parameter?

src/UseCases/CancelDonation/CancelDonationUseCase.php
```diff
@@ -22,12 +23,30 @@ public function __construct( DonationRepository $donationRepository, TemplateBas
...
+
+		if ( $donation === null ) {
+			return new CancelDonationResponse( $cancellationRequest->getDonationId(), false );
+		}
+
+		try {
+			$donation->cancel();
+		}
+		catch ( \RuntimeException $ex ) {
+			return new CancelDonationResponse( $cancellationRequest->getDonationId(), true );
```
KaiNissen: this should be an unsuccessful cancellation too, right?

src/UseCases/CancelDonation/CancelDonationUseCase.php
JeroenDeDauw: It's still unclear to me what the response model should be (https://phabricator.wikimedia.org/T123046), so I'm not spending time on making this nice while not knowing if it's the right stuff to begin with

src/UseCases/CancelDonation/CancelDonationUseCase.php
JeroenDeDauw: Seems like someone wrote production code that was not needed to make the test pass...

PR #222, opened 2016-02-26
Generate access and update tokens

KaiNissen:


KaiNissen:
overwrongly misbranched

PR #223, opened 2016-02-26
Generate tokens for access and update

KaiNissen:


src/Domain/TokenGenerator.php
```diff
@@ -0,0 +1,26 @@
...
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class TokenGenerator {
+
+	public function __construct( int $tokenLength, string $validityTimeSpan ) {
+		$this->tokenLength = $tokenLength;
+		$this->validityTimeSpan = $validityTimeSpan;
+	}
+
+	public function generateToken() {
```
gbirke: Type missing

src/Domain/TokenGenerator.php
```diff
@@ -0,0 +1,26 @@
...
+		return bin2hex( random_bytes( $this->tokenLength ) );
+	}
+
+	public function generateTokenExpiry() {
```
gbirke: I think this should rather generate a `DateTime` object, for easier comparison later. The serialization could then be done in the Repository.
On the other hand, the check if the token is expired has to be done with a serialized date coming the from the data, so maybe it's better for consistency to always use a serialized date.
What's your opinion, @JeroenDeDauw ?

JeroenDeDauw:
>>> naaaw

madcat

KaiNissen:
added the return types to `TokenGenerator` methods

JeroenDeDauw:
+2 from my side, though I modified it, so someone else better review

gbirke:
+2 from my side.

PR #227, opened 2016-02-29
Added DoctrineTokenAuthorizationService

JeroenDeDauw:


JeroenDeDauw:
Not entirely convinced about this Specify thing yet...

KaiNissen:
What made you introduce Specify? Can you specify?

JeroenDeDauw:
See Gabriels email on grouping tests.

My idea here is to try it out for a bit, and if we decide we don't like it, throw it out again (will not be a lot of work). So far I'm not sold on the thing, but I'd like to give it a bit more of a chance.

KaiNissen:
yes, read that in the meantime

PR #230, opened 2016-03-01
Fail at disabling SingleSpaceBeforeSingleLineComment

JeroenDeDauw:
If someone know how to do this... I tried a bunch of stuff and did not get it to work.

Ping @wmde-manicki

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -6,6 +6,10 @@
 <ruleset name="FunFunFundraisingFrontend">
     <rule ref="vendor/mediawiki/mediawiki-codesniffer/MediaWiki/Sniffs/WhiteSpace" />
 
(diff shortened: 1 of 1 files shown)


manicki:
```
<rule ref="vendor/mediawiki/mediawiki-codesniffer/MediaWiki/Sniffs/WhiteSpace/SpaceBeforeSingleLineCommentSniff.php">
    <severity>0</severity>
</rule>
```
should work but I believe PHPCS check on CI server would still fail, unless something like `--severity=0` is added to PHPCS command run on CI server. At least it seemed so when I was playing with this locally.
See also https://github.com/wmde/FundraisingFrontend/pull/231 which is saying PHPCS not to consider the sniff at all

PR #232, opened 2016-03-01
Client-side validation refactoring

gbirke:
Refactor the client-side validation so 
- validators can be functions or standalone objects
- each value can be validated with its own action
- validation results are stored in the global redux state
- validation results that are Promises (i.e. jQuery AJAX requests) are resolved to the return value of the Promise. 

The code does not display validation messages yet.

This PR contains the changes from https://github.com/wmde/FundraisingFrontend/tree/delete-all-the-js, which can be deleted when this PR is merged.


app/js/lib/actions.js
```diff
@@ -40,6 +33,13 @@ module.exports = {
 			type: 'SELECT_PAYMENT_TYPE',
 			payload: { paymentType: paymentType }
 		};
+	},
+
+	newValidateAmountAction: function ( validationResult ) {
+		return {
+			type: 'VALIDATE_AMOUNT',
+			payload: validationResult
```
KaiNissen: Is there a reason why the payload isn't an object here? Just curious.

KaiNissen:
looks good to me, but someone else should approve, too

app/js/lib/actions.js
gbirke: The payload actually _is_ an object - it's the JSON object returned by validation call ( `{ status: 'OK' }` or `{ status: 'ERR', message: 'Amount too low' }`). I'll add doc comments to all the action creators when this is merged.

In this case `validationResult` can also be the `Promise` returned by the jQuery call. The `redux-promise` middleware will handle it in then and create a new action with the validation result object returned by the promise.

KaiNissen:
this seems to break the form pagination, though

app/js/tests/test_form_validation.js
```diff
@@ -4,25 +4,44 @@ var test = require( 'tape' ),
...
+		amountValidator = validation.createAmountValidator(
+			'http://spenden.wikimedia.org/validate-amount',
+			postFunctionSpy
+		),
+		formValues = { amount: 23, paymentType: 'BEZ', otherStuff: 'foo' };
+
+	amountValidator.validate( formValues );
+	amountValidator.validate( formValues );
+
+	t.ok( postFunctionSpy.calledOnce, 'data is sent only once when input data is the same' );
```
JeroenDeDauw: That's an interesting way to get the call count from a Spy :)

app/js/lib/actions.js
```diff
@@ -40,6 +33,13 @@ module.exports = {
 			type: 'SELECT_PAYMENT_TYPE',
 			payload: { paymentType: paymentType }
 		};
+	},
+
+	newValidateAmountAction: function ( validationResult ) {
```
JeroenDeDauw: newValidate**d**AmountAction

app/templates/Form1.html.twig
```diff
@@ -509,17 +492,20 @@
...
                     $( '#donation-submit1 button' ).click( function() {
                         store.dispatch( actions.newNextPageAction() );
                     } );
                 },
-
+                connectValidatorsToStore: function( store ) {
+                    var validators = this.newValidators();
+                    WMDE.ReduxValidation.createValidationMapper( store, [
+                            validators.amountValidator
+                    ] );
```
JeroenDeDauw: If you just need this list of validators, why not just have a list to begin with, so you don't need to translate from map to list?

gbirke:
Changed the names and structure of the redux validation classes for more clarity. 

Also added some documentation.

app/js/lib/form_validation.js
```diff
@@ -4,31 +4,36 @@ var jQuery = require( 'jquery' ),
...
 		postFunction: null,
-		validate: function () {
-			var data = {
-				amount: this.amountAccessor.getValue(),
-				paymentType: this.paymentTypeAccessor.getValue()
+		validate: function ( formValues ) {
+			var postData;
+			if ( formValues.amount === this.previousAmount && formValues.paymentType === this.previousPaymentType ) {
+				return;
+			}
```
JeroenDeDauw: Did you forget to remove this?

gbirke: Yes, I did. Will make a new commit with the changes when you're done reviewing

doc/HOWTO_Create_a_form.md
```diff
@@ -0,0 +1,141 @@
...
+The `newSelectPaymentTypeAction` function is an **Action Creator** - a function that makes sure the correct action
+object with the correct payload is generated. Many actions have **payloads** - data that will be processed by the reducer functions. The payload given through the parameters of the action creator.
+
+### Setting up validation
+
+Fill the `newValidators` method body with instances of `ValidationDispatcher` classes. A `ValidationDispatcher` calls a validation function with values from the store and dispatches an action to store the validation result. It only does this when the values from the store change.
+
+The `ValidationDispatcher` constructor has three arguments:
+
+1. The **validation function** to call. The function has to have one argument - an object of the values to validate. The validation function can also be an object that has a `validate` method.
```
JeroenDeDauw: "an object of the values to validate" huh

doc/HOWTO_Create_a_form.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+The `newSelectPaymentTypeAction` function is an **Action Creator** - a function that makes sure the correct action
+object with the correct payload is generated. Many actions have **payloads** - data that will be processed by the reducer functions. The payload given through the parameters of the action creator.
+
+### Setting up validation
+
+Fill the `newValidators` method body with instances of `ValidationDispatcher` classes. A `ValidationDispatcher` calls a validation function with values from the store and dispatches an action to store the validation result. It only does this when the values from the store change.
+
+The `ValidationDispatcher` constructor has three arguments:
```
JeroenDeDauw: factory function, not constructor

gbirke: I'm not sure what you mean, please have a talk in person about that.

PR #233, opened 2016-03-02
Show validation messages on client

gbirke:
Add reducer that stores the validation messages and the source field in
the global state.

Add view handler that displays the names of the invalid fields. Messages
are not displayed at the moment, similar to the old application.

This is based on #232 , which must be merged first.


JeroenDeDauw:
Needs rebase

PR #234, opened 2016-03-02
Add authorization to CancelDonation UC

JeroenDeDauw:


JeroenDeDauw:
I'm going to make a separate PR to fix the issue in the repo

src/UseCases/CancelDonation/CancelDonationUseCase.php
```diff
@@ -49,7 +58,7 @@ public function cancelDonation( CancelDonationRequest $cancellationRequest ): Ca
 
 		$this->sendConfirmationEmail( $donation );
```
gbirke: In the old application, there was additional error checking here so when the mail was not sent, the user got the message "There was an error sending you the confirmation mail, but please beleive use that we really really canceled your donation". 
Maybe add a TODO here?

JeroenDeDauw: Does a "your donation was cancelled" message not get shown anyway? If so, I'm not sure what the added value is of mentioning the mail confirmation failed. Cause people that already read the confirmation on the page don't really need the mail to begin with no? Anyway, please do add a TODO if you think this needs further attention.

There certainly is a more general TODO with regard to transaction handling. Perhaps not in this UC, but I figure there are cases where when something fails after the DB write, such as sending a critical email, the whole UC fails and the write gets rolled back (or not committed in the first place).

tests/Integration/UseCases/CancelDonation/CancelDonationUseCaseTest.php
```diff
@@ -134,7 +144,8 @@ private function newUseCaseWithMailerMock( Donation $donation ): CancelDonationU
 
 		return new CancelDonationUseCase(
 			$this->getDonationRepositoryWithDonation( $donation ),
-			$mailer
+			$mailer,
+			new SucceedingAuthorizer()
```
gbirke: Shouldn't the authorizer failing also be a test case to cover all the code paths?

JeroenDeDauw: The new tests I wrote for the new functionality are in the route. Modifications in this file are to keep these tests working as they did before.

src/UseCases/CancelDonation/CancelDonationUseCase.php
gbirke: If something goes wrong when sending the mail, at the moment the user gets the error page instead of the cancellation result, while his donation is still canceled.

PR #235, opened 2016-03-02
display confirmation pages after successful donation

KaiNissen:


src/Presentation/Presenters/AddDonationHtmlPresenter.php
```diff
@@ -0,0 +1,26 @@
...
+class AddDonationHtmlPresenter {
+
+	private $template;
+
+	public function __construct( TwigTemplate $template ) {
+		$this->template = $template;
+	}
+
+	public function present( Donation $donation ): string {
+		return $this->template->render( [ 'donation' => $donation ] );
```
JeroenDeDauw: Like this the presenter does not add real value. Normally it should construct its own View Model. `AddDonationResponse` is already somewhat dangerous by passing the domain object to the controller code. Passing it on to the view itself is really not nice.

src/UseCases/AddDonation/AddDonationResponse.php
```diff
@@ -0,0 +1,27 @@
...
+class AddDonationResponse extends ValidationResponse {
+
+	private $donation;
+
+	public function __construct( Donation $donation ) {
+		$this->donation = $donation;
+		parent::__construct();
+	}
+
+	public function getDonation(): Donation {
```
JeroenDeDauw: I'm fine (though not very happy) with returning the domain object here, though then we should be real careful with not using domain logic in the controller or putting controller or presentation related concerns into the domain object.

src/UseCases/AddDonation/AddDonationResponse.php
```diff
@@ -0,0 +1,27 @@
...
+namespace WMDE\Fundraising\Frontend\UseCases\AddDonation;
+
+use WMDE\Fundraising\Frontend\Domain\Model\Donation;
+use WMDE\Fundraising\Frontend\ResponseModel\ValidationResponse;
+
+/**
+ * @license GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class AddDonationResponse extends ValidationResponse {
```
gbirke: I'm not sure what we gain by extending `ValidationResponse`, as the constructor is made in a way that the response is always successful. Do you really need the functionality provided by `ValidationResponse`?

JeroenDeDauw: hah, I had not even noticed this inheritance :)

src/Presentation/Presenters/AddDonationHtmlPresenter.php
KaiNissen: Right, changed it to accept an array of template arguments.

src/UseCases/AddDonation/AddDonationResponse.php
KaiNissen: Needed some of it. I removed the inheritance.

src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -177,4 +179,54 @@ private function getConfirmationMailTemplateArguments( Donation $donation, bool
...
+				'optsIntoNewsletter' => $donation->getOptsIntoNewsletter(),
+				'bankTransferCode' => $donation->getBankTransferCode(),
+				// TODO: use locale to determine the date format
+				'creationDate' => ( new \DateTime() )->format( 'd.M.Y' ),
+				// TODO: set cookie duration for "hide banner cookie"
+				'cookieDuration' => ''
+			],
+			$this->getPersonArguments( $donation ),
+			$this->getBankDataArguments( $donation )
+		] );
```
JeroenDeDauw: This code belongs in the presenter. The fact that there is a template is an implementation detail of the presenter.

KaiNissen: Right, moved it there.

JeroenDeDauw:
+2 from my side when #236 is merged, except for the template, which someone else should review

PR #240, opened 2016-03-06
redirect to paypal after successful donation

JeroenDeDauw:


src/Domain/PayPalUrlGenerator.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+class PayPalUrlGenerator {
+
+	private $config;
+
+	public function __construct( PayPalConfig $config ) {
+		$this->config = $config;
+	}
+
+	public function generateUrl( int $id, float $amount, int $interval, string $accessToken, string $updateToken ) {
```
JeroenDeDauw: I had to check the caller to verify this is the donation id, so $donationId is better

src/Domain/PayPalUrlGenerator.php
```diff
@@ -0,0 +1,65 @@
...
+	private function getSubscriptionParams( $amount, $interval ): array {
+		return [
+			'cmd' => '_xclick-subscriptions',
+			'no_shipping' => '1',
+			'src' => '1', # recurring
+			'sra' => '1', # reattempt
+			'srt' => '0', # number of payment cycles (0 for infinite)
+			'a3' => $amount,
+			'p3' => $interval,
+			't3' => 'M', # monthly payment cycle
```
JeroenDeDauw: How about using constants for these instead of adding comments?

src/Factories/FunFunFactory.php
```diff
@@ -664,6 +666,14 @@ private function newAddDonationMailer(): TemplateBasedMailer {
 		);
 	}
 
+	public function newPayPalUrlGenerator() {
+		return new PayPalUrlGenerator( $this->getPayPalConfig() );
+	}
+
+	private function getPayPalConfig() {
+		return ( PayPalConfig::newFromConfig( $this->config['paypal'] ) );
```
JeroenDeDauw: What's with the brackets?

src/Domain/PayPalUrlGenerator.php
KaiNissen: Good point, changed that.

src/Domain/PayPalUrlGenerator.php
```diff
@@ -0,0 +1,81 @@
...
+
+	private function getIntervalDependentParameters( float $amount, int $interval ): array {
+		if ( $interval > 0 ) {
+			return $this->getSubscriptionParams( $amount, $interval );
+		}
+
+		return $this->getSinglePaymentParams( $amount );
+	}
+
+	private function getSubscriptionParams( float $amount, int $interval ): array {
```
gbirke: Maybe add a function comment, and add a link to the paypal documentation  where these cryptic parameter names are explained?

gbirke:
+1

src/Domain/PayPalConfig.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->payPalAccountAddress = $payPalAccountAddress;
+		$this->payPalBaseUrl = $payPalBaseUrl;
+		$this->notifyUrl = $notifyUrl;
+		$this->returnUrl = $returnUrl;
+		$this->cancelUrl = $cancelUrl;
+		$this->itemName = $itemName;
+	}
+
+	/**
+	 * @param String[] $config
```
JeroenDeDauw: string[]

src/Domain/PayPalConfig.php
```diff
@@ -0,0 +1,84 @@
...
+	 * @return PayPalConfig
+	 */
+	public static function newFromConfig( array $config ) {
```
JeroenDeDauw: If we add constants for these keys, then we avoid a class of errors and make navigation and refactoring easier.

PR #243, opened 2016-03-08
Validate address

gbirke:
Extract validator for personal info from `DonationValidator` and use it in a new route.

This is for https://phabricator.wikimedia.org/T127001


src/Validation/PersonalInfoValidator.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Validation;
```
JeroenDeDauw: The copy-paste is weak in this one! You forgot the strict type thing :)

tests/Integration/Validation/DonationValidatorTest.php
```diff
@@ -207,4 +212,10 @@ private function newBankDataValidator() {
 		return new BankDataValidator( $ibanValidatorMock );
 	}
 
+	private function newMockPersonalInfoValidator() {
```
JeroenDeDauw: https://github.com/wmde/FundraisingFrontend/pull/245

Esp nice when using mocks, since else PhpStorm gives you a type warning

JeroenDeDauw:
Looks good, will merge when follow up is merged

PR #244, opened 2016-03-08
Split donation form into individual templates for pages 

gbirke:
Created DonationForm.html.twig that contains the form pages.

Event handling is disabled at the moment since events and view handlers
need to be added on a per-page basis to avoid infinite loops caused by
duplicate form fields.

Remove FormPageHighlight view handler since the current step display is
created through duplicated HTML code.

This is a first step for https://phabricator.wikimedia.org/T129227


JeroenDeDauw:
+1

KaiNissen:
err, btw: do we have coding conventions for HTML

JeroenDeDauw:
Guess not, at least not explicitly. Do you have anything in mind?

PR #252, opened 2016-03-09
add content for donation cancellation confirmation

KaiNissen:
This adds content to the donation cancellation confirmation page. The response model only needs to provide the donation ID.

### DIFF
--- a/app/templates/Donation_Cancellation_Confirmation.twig
+++ b/app/templates/Donation_Cancellation_Confirmation.twig
@@ -1,1 +1,18 @@
-TODO
+<div id="greenBox" class="inhalt boxShadow">
+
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
> The response model only needs to provide the donation ID.

What about the error cases? Does it not need some kinda of indication of the error type there?

KaiNissen:
Right, for failed cancellations, of course. As far as I remember, we agreed to only display a general error message in all cases, since the user wouldn't actually care about the reason. How about just passing a success flag to the view then?

JeroenDeDauw:
Sounds good. The Respone Model already has a success flag, so presumably that part is done already then

PR #254, opened 2016-03-10
Create form components

gbirke:
form components enable syncing of form values between form pages and storing form values in the global state.

Use form component for payment handling and clean up HTML and legacy JS code for payment handling

### DIFF
--- a/app/js/lib/form_components.js
+++ b/app/js/lib/form_components.js
@@ -1,0 +1,47 @@
+'use strict';
+var objectAssign = require( 'object-assign' ),
+	createDefaultChangeHandler = function ( store, contentName ) {
+		return function ( evt ) {
+			store.dispatch( { type: 'CHANGE_CONTENT', payload: {
+				value: evt.target.value,
+				contentName: contentName
+			} } );
+		};
+	},
+	RadioComponent = {
+		element: null,
+		contentName: '',
+		onChange: null,
+		render: function ( formContent ) {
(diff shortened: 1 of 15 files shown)


KaiNissen:
looks good to me. @JeroenDeDauw, mind taking a look?

app/js/lib/form_components.js
```diff
@@ -0,0 +1,45 @@
...
+	createDefaultChangeHandler = function ( store, contentName ) {
+		return function ( evt ) {
+			store.dispatch( actions.newChangeContentAction( contentName, evt.target.value ) );
+		};
+	},
+	RadioComponent = {
+		element: null,
+		contentName: '',
+		onChange: null,
+		render: function ( formContent ) {
```
JeroenDeDauw: setValue?

JeroenDeDauw: After reading the code calling this I realized this gets the whole form content... which the var name kinda suggests :) That still makes the method name odd though, as this component does not render the whole form content.

gbirke: Would `formData` and `fieldName` be better names?

JeroenDeDauw: Do you mean `formData` instead of `formContent`?

I was talking about the method name. I think `formContent` is clear enough.

gbirke: Many JavaScript view libraries have a `render` method that signifies to the view/component that it should render itself. So the mismatch IMHO comes from the parameter name, not from the method name.

PR #255, opened 2016-03-10
presenter for the cancel donation use case

KaiNissen:
This PR adds the presenter class for the cancel donation use case as well as corresponding route tests. It is part of [phab:123046](https://phabricator.wikimedia.org/T123046).

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -316,21 +316,12 @@ $app->post(
 
 		$response = $ffFactory->newCancelDonationUseCase( $request->get( 'utoken', '' ) )->cancelDonation( $cancellationRequest );
 
-		// TODO: response should be HTML/redirect?
+		// TODO: reset spenden_stamp in cookie
 
-		if ( $response->cancellationWasSuccessful() ) {
-			// TODO: reset spenden_stamp in cookie
-
-			return $app->json( [
-				'status' => 'OK',
-				'message' => '',
-			] );
-		}
-
-		return $app->json( [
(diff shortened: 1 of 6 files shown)


app/routes.php
```diff
@@ -316,21 +316,12 @@ function( Request $request ) use ( $app, $ffFactory ) {
...
-				'status' => 'OK',
-				'message' => '',
-			] );
-		}
-
-		return $app->json( [
-			'status' => 'ERR',
-			'message' => 'TODO',
-		] );
+		return $ffFactory->newCancelDonationHtmlPresenter()->present(
```
JeroenDeDauw: You can just pass the response to presenter.

JeroenDeDauw:
Looks good +1. Probably good if someone also double checks the german stuffs

KaiNissen:
rebased

app/templates/Donation_Cancellation_Confirmation.twig
```diff
@@ -1,9 +1,19 @@
 <div id="greenBox" class="inhalt boxShadow">
 
 	<div class="text">
-		<h2>{$ 'donation-cancellation-header'|trans({}) $}</h2>
+		<h2>
+			{% if cancellationSuccessful %}
+			{$ 'donation-cancellation-header'|trans({}) $}</h2>
```
gbirke: You can do just `{$ 'donation-cancellation-header'|trans $}` when you have no variables.

PR #259, opened 2016-03-11
Store recurring payment data in global state

gbirke:
Use form components to store and synchronize the data on recurring
payments in the global state.

Adjust form HTML to have unique IDs, correct legacy JavaScript for the
new IDs.

This is based on #254, please merge that first.

This is work for https://phabricator.wikimedia.org/T129227

### DIFF
--- a/app/js/lib/reducers/form_content.js
+++ b/app/js/lib/reducers/form_content.js
@@ -5,7 +5,8 @@ var objectAssign = require( 'object-assign' ),
 	initialState = {
 		amount: 0,
 		isCustomAmount: false,
-		paymentType: 'BEZ'
+		paymentType: 'BEZ',
+		paymentPeriodInMonths: 0 // 0, 1, 3, 6 or 12, 0 = non-recurring payment
 	};
 
 module.exports = function formContent( state, action ) {
--- a/app/js/lib/form_components.js
+++ b/app/js/lib/form_components.js
@@ -1,11 +1,13 @@
 'use strict';
 var objectAssign = require( 'object-assign' ),
 	actions = require( './actions' ),
+
 	createDefaultChangeHandler = function ( store, contentName ) {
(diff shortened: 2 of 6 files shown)


app/js/lib/reducers/form_content.js
```diff
@@ -5,9 +5,31 @@ var objectAssign = require( 'object-assign' ),
 	initialState = {
 		amount: 0,
 		isCustomAmount: false,
-		paymentType: 'BEZ'
+		paymentType: 'BEZ',
+		recurringPayment: 0, // 0 = one-off, 1 = recurring Payment
```
JeroenDeDauw: These are fields right? Then can't you have an isRecurring method? I don't like the `isFalsy` stuff below. Defensive coding like this obfuscates intend to some degree.

app/js/lib/reducers/form_content.js
```diff
@@ -5,9 +5,31 @@ var objectAssign = require( 'object-assign' ),
...
+		paymentPeriod: 0 // 0, 1, 3, 6 or 12
```
JeroenDeDauw: paymentPeriodInMethods? (Assuming it's months based on the values in the comment)

JeroenDeDauw: Is it really easier to have two fields here? Looks like `recurringPayment` is made redundant by `paymentPeriod` (assuming 0 for the former implies 0 for the later).  If so, then an `isRecurring` method removes the need for the first field no?

gbirke: You're right, using one field instead of two will get rid of that rule stuff and its tests and still achieve the same effect. I'll change this PR.

JeroenDeDauw:
+1 from my side so far

gbirke:
Changed to `paymentPeriodInMonths`.

PR #262, opened 2016-03-11
The AddDonation UC now writes its token to a cookie

JeroenDeDauw:
Added a test verifying this from an as high level as possible,
to not bind to implementation details and make sure the combination
of the routes works correctly. I'm not sure we're already
providing the interface we want to the clients with regards to
the update token, so that might still need further work.

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -5,13 +5,16 @@ declare( strict_types = 1 );
 namespace WMDE\Fundraising\Frontend\App\RouteHandlers;
 
 use Silex\Application;
+use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use WMDE\Fundraising\Frontend\Domain\Model\PaymentType;
 use WMDE\Fundraising\Frontend\Domain\Model\PersonalInfo;
 use WMDE\Fundraising\Frontend\Domain\Model\PersonName;
 use WMDE\Fundraising\Frontend\Domain\Model\PhysicalAddress;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\UseCases\AddDonation\AddDonationRequest;
+use WMDE\Fundraising\Frontend\UseCases\AddDonation\AddDonationResponse;
 
 /**
  * @license GNU GPL v2+
@@ -19,6 +22,9 @@ use WMDE\Fundraising\Frontend\UseCases\AddDonation\AddDonationRequest;
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
 class AddDonationHandler {
+
+	const UPDATE_TOKEN_COOKIE_NAME = 'wmde-fundraising-utoken';
+	const DONATION_ID_COOKIE_NAME = 'wmde-fundraising-donation-id';
 
 	private $ffFactory;
 	private $app;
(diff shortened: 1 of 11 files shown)


app/RouteHandlers/AddDonationHandler.php
```diff
@@ -28,36 +34,62 @@ public function __construct( FunFunFactory $ffFactory, Application $app ) {
...
-				case PaymentType::CREDIT_CARD:
-					return $this->ffFactory->newCreditCardPaymentHtmlPresenter()->present( $responseModel );
-			}
-			// TODO: take over confirmation page selection functionality from old application
-			// TODO: return update token
+		if ( !$responseModel->isSuccessful() ) {
+			return new Response( 'TODO: error occurred' ); // TODO
+		}
+
+		// TODO: take over confirmation page selection functionality from old application
```
gbirke: Isn't that what `newHttpResponseFromResponseModel ` does? I'm confused.

tests/System/MultiRoute/DonationAndCancellationTest.php
```diff
@@ -0,0 +1,46 @@
...
+				AddDonationRouteTest::newValidFormInput()
+			);
+
+			$this->assertSame( 200, $client->getResponse()->getStatusCode() );
+
+			$client->request(
+				'POST',
+				'/donation/cancel',
+				[
+					'sid' => $client->getCookieJar()->get( self::DONATION_ID_COOKIE_NAME )->getValue(),
```
gbirke: I had to look up the method [signature of the `request` method]( http://api.symfony.com/2.7/Symfony/Component/HttpKernel/Client.html#method_request) to see that the cookie values you are passing in are not  GET parameters but general "request" parameters. Otherwise I would have wondered how the cookie values will be converted to URLs.

gbirke:
+2 from me, but @KaiNissen should also look at it.

app/RouteHandlers/AddDonationHandler.php
JeroenDeDauw: That might be, I'm not sure. I did not change the functionality of that code (just refactored it a bit). Same with the TODO, I just moved it. Perhaps Kai finished the TODO and did not remove it yet?

tests/System/MultiRoute/DonationAndCancellationTest.php
JeroenDeDauw: I actually was thinking that the cancel route as we implemented it here so far as using URL parameters, and was planning to ask you or Kai about that, as it seemed rather odd when I was working on this cookie stuff. So I'm winning the "most confoosed" badge :)

Looking at what is going on in `Request`, the `get` method which we're using in the routes will use GET and POST params, with the former getting precedence. No cookies though. (Not 100% sure I understood the code correctly.)

So we probably want to change that, and also stop using `get`, in favour of more explicit accessors. (Which is actually recommended by the `get` docs). Let's do that in a different PR tho.

app/RouteHandlers/AddDonationHandler.php
KaiNissen: By "confirmation page selection functionality" I meant different confirmation pages to a/b-test the membership form. This is not done yet.

tests/System/MultiRoute/DonationAndCancellationTest.php
KaiNissen: This way the information is sent twice (as a cookie and as a POST parameter). Shouldn't the route try to get the needed values from either cookie, POST, or GET?

JeroenDeDauw:
I'll submit some other stuff to fix the issues in the routes, as this goes beyond the scope of this PR

PR #267, opened 2016-03-13
Mark unfinished auth update tests as skipped

JeroenDeDauw:


gbirke:
+2 if you use `markTestIncomplete` instead. `markTestSkipped` should be reserved for environment-related reasons IMHO.

JeroenDeDauw:
Nice catch

PR #269, opened 2016-03-14
Added stubs for membership UCs

JeroenDeDauw:


gbirke:
+1 from my side, although `CancelMembershipRequestRequest` is a nasty class name.

I'd like @KaiNissen to chime in what we are modeling here, memberships (which may also be processed and displayed by the backend as such) or membership requests? Related to the discussion at https://github.com/wmde/FundraisingStore/pull/61

JeroenDeDauw:
> CancelMembershipRequestRequest is a nasty class name.

I know right. CancelMembershipRequestResponse as well :) Still better than dropping critical parts of the name for the sake of brevity.

JeroenDeDauw:
The modeling question is a lot easier here than in the store, as we have a domain with well defined boundary here. It does not matter for the UCs and request/response models of this application what the persistence layout looks like, or which other applications might be talking to it, and what their view on data that happens to be in the same table is.

KaiNissen:
As mentioned in [wmde/FundraisingStore/pull/61](https://github.com/wmde/FundraisingStore/pull/61#issuecomment-196812594) all those are actually only requests. Didn't we start to call them "inquiries" at some point to reduce the confusement factor?

gbirke:
It was called "Inquiry" while memberships were lumped together with subscriptions. 

@KaiNissen said in a comment on the fundraising store: 
> So this table is more like a stack of membership application forms. 

The suffix "Application" would also be an option (although "application" can also be confusing). [Looking at the thesaurus](http://www.thesaurus.com/browse/application?s=t), we could also call the application "submission". 

Personally, I'd like to keep the "Membership" in the name to make it more clear what kind of "inquiry",  "request" or "application" we have here. 

Here is a list of possible new names for the Add/Cancel Membership use cases

- ApplyForMembership/CancelMembershipApplication
- AddMembershipSubmission/CancelMembershipSubmission
- AddMembershipInquiry/CancelMembershipInquiry

My personal favourite is the "Inquiry" one.

JeroenDeDauw:
> ApplyForMembership/CancelMembershipApplication

I like this one, and don't think "Application" will be confused here.

> My personal favourite is the "Inquiry" one.

If I'm not mistaken, the meaning of this one is off. One can send a membership inquiry without actually applying for membership, or inquire about ones existing membership, or even inquire about the membership application.

manicki:
+1 for the "Application". I like @KaiNissen's analogy to membership application form.
And I have also had similar concerns about the "Inquiry", although couldn't verbalize them.
Commenting just to add some more confoosion here:)

PR #271, opened 2016-03-15
Update token in AddDonation use case

KaiNissen:
- removed access token from paypal url
- actually use update token

This PR is part of [phab:T123040](https://phabricator.wikimedia.org/T123040).


JeroenDeDauw:
Naaaw, now #262 will has a conflicts!

PR #282, opened 2016-03-18
Optimized test imports

JeroenDeDauw:


gbirke:
Rather "organize alphabetically" than "optimize", right? I always think of performance improvements when hearing "optimize".

JeroenDeDauw:
Name comes from IDE. Action includes removal of unused imports

oi

PR #285, opened 2016-03-18
Show and hide bank data

gbirke:
Show and hide the input fields of the bank data, depending on the selected payment method.

Remove the old "slide" code in other places (personal info, payment period) and replace it with the new view handler class.


JeroenDeDauw:
Needs rebase

gbirke:
Rebased.

app/js/lib/store_update_handling.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
 			viewHandlers.forEach( function ( viewHandlerConfig ) {
 				viewHandlerConfig.viewHandler.update.call(
 					viewHandlerConfig.viewHandler,
-					state[ viewHandlerConfig.stateKey ]
+					_.get( state, viewHandlerConfig.stateKey )
```
JeroenDeDauw: why?

app/js/lib/view_handler/element_slide.js
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * View Handler for showing and hiding elements if the update value matches a regular expression
+	 * @class
+	 */
+	ElementSlider = {
+		el: null,
+		showOnValue: null,
+
+		// internal fields
```
JeroenDeDauw: This makes me think of space ships :)

app/js/lib/view_handler/element_slide.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+
+	/**
+	 * View Handler for showing and hiding elements if the update value matches a regular expression
+	 * @class
+	 */
+	ElementSlider = {
```
JeroenDeDauw: file name is missing the "r"

JeroenDeDauw: At first glance I was expecting this to be some kind of control element, while it actually is an encapsulation of animation stuff. Perhaps something like ElementSlideAnimator?

app/js/lib/view_handler/element_slide.js
```diff
@@ -0,0 +1,60 @@
...
+		el: null,
+		showOnValue: null,
```
JeroenDeDauw: name does not indicate this is a regex

app/js/lib/view_handler/element_slide.js
```diff
@@ -0,0 +1,60 @@
...
+
+		// internal fields
+		slideSpeed: 600,
+		isHidden: null,
```
JeroenDeDauw: should this not be initialized as false?

app/js/lib/view_handler/element_slide.js
```diff
@@ -0,0 +1,60 @@
...
+					return;
+				}
+				this.el
+					.slideUp( this.slideSpeed )
+					.animate(
+						{ opacity: 0 },
+						{ queue: false, duration: this.slideSpeed }
+					);
+				this.isHidden = true;
+			}
```
JeroenDeDauw: How about extracting `showValue` and `hideValue`?

JeroenDeDauw:
Generally looks OK to me

app/js/lib/store_update_handling.js
gbirke: To allow for accessing values that are nested somewhere in the global object. See DonationForm initialization for examples, where the element slider view handlers needs a simple value instead of the whole `formValues` object.

app/js/lib/view_handler/element_slide.js
gbirke: No, it should be null to always trigger the show/hide event when the store is initialized. This makes the view Handler more independent form the initial HTML/CSS state, which may also be different for different instances of the view handler. Take for example period and bank data fields. In the default form, period is hidden and bank data fields are shown. If we initialize with `false` instead of `null`, the creator of the form must remember to remove the "display:none" code from the bank data fields, otherwise the visual state and the intended state differ.

Any recommendations on how to make this intention more clear in the code?

app/js/lib/store_update_handling.js
JeroenDeDauw: I'm confoosed. And you ran away from your spot <o>

app/js/lib/view_handler/element_slide.js
JeroenDeDauw: > Any recommendations on how to make this intention more clear in the code?

Not really. Perhaps a brief to the point comment?

    // Null to always trigger show/hide on first call to update

PR #288, opened 2016-03-18
Remove currently not used and broken DonationAndCancellationTest

JeroenDeDauw:


gbirke:
Ok, I get why it's "broken" (=incomplete) at the moment, but why do you want to delete it?

JeroenDeDauw:
I do not see how we will be able to re-enable this if we don't go with the cookie approach (this is also why the code was added originally)

PR #291, opened 2016-03-21
Added initial version of MembershipApplication domain entity

JeroenDeDauw:
https://phabricator.wikimedia.org/T130199

Includes commits from the euro branch: https://github.com/wmde/FundraisingFrontend/pull/301

Ready to go from my side. I think I got all the stuff needed in the model. Not sure about a bunch of stuff, such as what the MembershipPayment type actually means. We can improve that in follow ups though. This code already allows most of creation of the repos, UCs, etc

### DIFF
--- a/src/Domain/Model/MembershipApplication.php
+++ b/src/Domain/Model/MembershipApplication.php
@@ -11,27 +11,23 @@ namespace WMDE\Fundraising\Frontend\Domain\Model;
 class MembershipApplication {
 
 	/**
-	 * @var int
+	 * @var int|null
 	 */
 	private $id;
 
-	/**
-	 * @var MembershipApplicant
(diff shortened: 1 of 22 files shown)


src/Domain/Model/MembershipApplication.php
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class MembershipApplication {
+
+	/**
+	 * @var int
+	 */
+	private $id;
```
JeroenDeDauw: Needs to be nullable :( Unless we do something different than having the repository assign the ID on write

PR #298, opened 2016-03-24
confirmation page selector

KaiNissen:
solves [phab:T131239](https://phabricator.wikimedia.org/T131239)

### DIFF
--- a/src/Domain/SelectedConfirmationPage.php
+++ b/src/Domain/SelectedConfirmationPage.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Domain;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class SelectedConfirmationPage {
+
+	private $campaignCode;
+	private $pageTitle;
+
+	public function __construct( string $campaignCode, string $pageTitle ) {
+		$this->campaignCode = $campaignCode;
+		$this->pageTitle = $pageTitle;
+	}
+
+	public function getCampaignCode(): string {
+		return $this->campaignCode;
+	}
+
+	public function getPageTitle(): string {
+		return $this->pageTitle;
+	}
+
+}
--- a/src/Domain/TemplateTestCampaign.php
+++ b/src/Domain/TemplateTestCampaign.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace WMDE\Fundraising\Frontend\Infrastructure;
+namespace WMDE\Fundraising\Frontend\Domain;
 
 /**
  * @licence GNU GPL v2+
--- a/src/Presentation/DonationConfirmationPageSelector.php
+++ b/src/Presentation/DonationConfirmationPageSelector.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Presentation;
+
+/**
+ * Selects a random confirmation page from given options
+ *
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DonationConfirmationPageSelector {
+
+	private $config;
+
+	public function __construct( array $config ) {
+		$this->config = $config;
+	}
+
+	public function selectPage(): string {
+		return empty( $this->config ) ? '' : $this->config[$this->getRandomIndex()];
+	}
+
+	private function getRandomIndex() {
+		return random_int( 0, count( $this->config ) - 1 );
+	}
+
+}
--- a/src/Presentation/DonationConfirmationPageSelector.php
+++ b/src/Presentation/DonationConfirmationPageSelector.php
@@ -2,7 +2,8 @@
 
 namespace WMDE\Fundraising\Frontend\Presentation;
 
-use WMDE\Fundraising\Frontend\Infrastructure\TemplateTestCampaign;
+use WMDE\Fundraising\Frontend\Domain\SelectedConfirmationPage;
+use WMDE\Fundraising\Frontend\Domain\TemplateTestCampaign;
 
 /**
  * Selects a random confirmation page from given options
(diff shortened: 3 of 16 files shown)


src/Presentation/DonationConfirmationPageSelector.php
```diff
@@ -0,0 +1,27 @@
...
+class DonationConfirmationPageSelector {
+
+	private $config;
+
+	public function __construct( array $config ) {
+		$this->config = $config;
+	}
+
+	public function selectPage(): string {
+		return empty( $this->config ) ? '' : $this->config[$this->getRandomIndex()];
```
JeroenDeDauw: Returning an empty string here seems odd. How about throwing an exception in the constructor if the $config thing is empty?

src/Presentation/DonationConfirmationPageSelector.php
```diff
@@ -0,0 +1,27 @@
...
+ * Selects a random confirmation page from given options
+ *
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DonationConfirmationPageSelector {
+
+	private $config;
+
+	public function __construct( array $config ) {
```
JeroenDeDauw: $confirmationPages

You can do `string ...$confirmationPages`

src/Presentation/DonationConfirmationPageSelector.php
```diff
@@ -0,0 +1,27 @@
...
+		$this->config = $config;
+	}
+
+	public function selectPage(): string {
+		return empty( $this->config ) ? '' : $this->config[$this->getRandomIndex()];
+	}
+
+	private function getRandomIndex() {
```
JeroenDeDauw: Can't you just use `array_rand`?

src/Presentation/Presenters/DonationConfirmationHtmlPresenter.php
```diff
@@ -19,13 +19,14 @@ public function __construct( TwigTemplate $template ) {
 		$this->template = $template;
 	}
 
-	public function present( Donation $donation, string $updateToken ): string {
-		return $this->template->render( $this->getConfirmationPageArguments( $donation, $updateToken ) );
+	public function present( Donation $donation, string $updateToken, string $templateName = '' ): string {
```
JeroenDeDauw: What's the reasoning behind switching this template included in the main one, rather than just switching the main one? The later case seems more straightforewards to me, as it avoids dealing with two templates, and having this $templateName thing here that can be empty, and then default to something, but only if the provided main template takes care of that.

KaiNissen: MediaWiki and Twig don't work together that well in this particular case:

The content in MediaWiki needs to be enclosed within `<html>` to prevent `{% extends 'BaseLayout.twig' %}` from being url encoded as `{% extends 'BaseLayout.twig'&#160;%}`, which makes Twig complain about an unexpected `&`.

The MW parser at least encloses the content within a `<p />` tag which makes Twig complain about content being outside of blocks.

src/Presentation/DonationConfirmationPageSelector.php
KaiNissen: `array_rand()` is stated as being "not so random". But right, actually in this case "not so random" should still be random enough.

src/Presentation/DonationConfirmationPageSelector.php
KaiNissen: The configuration will be extended by several values anyway. How about defining the default confirmation page in the config and returning that? We would also get rid of the hard coded one in DonationConfirmation.twig:4

src/Presentation/Presenters/DonationConfirmationHtmlPresenter.php
JeroenDeDauw: Ok. Then how about just making the $templateName here required? And perhaps rename it somehow to indicate this is not the name of the main template.

src/Presentation/DonationConfirmationPageSelector.php
JeroenDeDauw: Is there a reason to have a possibly empty list, plus one fallback value, rather than a list that is never empty?

KaiNissen: the selector now knows the default value and returns it in case there is no running and/or configured campaign

src/Domain/SelectedConfirmationPage.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Domain;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class SelectedConfirmationPage {
```
JeroenDeDauw: I think this makes most sense when regarded as part of the presentation layer, not the

src/Domain/TemplateTestCampaign.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Domain;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Leszek Manicki <leszek.manicki@wikimedia.de>
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class TemplateTestCampaign {
```
JeroenDeDauw: Things in the domain cannot know about templates, which are a presentation concern

src/Presentation/DonationConfirmationPageSelector.php
```diff
@@ -0,0 +1,53 @@
...
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DonationConfirmationPageSelector {
+
+	private $defaultPageTitle;
+	/** @var TemplateTestCampaign[] */
+	private $campaigns;
+
+	public function __construct( array $config ) {
```
JeroenDeDauw: I think its better to have `string $defaultPageTitle, array $campaigns` here

JeroenDeDauw:
I'll submit follow up(s) addressing my last comments

src/Presentation/DonationConfirmationPageSelector.php
JeroenDeDauw: Gave this a try, not convinced this is actually better here given $campaigns still has the config format

PR #299, opened 2016-04-01
reload form and pass request vars on validation errors

KaiNissen:
[phab:T130360](https://phabricator.wikimedia.org/T130360)

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -39,6 +39,7 @@ use WMDE\Fundraising\Frontend\Domain\Model\MailAddress;
 use WMDE\Fundraising\Frontend\Domain\Model\PaymentType;
 use WMDE\Fundraising\Frontend\Presentation\PayPalConfig;
 use WMDE\Fundraising\Frontend\Presentation\PayPalUrlGenerator;
+use WMDE\Fundraising\Frontend\Presentation\Presenters\DonationFormViolationPresenter;
 use WMDE\Fundraising\Frontend\Domain\ReferrerGeneralizer;
 use WMDE\Fundraising\Frontend\Domain\Repositories\CommentFinder;
 use WMDE\Fundraising\Frontend\DataAccess\DbalSubscriptionRepository;
@@ -802,6 +803,10 @@ class FunFunFactory {
 
 	public function getDonationConfirmationPageSelector() {
 		return $this->pimple['confirmation-page-selector'];
+	}
+
+	public function newDonationFormViolationPresenter() {
+		return new DonationFormViolationPresenter( $this->getLayoutTemplate( 'DonationForm.html.twig' ) );
 	}
 
 }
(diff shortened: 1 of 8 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -791,4 +792,8 @@ public function newShowDonationConfirmationUseCase( string $accessToken ): ShowD
 		);
 	}
 
+	public function newViolationPresenter() {
```
JeroenDeDauw: Think DonationForm should be in the name here as well

src/UseCases/AddDonation/AddDonationResponse.php
```diff
@@ -50,6 +50,13 @@ public static function newFailureResponse( array $errors ): self {
 	private function __construct() {
 	}
 
+	public static function newViolationResponse( array $errors, AddDonationRequest $request ) {
+		$response = new self();
+		$response->validationErrors = $errors;
+		$response->precedingRequest = $request;
```
JeroenDeDauw: field does not exist

src/UseCases/AddDonation/AddDonationResponse.php
```diff
@@ -50,6 +50,13 @@ public static function newFailureResponse( array $errors ): self {
...
+		return $response;
+	}
```
JeroenDeDauw: I don't understand why you added this

JeroenDeDauw:
Needs rebase

src/UseCases/AddDonation/AddDonationResponse.php
KaiNissen: Meh, those are leftovers from an approach I dropped. Removed, rebased, repushed.

JeroenDeDauw:
Follow mup: https://github.com/wmde/FundraisingFrontend/pull/303

tests/templates/DonationForm.html.twig
```diff
@@ -0,0 +1,18 @@
...
+Salutation: {$ anrede $}
+Title: {$ titel $}
+Company: {$ firma $}
+First name: {$ vorname $}
+Last name: {$ nachname $}
+Street: {$ strasse $}
+Postal code: {$ plz $}
+City: {$ ort $}
+Country code: {$ country $}
+Email address: {$ email $}
```
JeroenDeDauw: So this is a placeholder form, that should still be replaced by some real frontend stuffs? How would that work?

src/Presentation/Presenters/DonationFormViolationPresenter.php
```diff
@@ -0,0 +1,93 @@
...
+ */
+class DonationFormViolationPresenter {
+
+	private $template;
+
+	public function __construct( TwigTemplate $template ) {
+		$this->template = $template;
+	}
+
+	public function present( AddDonationRequest $request ): string {
```
JeroenDeDauw: So the errors themselves are not shown?

tests/templates/DonationForm.html.twig
KaiNissen: This is for the system test to make sure all required data is passed to the presenter. The actual form will use the parameters to populate/restore the form fields with the given values so users don't have to fill the form again.

PR #301, opened 2016-04-05
Correctly handle money amounts

JeroenDeDauw:
https://phabricator.wikimedia.org/T130525

### DIFF
--- a/src/Domain/Model/Euro.php
+++ b/src/Domain/Model/Euro.php
@@ -1,0 +1,83 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Domain\Model;
+
+use InvalidArgumentException;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+final class Euro {
+
+	private static $DECIMAL_COUNT = 2;
+	private static $CENTS_PER_EURO = 100;
+
+	private $cents;
+
+	/**
+	 * @param int $cents
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( int $cents ) {
+		if ( $cents < 0 ) {
+			throw new InvalidArgumentException( 'Amount needs to be positive' );
+		}
+
+		$this->cents = $cents;
+	}
+
+	/**
+	 * Constructs a Euro object from a string representation such as "13.37".
+	 *
+	 * This method takes into account the errors that can arise from floating
+	 * point number usage. Amounts with too many decimals are rounded to the
+	 * nearest whole euro cent amount.
+	 *
+	 * @param string $euroAmount
+	 * @return self
+	 * @throws InvalidArgumentException
+	 */
+	public static function newFromString( string $euroAmount ): self {
+		if ( !is_numeric( $euroAmount ) ) {
+			throw new InvalidArgumentException( 'Not a number' );
+		}
+
+		return new self( intval(
+			round( $euroAmount, self::$DECIMAL_COUNT ) * self::$CENTS_PER_EURO
+		) );
+	}
+
+	/**
+	 * This method takes into account the errors that can arise from floating
+	 * point number usage. Amounts with too many decimals are rounded to the
+	 * nearest whole euro cent amount.
+	 *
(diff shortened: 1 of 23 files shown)


JeroenDeDauw:
@WMDE-Fisch @manicki Please beware of this class, it might be of use in the backend app as well

JeroenDeDauw:
If anyone can think of a case that is not tested yet, please point it out. Kinda important this code does not subtly alter the values.

src/Domain/Model/Euro.php
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	public static function newFromString( string $euroAmount ): self {
+		if ( !is_numeric( $euroAmount ) ) {
+			throw new InvalidArgumentException( 'Not a number' );
+		}
+
+		return new self( intval(
+			round( $euroAmount, self::$DECIMAL_COUNT ) * self::$CENTS_PER_EURO
+		) );
+	}
```
JeroenDeDauw: Method loosely based on https://github.com/sebastianbergmann/money/blob/master/src/Money.php#L80

JeroenDeDauw:
I don't like how this reads: `new Euro( 1000 )`. Looks like that is 1000 Euro, while it's just 10. Easy for someone to get confused. Perhaps we should make the constructor private and have `newFromCents( int $i )` and `newWholeAmount( int $i )` methods or something like that?

src/Domain/Model/Euro.php
```diff
@@ -0,0 +1,83 @@
...
+use InvalidArgumentException;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+final class Euro {
+
+	private static $DECIMAL_COUNT = 2;
+	private static $CENTS_PER_EURO = 100;
```
KaiNissen: Is there a reason to use static members rather than constants here?

src/Domain/Model/Euro.php
```diff
@@ -0,0 +1,83 @@
...
+
+	private $cents;
+
+	/**
+	 * @param int $cents
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( int $cents ) {
+		if ( $cents < 0 ) {
+			throw new InvalidArgumentException( 'Amount needs to be positive' );
```
KaiNissen: Not sure about this. Our use cases require amounts to be positive, in general a currency value might as well be negative. So isn't this something that should be handled by the validators?

src/Infrastructure/AmountParser.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class AmountParser {
+
+	private $locale;
+
+	public function __construct( string $locale = 'de_DE' ) {
```
KaiNissen: Make `$locale` required maybe? There are no cases in which `AmountParser` is created without passing this parameter, anyway.

KaiNissen:
I totally agree to your comment above regarding the constructor taking cents.

Nice production code/test code ratio, by the way! :)

src/Domain/Model/Euro.php
JeroenDeDauw: Yeah - they can be private. Once we switch to PHP 7.1, we can have private consts. Not that it matters much anyway

src/Domain/Model/Euro.php
JeroenDeDauw: This class is geared towards our usecases. It is not meant to be a general thing, which would be more complex and less safe for us.

JeroenDeDauw:
> Nice production code/test code ratio, by the way! :)

I actually ran phploc on the files to get the LLOC. That is a lot more skewed than the LOC.

src/Infrastructure/AmountParser.php
JeroenDeDauw: Hmm, sure. Beware that it is you that added the optional thing here, I just moved it around

```php
public function setAmountFromString( string $value, string $locale = 'de_DE' ) {
```

PR #306, opened 2016-04-08
Added initial version of DoctrineMembershipApplicationRepository

JeroenDeDauw:
The data field still needs to be handled in both write and read methods, the phone number is still missing, and the timestamp thing needs work.

Though by all means merge away. The code here is complete in the sense of that it does not fail, is tested, and is clean.

The tests here are an iteration on the ones for the Donation repository, and I'm liking these a lot more now. They are nicer and more complete.

Part of https://phabricator.wikimedia.org/T123117 and https://phabricator.wikimedia.org/T130199

### DIFF
--- a/src/DataAccess/DoctrineMembershipApplicationRepository.php
+++ b/src/DataAccess/DoctrineMembershipApplicationRepository.php
@@ -1,0 +1,96 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\DataAccess;
+
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\ORMException;
+use WMDE\Fundraising\Entities\MembershipApplication as DoctrineApplication;
+use WMDE\Fundraising\Frontend\Domain\Model\MembershipApplicant;
+use WMDE\Fundraising\Frontend\Domain\Model\MembershipApplication;
+use WMDE\Fundraising\Frontend\Domain\Model\MembershipPayment;
+use WMDE\Fundraising\Frontend\Domain\Repositories\GetMembershipApplicationException;
+use WMDE\Fundraising\Frontend\Domain\Repositories\MembershipApplicationRepository;
+use WMDE\Fundraising\Frontend\Domain\Repositories\StoreMembershipApplicationException;
+
+/**
+ * @license GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class DoctrineMembershipApplicationRepository implements MembershipApplicationRepository {
+
+	private $entityManager;
+
+	public function __construct( EntityManager $entityManager ) {
+		$this->entityManager = $entityManager;
+	}
+
+	public function storeApplication( MembershipApplication $application ) {
+		$doctrineApplication = $this->newDoctrineApplication( $application );
+
+		try {
+			$this->entityManager->persist( $doctrineApplication );
+			$this->entityManager->flush();
+		}
+		catch ( ORMException $ex ) {
+			throw new StoreMembershipApplicationException( $ex );
+		}
+
+		$application->setId( $doctrineApplication->getId() );
+	}
+
+	private function newDoctrineApplication( MembershipApplication $application ): DoctrineApplication {
+		$doctrineApplication = new DoctrineApplication();
+
+		$this->setApplicantFields( $doctrineApplication, $application->getApplicant() );
+		$this->setPaymentFields( $doctrineApplication, $application->getPayment() );
+
+		$doctrineApplication->encodeAndSetData( [] ); // TODO
+
+		return $doctrineApplication;
+	}
+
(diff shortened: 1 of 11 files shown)


src/DataAccess/DoctrineMembershipApplicationRepository.php
```diff
@@ -0,0 +1,96 @@
...
+		$application->setCity( $address->getCity() );
+		$application->setCountry( $address->getCountryCode() );
+		$application->setPostcode( $address->getPostalCode() );
+		$application->setAddress( $address->getStreetAddress() );
+	}
+
+	private function setPaymentFields( DoctrineApplication $application, MembershipPayment $payment ) {
+		$application->setMembershipType( $payment->getType() );
+		$application->setMembershipFeeInterval( $payment->getIntervalInMonths() );
+		$application->setMembershipFee( (int)$payment->getAmount()->getEuroFloat() );
```
JeroenDeDauw: Am I correct in understanding that the db has an integer field that represents the amount of Euro (as opposed to Euro cents)?

src/DataAccess/DoctrineMembershipApplicationRepository.php
```diff
@@ -0,0 +1,96 @@
...
+		$address = $applicant->getPhysicalAddress();
+
+		$application->setCity( $address->getCity() );
+		$application->setCountry( $address->getCountryCode() );
+		$application->setPostcode( $address->getPostalCode() );
+		$application->setAddress( $address->getStreetAddress() );
+	}
+
+	private function setPaymentFields( DoctrineApplication $application, MembershipPayment $payment ) {
+		$application->setMembershipType( $payment->getType() );
```
JeroenDeDauw: See mailing list question about this type stuff

tests/Data/ValidMembershipApplication.php
```diff
@@ -0,0 +1,142 @@
...
+
+		$application->setMembershipType( self::PAYMENT_TYPE );
+		$application->setMembershipFeeInterval( self::PAYMENT_PERIOD_IN_MONTHS );
+		$application->setMembershipFee( self::PAYMENT_AMOUNT_IN_EURO );
+
+		$application->setAccountNumber( self::PAYMENT_BANK_ACCOUNT );
+		$application->setBankCode( self::PAYMENT_BANK_CODE );
+		$application->setBankName( self::PAYMENT_BANK_NAME );
+		$application->setBic( self::PAYMENT_BIC );
+		$application->setIban( self::PAYMENT_IBAN );
```
JeroenDeDauw: This prompted me to open https://github.com/wmde/FundraisingStore/issues/73

src/DataAccess/DoctrineMembershipApplicationRepository.php
KaiNissen: That is correct, yes. It's possible to enter floats into the form field, which gets floored on insert. We should either restrict the fee to integers by a validator or change the field type accordingly.

JeroenDeDauw: I think it'd be nice if any such shenanigans would be handled by the client side code.

src/DataAccess/DoctrineMembershipApplicationRepository.php
```diff
@@ -0,0 +1,176 @@
...
+
+	public function __construct( EntityManager $entityManager ) {
+		$this->entityManager = $entityManager;
+	}
+
+	public function storeApplication( MembershipApplication $application ) {
+		$doctrineApplication = $this->newDoctrineApplication( $application );
+
+		try {
+			if ( $doctrineApplication->getId() === null ) {
```
gbirke: Is there a particular reason why you set the timestamp manually instead of relying entity mechanisms like in the donation case?
When adding the timestamp on entity creation, we have more consistency across applications.
See https://github.com/wmde/FundraisingStore/blob/master/src/Entities/Donation.php#L131

src/Domain/Repositories/GetMembershipApplicationException.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Domain\Repositories;
+
+/**
+ * @license GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class GetMembershipApplicationException extends \RuntimeException {
```
gbirke: This code made me wonder if the Liskov Substitution Principle applies to constructors, and found the "right" answer on StackOverflow: http://stackoverflow.com/questions/5490824/should-constructors-comply-with-the-liskov-substitution-principle

Might be worth pointing out some times.

tests/Data/ValidMembershipApplication.php
gbirke: Since that PR is merged, you need to change the code, right?

PR #307, opened 2016-04-11
Preserve MembershipApplication creation timestamp

gbirke:
The `creationTime` timestamp was previously updated on every update through `DoctrineMembershipApplicationRepository::storeApplication`. Now it's preserved on update. 
On creation, it's set by the `Timestampable` annotation, no changes there.


JeroenDeDauw:
This is kinda silly and seems like something we should not need to care about here. Still an excellent way to get around the problem: small and hidden from the rest of our code.

PR #313, opened 2016-04-12
Added CreditCardService

JeroenDeDauw:
https://phabricator.wikimedia.org/T130186

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,29 @@
 		"swiftmailer/swiftmailer": "~5.4",
 		"symfony/translation": "^3.0",
 		"symfony/twig-bridge": "^3.0",
-		"twig/extensions": "^1.3"
+		"twig/extensions": "^1.3",
+		"micropayment-de/service-client": "~1.19.0"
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
As part of https://phabricator.wikimedia.org/T130191, the date should stored in `$data['mcp_cc_expiry_date']`

KaiNissen:
The payment provider will provide us with a test account that we can use to write edge-to-edge tests. They also updated their client code library for PHP and fixed a bug that occurs in PHP 7. Example code can be found inside the [library code](https://techdoc.micropayment.de/payment/serviceclient/download/mcp-serviceclient_1_19.zip). Unfortunately there is some Germanness in the example code.

KaiNissen:
Currently the month is stored in two digit format. Not sure, if leaving out the padded '0' can lead to problems when importing our dump to Grün Software's VEWA.

JeroenDeDauw:
It's the responsibility of the persistence service to format the data as required by the persistence.

PR #322, opened 2016-04-14
Tackled repository TODOs

JeroenDeDauw:

PR #323, opened 2016-04-14
 	Changed Donation repo to no longer return deleted donations

JeroenDeDauw:
Includes #322 

Note: I'm not actually sure this specific filter is correct. Though I am thinking we need to filter on several conditions, each which make the Donation no longer relevant for this app. Full list anyone?

### DIFF
--- a/src/DataAccess/DoctrineDonationRepository.php
+++ b/src/DataAccess/DoctrineDonationRepository.php
@@ -207,15 +207,11 @@ class DoctrineDonationRepository implements DonationRepository {
 	 * @return DoctrineDonation|null
 	 * @throws ORMException
 	 */
-	private function getDoctrineDonationById( int $id ) {
-		/*
-		 * FIXME: only select donations that have
-			 * 'isPublic' => true,
-			 * 'dtDel' => null
-			 * correct status
-		 */
-
-		return $this->entityManager->find( DoctrineDonation::class, $id );
+	public function getDoctrineDonationById( int $id ) {
+		return $this->entityManager->getRepository( DoctrineDonation::class )->findOneBy( [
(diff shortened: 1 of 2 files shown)


gbirke:
This needs a rebase, #322 was modified and merged.

JeroenDeDauw:
It can still be merged no? I'll do a rebase anyway though am not sure what this gains us.

src/DataAccess/DoctrineDonationRepository.php
```diff
@@ -207,15 +207,11 @@ public function getDonationById( int $id ) {
...
-			 * 'isPublic' => true,
-			 * 'dtDel' => null
-			 * correct status
-		 */
-
-		return $this->entityManager->find( DoctrineDonation::class, $id );
+	public function getDoctrineDonationById( int $id ) {
+		return $this->entityManager->getRepository( DoctrineDonation::class )->findOneBy( [
+			'id' => $id,
+			'dtDel' => null
```
JeroenDeDauw: This field is about deletion via the backend app right? Or am I confusing it with the cancellation stuff we have in this app?

KaiNissen: Ye Olde Application™ also sets this timestamp, when the dataset is updated with an inactive status (**D**eleted, **P**ending, in**X**omplete, promi**Z**ed bank transfer). I think only the first one really makes sense, the other three don't really relate to a deletion/cancellation.

I think both the deletion in the management/backend application and the cancellation by the user should trigger this timestamp to be set.

JeroenDeDauw: We talked about this and figured that is another task for another PR

PR #325, opened 2016-04-14
Small CS fixes

gbirke:


JeroenDeDauw:
Ugh, this is what happens when being lazy and making edits via the GitHub web UI thing

PR #326, opened 2016-04-14
Validate address data fields on client

gbirke:
"Client-Side" validation of address

* Set up DonationForm template with the correct validator initialization
* Send data to address-validation route on server
* Only send data to address-validation route when required fields are filled.


app/js/lib/reducers/form_content.js
```diff
@@ -25,6 +35,32 @@ function stateContainsUnknownKeys( state ) {
 	return !_.isEmpty( getInvalidKeys( state ) );
 }
 
+function clearFieldsIfAddressTypeChanges( newState, payload ) {
+	if ( payload.contentName !== 'addressType'  ) {
+		return;
+	}
+	switch ( payload.value ) {
+		case 'privat':
```
manicki: Random comment: so those things are not going to be switched to some English equivalents? BTW: if I remember right, the old application used `person` instead of `privat`, along with other address types `firma` and `anonym` which I see are still used below.
Not that the old application should be seen as a reference of what is good. Just pointing this out as my guess is to have those values here "compatible" with those from the old app.

manicki: Few seconds later: wait, I have just noticed above that in ` app/js/lib/reducers/form_content.js` there is line 11:
```js
addressType: 'person', // person, firma and anonym
```
I'm confused. Shouldn't this be `person` instead of `privat` then?

gbirke: I mistakenly put `privat` in here because that's the value from the `title` attribute of the form field. I'll change that in a future commit.

For backwards compatibility in the backend export scripts we still need the German address types.

JeroenDeDauw:
Needs rebase

gbirke:
Rebased.

PR #331, opened 2016-04-18
Shortened names of MembershipApplication cancellation request and response

JeroenDeDauw:
They are namespaced already anyway


KaiNissen:
Okay, I thought it was some kind of unwritten law to name the request class after the use case it is serving. Let me change `HandlePayPalPaymentNotificationRequest` quickly...

JeroenDeDauw:
Yeah, `NotificationRequest` seems clear enough.

We already derivated (how the smag is this spelled?) from this for at least `DisplayPage` and `AddSubscription`

manicki:
> We already derivated (how the smag is this spelled?) 

derived

JeroenDeDauw:
The one I was thinking about is "deviated"

PR #332, opened 2016-04-18
 	Added services to CancelMembershipApplicationUseCase constructor

JeroenDeDauw:
Includes https://github.com/wmde/FundraisingFrontend/pull/331


JeroenDeDauw:
Note that I consider this complete for now even though it does not have tests verifying the correctness. The code is executed so its clear it's not broken. I think it does not make sense to test drive the service injections. Based on previous UCs this strikes me as a waste of time, as you need to keep updating the tests to work with the constructor needing new services, which is cumbersome and is likely to lead to the tests being less clean.

PR #335, opened 2016-04-19
Remove FormFieldAccessor

gbirke:
This class was made superfluous by the FormComponent class and the Redux
architecture.

Removing it, together with the JSDOM library makes the unit test much
faster.


JeroenDeDauw:
+1

PR #338, opened 2016-04-19
Implementation of CancelMembershipApplicationUseCase

JeroenDeDauw:
https://phabricator.wikimedia.org/T123117

### DIFF
--- a/src/UseCases/CancelMembershipApplication/CancelMembershipApplicationUseCase.php
+++ b/src/UseCases/CancelMembershipApplication/CancelMembershipApplicationUseCase.php
@@ -31,6 +31,10 @@ class CancelMembershipApplicationUseCase {
 	}
 
 	public function cancelApplication( CancellationRequest $request ): CancellationResponse {
+		if ( !$this->authorizer->canModifyApplication( $request->getApplicationId() ) ) {
+			return $this->newFailureResponse( $request );
+		}
+
 		$application = $this->getApplicationById( $request->getApplicationId() );
 
 		if ( $application === null ) {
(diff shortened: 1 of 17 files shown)


src/UseCases/CancelMembershipApplication/CancelMembershipApplicationUseCase.php
```diff
@@ -27,7 +31,63 @@ public function __construct( MembershipApplicationAuthorizer $authorizer,
...
+
+		try {
+			$this->repository->storeApplication( $application );
+		}
+		catch ( StoreMembershipApplicationException $ex ) {
+			// TODO: log?
+			return $this->newFailureResponse( $request );
+		}
+
+		$this->sendConfirmationEmail( $application );
```
gbirke: I still think that errors from the mailer should be caught here cause a message on the success page. Something like "We have canceled your application but could not send an email." Otherwise, the user would see an error page (exceptions will be caught by general error handling) and think that the cancellation failed, when in fact it succeeded.

JeroenDeDauw: Can you create a new ticket for that under "new stuff"? (AFAIK this is not needed for feature parity)

JeroenDeDauw:
The Integration test for the UC is using the real persistence, just like we're doing for donations. I've come to think this is bad and will refactor this to use a simple in memory fake after I've added some minimal edge-to-edge tests for the route.

src/UseCases/CancelMembershipApplication/CancelMembershipApplicationUseCase.php
gbirke: https://phabricator.wikimedia.org/T133161
The[ feature is in the old application](https://github.com/wmde/fundraising/blob/master/spenden/contact.inc.php#L334-L342), albeit with a bad wording of the message.

JeroenDeDauw: Meh ok... I still rather first finish more critical things

PR #343, opened 2016-04-20
Show the right buttons at the end of the form

gbirke:
Refactored ElementSlider to be more generic, now it allows different "Animators" - simple show/hide and sliding animations.

Added show/hide of form buttons, depending on payment type.


app/js/lib/view_handler/element_visibility_switcher.js
```diff
@@ -0,0 +1,118 @@
...
+			this.el
+				.slideUp( this.slideSpeed )
+				.animate(
+					{ opacity: 0 },
+					{ queue: false, duration: this.slideSpeed }
+				);
+		}
+	},
+
+	ElementHideAnimator = {
```
JeroenDeDauw: Not sure about the name here. It took me a bit to realize that this class is meant to be of the same type as `ElementSlideAnimator`. That might be somewhat clear if you look at the methods they have, though I started looking at their usage in the factory functions. Both of these do show/hide animations. How about `SlidingElementAnimator` (or `SlideElementAnimator`) and something like `SimpleElementAnimator`? Having the same suffix also makes it more clear they are of the same type. `ElementAnimator` is a lot more general than what these things actually do, so the name is still not very good.

gbirke: Added a commit where I changed the names.

To make the shared type more explicit, I could use an abstract base class. See the branch https://github.com/wmde/FundraisingFrontend/tree/js-inherit-example

Not sure if this actually improves the code.

JeroenDeDauw: Like that you can indicate that the animator field of VisibilitySwitcher needs to be of that type. I'd go for `ElementAnimator` (without `Base` prefix). Do you run into single inheritance problems if you do this in JS?

JeroenDeDauw:
FU: https://github.com/wmde/FundraisingFrontend/pull/346

PR #348, opened 2016-04-21
Make the donation form submittable

gbirke:
Add ViewHandlers that dynamically adjust the functionality of the submit buttons depending on the validation status.

Add TODOs where work needs to be done in case the donation does not validate on the server side (as it is the case at the moment, because the amount is not sent properly).

### DIFF
--- a/app/js/tests/view_handlers/test_value_dependent_event_handler.js
+++ b/app/js/tests/view_handlers/test_value_dependent_event_handler.js
@@ -1,0 +1,67 @@
+'use strict';
+
+var test = require( 'tape' ),
+	sinon = require( 'sinon' ),
+	createValueDependentEventHandler = require( '../../lib/view_handler/value_dependent_event_handler' ).createHandler;
+
+test( 'When the handler is created it attaches it attaches an event listener', function ( t ) {
+	var eventHandlerConfig = [],
+		eventName = 'click',
+		element = {
+			on: sinon.spy()
+		},
+		onArguments;
+
+	createValueDependentEventHandler( element, eventName, eventHandlerConfig );
+	t.ok( element.on.calledOnce, 'An event handler was attached' );
+	onArguments = element.on.args[ 0 ];
+	t.equals( onArguments[ 0 ], eventName );
+	t.equals( typeof onArguments[ 1 ], 'function' );
+	t.end();
+
+} );
+
+test( 'Given a matching value, only the configured event handler is triggered', function ( t ) {
+	var fooHandler = sinon.spy(),
+		barHandler = sinon.spy(),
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
> the functionality of the submit buttons

I had a confused there for a second :)

app/templates/Form1.html.twig
```diff
@@ -1,4 +1,4 @@
-<form name="donForm" id="donForm" method="POST" action="#">
+<form name="donForm2" id="donForm1" method="POST" action="#">
```
JeroenDeDauw: ?

src/Factories/FunFunFactory.php
```diff
@@ -852,7 +852,11 @@ public function getDonationConfirmationPageSelector() {
 	}
 
 	public function newDonationFormViolationPresenter() {
-		return new DonationFormViolationPresenter( $this->getLayoutTemplate( 'DonationForm.html.twig' ) );
+		$template = $this->getLayoutTemplate( 'DisplayPageLayout.twig' );
+		// TODO make this dependent on the 'form' value from the HTTP POST request
+		// (we need different form pages for A/B testing)
```
JeroenDeDauw: We have stuff in place already for showing different donation confirmation pages. There no HTTP parameter is used. Does that make sense?

app/templates/DonationForm.html.twig
```diff
@@ -142,6 +163,27 @@
                     {
                         viewHandler: WMDE.View.createSlidingVisibilitySwitcher( $( '.personal-data-full' ), /firma|person/ ),
                         stateKey: 'formContent.addressType'
+                    },
+                    {
+                        viewHandler: WMDE.View.createValueDependentEventHandler( $( '#continueFormSubmit1' ), 'click', [
```
JeroenDeDauw: What's the deal with this 1 and 2 stuff? The first and second steps on the client?

app/js/tests/view_handlers/test_value_dependent_event_handler.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+var test = require( 'tape' ),
+	sinon = require( 'sinon' ),
+	createValueDependentEventHandler = require( '../../lib/view_handler/value_dependent_event_handler' ).createHandler;
+
+test( 'When the handler is created it attaches it attaches an event listener', function ( t ) {
```
JeroenDeDauw: 2x it attaches

gbirke:
The JS classes must be rewritten and one commit was cherry-picked and modified into another commit, so I'm closing this to avoid confusion.

PR #351, opened 2016-04-22
make donation cancellation confirmation page look nice

KaiNissen:


JeroenDeDauw:
+1

gbirke:
I made a pull request how this could be handled without adding a parameter to the presenter and "re-/mis-using" the template that was only meant for the display page use case. But it's just a stylistic thing, @JeroenDeDauw  could also go ahead and merge this as-is.

JeroenDeDauw:
Still +1 from my side

PR #356, opened 2016-04-24
Implemented ApplyForMembershipUseCase

JeroenDeDauw:
https://phabricator.wikimedia.org/T130199


JeroenDeDauw:
The presentation still needs to be done (email and presenter). Unless I'm forgetting about something, I think we can move the remaining tasks (exception handling) to a new ticket with lower priority, once the presentation stuff is done.

PR #354, opened 2016-04-23
add data sent with PayPal's IPN

KaiNissen:

PR #357, opened 2016-04-25
Make Response generation for PPL UC more clear.

gbirke:
Followup to #354 and #355, just clearing up something that confused me while reviewing.

### DIFF
--- a/app/RouteHandlers/PayPalNotificationHandler.php
+++ b/app/RouteHandlers/PayPalNotificationHandler.php
@@ -61,13 +61,13 @@ class PayPalNotificationHandler {
 					->setPaymentType( $post->get( 'payment_type', '' ) )
 			);
 
-			return new Response( '', 200 );
+			return new Response( '', Response::HTTP_OK ); # PayPal expects an empty response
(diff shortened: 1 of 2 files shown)


app/RouteHandlers/PayPalNotificationHandler.php
```diff
@@ -61,13 +61,13 @@ public function handle( Request $request ): Response {
 					->setPaymentType( $post->get( 'payment_type', '' ) )
 			);
 
-			return new Response( '', 200 );
+			return new Response( '', Response::HTTP_OK ); # PayPal expects an empty response
 		} catch ( PayPalPaymentNotificationVerifierException $e ) {
 			// TODO: log error
 			// TODO: let PayPal resend IPN?
+			// TODO: is this the right Response?
```
KaiNissen: Not sure. PayPal will resend the IPN over and over again if the returned status code is something other than 200. And it will be failing over and over again, because the data that is verified will never change. The old application is doing this, though.

I think, we should send back 200, but notify the fundraising team for further investigation.

PR #358, opened 2016-04-25
add missing ext_payment_* fields

KaiNissen:
last missing bits of [phab:T130118](https://phabricator.wikimedia.org/T130118)


tests/EdgeToEdge/Routes/HandlePayPalPaymentNotificationRouteTest.php
```diff
@@ -152,7 +163,11 @@ private function newRequest() {
 			'address_name' => 'address_name',
 			'item_name' => self::ITEM_NAME,
 			'item_number' => 1,
-			'custom' => '{"id": "1", "utoken": "my_secret_token"}'
+			'custom' => '{"id": "1", "utoken": "my_secret_token"}',
+			'txn_id' => 'transaction_id',
+			'payment_type' => 'transaction_id',
```
JeroenDeDauw: I think it's nicer to have realistic values here, at least type wise. If it normally is a number, then putting a string might cause different behaviour.

PR #365, opened 2016-04-27
Add JS component for entering and validating bank data

gbirke:


app/js/lib/form_validation.js
```diff
@@ -40,37 +40,74 @@ var jQuery = require( 'jquery' ),
...
-			postFunction: postFunction || jQuery.post
+			sendFunction: sendFunction || jQuery.post
+		} );
+	},
+
+	createBankDataValidator = function ( validationUrlForSepa, validationUrlForNonSepa, sendFunction ) {
+		return objectAssign( Object.create( BankDataValidator ), {
+			validationUrlForSepa: validationUrlForSepa,
+			validationUrlForNonSepa: validationUrlForNonSepa,
+			sendFunction: sendFunction || jQuery.get
```
KaiNissen: Is there a reason for using GET here, and POST when validating the amount?

app/js/lib/view_handler/display_bank_data.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+	_ = require( 'lodash' ),
+	BankDataDisplayHandler = {
+		accountNumber: null,
+		bankCode: null,
+		bankName: null,
+		update: function ( formContent ) {
+			if ( formContent.debitType === 'sepa' ) {
```
KaiNissen: The validator returns all the bank data (including non-sepa data when sepa data was entered and vice-versa). They could all be set here, otherwise the server-side validation will (unnecessarily) try to complement the missing data.

KaiNissen: Oh, this is just the pre-confirmation display. Please ignore my comment above...

app/js/lib/view_handler/display_bank_data.js
```diff
@@ -0,0 +1,34 @@
...
+	};
+
+module.exports = {
+	createBankDataDisplayHandler: function ( elementConfig ) {
+		var expectedConfigProperties = [
+				'accountNumber',
+				'bankCode',
+				'bankName'
+			],
+			uncofiguredElements = _.difference( expectedConfigProperties, _.keys( elementConfig ) );
```
KaiNissen: There's a typo in this variable's name.

app/js/lib/form_validation.js
gbirke: Don't know. The route just accepts GET, but I could also change the route instead of jQuery method.

app/js/tests/test_form_validation.js
```diff
@@ -95,3 +95,86 @@ test( 'Given an incomplete private adddress, address validation sends no values
...
+		bankCode: '50010517',
+		bankName: 'ING-DiBa',
+		debitType: 'sepa',
+		paymentType: 'BEZ'
+	} );
+
+	t.ok( postFunctionSpy.calledOnce, 'data is sent once' );
+	callParameters = postFunctionSpy.getCall( 0 ).args;
+	t.equals( callParameters[ 0 ], 'http://spenden.wikimedia.org/check-iban', 'validation calls URL for SEPA' );
+	t.deepEquals( callParameters[ 1 ], { iban: 'DE12500105170648489890' }, 'validation sends only necessary data' );
```
KaiNissen: Wait, is `deepEquals()` an alias for `deepEqual()`? The qunit docs don't mention `deepEquals()`.

app/js/lib/form_validation.js
KaiNissen: I don't have a strong opinion about that. Maybe we should use POST in all cases to ensure potential "sensitive banners" can use the validators as well. But that is actually out of scope of this PR.

app/js/tests/test_form_validation.js
gbirke: Seems like it's aliased (and `equals` too). But the docs only list the versions without s, so I'll make a commit that restores consistency.

PR #367, opened 2016-04-28
Add bank data display to SEPA confirmation

gbirke:
Refactored view handler and added constructor calls to factory.


JeroenDeDauw:
+1

PR #368, opened 2016-05-02
Validate submit buttons

gbirke:
Validate all the submit buttons

Please merge #367 first

### DIFF
--- a/app/templates/DonationForm.html.twig
+++ b/app/templates/DonationForm.html.twig
@@ -190,14 +190,31 @@
 
         // connect DOM elements to actions
 
-        $( '#continueFormSubmit1,#continueFormSubmit2' ).click( function () {
-            // TODO check if page is valid
-            store.dispatch( actions.newNextPageAction() );
+        $( '#continueFormSubmit1' ).click( function () {
+            if ( store.getState().validity.amount ) {
+                store.dispatch( actions.newNextPageAction() );
+            } else {
+                alert( 'Bitte füllen Sie das Formular komplett aus.' );
+            }
+        } );
+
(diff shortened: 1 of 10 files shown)


app/templates/DonationForm.html.twig
```diff
@@ -175,21 +183,54 @@
...
         // connect DOM elements to actions
 
-        $( '#continueFormSubmit1,#continueFormSubmit2' ).click( function () {
-            // TODO check if page is valid
-            store.dispatch( actions.newNextPageAction() );
+        $( '#continueFormSubmit1' ).click( function () {
+            if ( store.getState().validity.amount ) {
+                store.dispatch( actions.newNextPageAction() );
+            } else {
+                alert( 'Bitte füllen Sie das Formular komplett aus.' );
```
KaiNissen: In my opinion, calling `alert()` is not so nice. Do you plan to replace this later?

gbirke: I agree, the better way would be to scroll up to the validation message. **But** if the user just tries to send the form without doing anything where is no validation message. So the click handlers would have to add an additional check check if validity === null and trigger the validation. However, since most of the validation is done server side which may take some time, we'd need to signal to the user that there is some validation in progress (e.g. spinner on/beside the submit button, etc).

Would you be willing to merge this if I create a new task for improving on the alerts?

KaiNissen: :+1:

gbirke: https://phabricator.wikimedia.org/T134166
https://phabricator.wikimedia.org/T134168

PR #369, opened 2016-05-03
Display amount

gbirke:
Display payment period and currency-formatted amount in donation form.


app/js/lib/simple_currency_formatter.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+/**
+ * This number formatter class is a primitive subset of the intl library ( https://www.npmjs.com/package/intl ),
+ * which is not available in all browsers and node versions.
+ */
+
+var parseGermanFloat = function ( amountStr ) {
```
JeroenDeDauw: perhaps put the first var name on the next line as well, so its clear this is a pile of assigns, and not a single one?

app/js/lib/simple_currency_formatter.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+/**
+ * This number formatter class is a primitive subset of the intl library ( https://www.npmjs.com/package/intl ),
+ * which is not available in all browsers and node versions.
```
JeroenDeDauw: Is this something you wrote or something copied from elsewhere? If the later, we should probably specify from where, and what license it has.

JeroenDeDauw:
+1

app/js/lib/simple_currency_formatter.js
```diff
@@ -0,0 +1,37 @@
...
+		format: function ( amountStr ) {
+			var amount = this.parse( amountStr );
+			return String( amount.toFixed( 2 ) ).replace( '.', ',' ) + String.fromCharCode( 160 ) + '€';
+		},
+		parse: parseGermanFloat
+	},
+	EnglishCurrencyFormatter = {
+		format: function ( amountStr ) {
+			var amount = this.parse( amountStr );
+			return String( amount.toFixed( 2 ) ) + String.fromCharCode( 160 ) + '€';
```
manicki: Correct me if I'm wrong but isn't it the common convention in English to put the major currency symbol/abbreviation before the amount, ie. to write `EUR 666` instead of `666 EUR`? (I've only used EUR instead of euro sign because I have never learned how you type it). At least this is my experience. It could be a bit different with Euro, though.
I've just checked with couple of (German-based) webstores I use often and they all put either EUR or euro sign before the amount when using English UI.

Do we have an expert on such things in the room?

gbirke: Your'e right. I'd think the best approach would be to emulate the locale behavior of the JS  `Intl` library as close as possible. For the English locale, it puts the euro sign in the front. I'll change this.

app/js/lib/simple_currency_formatter.js
gbirke: Added newlines between the assignments, should be better now.

PR #370, opened 2016-05-03
Membership form client code

KaiNissen:


app/js/lib/form_validation.js
```diff
@@ -56,7 +56,7 @@ var jQuery = require( 'jquery' ),
 		sendFunction: null,
 		validate: function ( formValues ) {
 			var data, validationUrl;
-			if ( formValues.paymentType !== 'BEZ' ) {
+			if ( formValues.paymentType && formValues.paymentType !== 'BEZ' ) {
```
JeroenDeDauw: Ugh. Such a JS!

JeroenDeDauw:
+1 to first two commits. Think the third one is better reviewed by @gbirke

app/js/lib/view_handler/fee_option_switcher.js
```diff
@@ -0,0 +1,76 @@
...
+					return;
+				}
+				this.toggler.enableElement();
+				this.isDisabled = false;
+			}
+		}
+	},
+
+	ElementAvailabilityToggler = {
+		element: null,
```
gbirke: An abstract class plus an implementing class are not strictly necessary. You probably got this idea from the visibility switcher where we have different methods of switching the visibility (animated and non-animated). Since the availability is always toggled the same way, you could put the code from `SimpleElementAvailabilityToggler` directly into `FeeOptionSwitcher`

KaiNissen: Into `ElementAvailabilityToggler` you mean? Because if it's in `FeeOptionSwitcher` I can't inject a toggler spy anymore.

gbirke: Nope, into `FeeOptionSwitcher` and use an `ElementSpy` instead of a toggler spy.

KaiNissen: Yes, you're right. Fixed a type related bug and corrected the tests as well.

KaiNissen:
This can actually already be merged as is. Removing [WIP] label.

web/skin/10h16/_js/form-methods.js
```diff
@@ -49,31 +49,11 @@ $( document ).ready( function() {
...
 	});
 
-	$( '#memFormSubmit' ).on( 'click', function( e ) {
-		if ( memCheckPending ) {
-			checkMembershipFee( true );
-			return false;
-		}
-	});
-
 	$( "#email" ).on( 'change', function( evt ) {
```
gbirke: This should also be removed. Email is checked together with address data.

KaiNissen: Actually it seems, that the whole file can be removed.

PR #371, opened 2016-05-03
Disallow anonymous donations for BEZ

gbirke:
When doing a direct debit donation, force using a person/firma address
type.

Added comments for the visibility view handlers

### DIFF
--- a/app/js/lib/reducers/form_content.js
+++ b/app/js/lib/reducers/form_content.js
@@ -40,6 +40,14 @@ function clearFieldsIfAddressTypeChanges( newState, payload ) {
 	}
 }
 
+function forcePersonalDataForDirectDebit( state ) {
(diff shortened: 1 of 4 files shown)


app/js/lib/reducers/form_content.js
```diff
@@ -40,6 +40,14 @@ function clearFieldsIfAddressTypeChanges( newState, payload ) {
 	}
 }
 
+function forcePersonalDataForDirectDebit( state ) {
+	if ( state.paymentType === 'BEZ' && state.addressType === 'anonym' ) {
+		return objectAssign( {}, state, { addressType: 'person' } );
+	} else {
```
JeroenDeDauw: You can drop the `else` as the if already forms a guard.

PR #372, opened 2016-05-03
Added logging decorators for our repositories

JeroenDeDauw:
https://phabricator.wikimedia.org/T133549

I only created a test for LoggingDonationRepository as I'm not really happy with it. The decorator code is really simple and unlikely will change at some future point. So I'm not sure the maintenance cost of the tests is worth it. They are for sure more complex and verbose than the production code itself.

Question on alternative approach at http://stackoverflow.com/questions/36992540/type-safe-generic-decorator-in-php, though lets not get hung up on perfection here and move things along.


gbirke:
It is intentional not to use the new classes in the `FunFunFactory`?

JeroenDeDauw:
Yeah, that's a different step, in which we presumably want to figure out here to log what.

PR #376, opened 2016-05-04
Fix anonymous donation validation

gbirke:
This fixes validation errors when donating anonymously (payment type is
not BEZ).

### DIFF
--- a/app/templates/DonationForm.html.twig
+++ b/app/templates/DonationForm.html.twig
@@ -222,6 +222,14 @@
 
         // connect DOM elements to actions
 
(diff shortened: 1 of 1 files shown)


app/templates/DonationForm.html.twig
```diff
@@ -222,6 +222,14 @@
 
         // connect DOM elements to actions
 
+        function formPage2IsValid() {
```
KaiNissen: The method name could be more descriptive. Maybe something like `donationDataIsValid()` or `donationDataIsComplete()`?

PR #381, opened 2016-05-10
Initial work on MembershipApplicationValidator

JeroenDeDauw:
As per mailing list discussion, using request model instead of domain object. Also not binding to HTML field names.


src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
```diff
@@ -0,0 +1,64 @@
...
+
+	private function addViolation( string $source, string $type ) {
+		$this->violations[$source] = $type;
+	}
+
+	private function validateAmountMeetsYearlyMinimum( Euro $amount ) {
+		$yearlyAmount = $amount->getEuroFloat() * self::MONTHS_PER_YEAR / $this->request->getPaymentIntervalInMonths();
+
+		if ( $yearlyAmount < self::MIN_YEARLY_PAYMENT_IN_EURO ) {
+			// TODO: also add the violation to SOURCE_PAYMENT_INTERVAL?
```
JeroenDeDauw: Do we want this? (Do we want to also display an error indicator by the interval field?)

gbirke: I think the amount field is enough.

gbirke:
I like this approach. It means more work/code on the client side or in the presenter since the ValidationResult has to be mapped to error messages and HTML fields, but overall I like it.

src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
```diff
@@ -0,0 +1,64 @@
...
+use WMDE\Fundraising\Frontend\Domain\Model\Euro;
+use WMDE\Fundraising\Frontend\UseCases\ApplyForMembership\ApplicationValidationResult as Result;
+
+/**
+ * @license GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class MembershipApplicationValidator {
+
+	const MIN_YEARLY_PAYMENT_IN_EURO = 25;
```
KaiNissen: The current minimum yearly fee is €24 for a private person and €100 for institutions/companies.

JeroenDeDauw: Thanks, I'll change the code accordingly

JeroenDeDauw:
> more work/code on the client side or in the presenter since the ValidationResult has to be mapped to error messages and HTML fields

I was thinking that this would go in the presenter. Then this approach looks the same to the client as for the other validators. And if we pass the constants the client, then it binds to their values, which seems a bit odd. Plus the i18n is on the server side. Just convinced myself it should be the presenter :)

src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
JeroenDeDauw: Done in https://github.com/wmde/FundraisingFrontend/pull/385

PR #383, opened 2016-05-10
fix IPN verification

KaiNissen:


app/routes.php
```diff
@@ -377,6 +377,30 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
 );
 
 $app->post(
+	'show-donation-confirmation',
```
JeroenDeDauw: huh? You added it a second time?!

JeroenDeDauw:
+1 except for the double route all the way across the sky part.

app/routes.php
KaiNissen: It has to accept GET as well as POST. But you're right, there are smarter ways to do this apart from copying all the stuff.

PR #385, opened 2016-05-10
Finished amount validation in MembershipApplicationValidator

JeroenDeDauw:
https://phabricator.wikimedia.org/T130202

### DIFF
--- a/src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
+++ b/src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
@@ -14,8 +14,9 @@ use WMDE\Fundraising\Frontend\UseCases\ApplyForMembership\ApplicationValidationR
  */
 class MembershipApplicationValidator {
 
-	const MIN_YEARLY_PAYMENT_IN_EURO = 25;
-	const MONTHS_PER_YEAR = 12;
+	/* private */ const MIN_PERSON_YEARLY_PAYMENT_IN_EURO = 24;
+	/* private */ const MIN_COMPANY_YEARLY_PAYMENT_IN_EURO = 100;
+	/* private */ const MONTHS_PER_YEAR = 12;
 
 	/**
 	 * @var ApplyForMembershipRequest
@@ -53,12 +54,19 @@ class MembershipApplicationValidator {
 	}
 
 	private function validateAmountMeetsYearlyMinimum( Euro $amount ) {
-		$yearlyAmount = $amount->getEuroFloat() * self::MONTHS_PER_YEAR / $this->request->getPaymentIntervalInMonths();
-
-		if ( $yearlyAmount < self::MIN_YEARLY_PAYMENT_IN_EURO ) {
-			// TODO: also add the violation to SOURCE_PAYMENT_INTERVAL?
+		if ( $this->getYearlyPaymentAmount( $amount ) < $this->getYearlyPaymentRequirement() ) {
 			$this->addViolation( Result::SOURCE_PAYMENT_AMOUNT, Result::VIOLATION_TOO_LOW );
 		}
+	}
+
+	private function getYearlyPaymentAmount( Euro $amount ): float {
+		return $amount->getEuroFloat() * self::MONTHS_PER_YEAR / $this->request->getPaymentIntervalInMonths();
+	}
+
+	private function getYearlyPaymentRequirement(): float {
+		return $this->request->isCompanyApplication() ?
+			self::MIN_COMPANY_YEARLY_PAYMENT_IN_EURO :
+			self::MIN_PERSON_YEARLY_PAYMENT_IN_EURO;
 	}
 
 }
--- a/tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
+++ b/tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
@@ -36,10 +36,8 @@ class MembershipApplicationValidatorTest extends \PHPUnit_Framework_TestCase {
 	/**
 	 * @dataProvider invalidAmountProvider
 	 */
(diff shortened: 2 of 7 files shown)


src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
```diff
@@ -14,8 +14,9 @@
  */
 class MembershipApplicationValidator {
 
-	const MIN_YEARLY_PAYMENT_IN_EURO = 25;
-	const MONTHS_PER_YEAR = 12;
+	/* private */ const MIN_PERSON_YEARLY_PAYMENT_IN_EURO = 24;
+	/* private */ const MIN_COMPANY_YEARLY_PAYMENT_IN_EURO = 100;
+	/* private */ const MONTHS_PER_YEAR = 12;
```
gbirke: I see, preparing for PHP 7.1 ;-)

src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
```diff
@@ -53,12 +54,20 @@ private function addViolation( string $source, string $type ) {
 	}
 
 	private function validateAmountMeetsYearlyMinimum( Euro $amount ) {
-		$yearlyAmount = $amount->getEuroFloat() * self::MONTHS_PER_YEAR / $this->request->getPaymentIntervalInMonths();
-
-		if ( $yearlyAmount < self::MIN_YEARLY_PAYMENT_IN_EURO ) {
+		if ( $this->getYearlyPaymentAmount( $amount ) < $this->getYearlyPaymentRequirement() ) {
 			// TODO: also add the violation to SOURCE_PAYMENT_INTERVAL?
```
gbirke: You can remove the todo

tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
```diff
@@ -50,12 +48,57 @@ public function testGivenInvalidAmount_validationFails( string $amount, string $
 
 	public function invalidAmountProvider() {
 		return [
-			'too low' => [ '1.00', Result::VIOLATION_TOO_LOW ],
-			'just too low' => [ '24.99', Result::VIOLATION_TOO_LOW ],
-			'max too low' => [ '0', Result::VIOLATION_TOO_LOW ],
-			'invalid: negative' => [ '-1.00', Result::VIOLATION_NOT_MONEY ],
-			'invalid' => [ 'y u no btc', Result::VIOLATION_NOT_MONEY ],
+			'invalid: negative' => [ '-1.00', 3, Result::VIOLATION_NOT_MONEY ],
```
gbirke: Why is the Violation "Not money" instead of "too low"?

tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
```diff
@@ -50,12 +48,57 @@ public function testGivenInvalidAmount_validationFails( string $amount, string $
...
+
+		$this->assertTrue( $this->newValidator()->validate( $request )->isSuccessful() );
+	}
+
+	public function validAmountProvider() {
+		return [
+			'single payment' => [ '50.00', 12 ],
+			'just enough single payment' => [ '24.00', 12 ],
+			'high single payment' => [ '31333.37', 12 ],
+
```
gbirke: These are amounts with decimal points instead of comma. Should't we be consistent with the donation page and expect comma-formatted amounts (depending on locale)?

And maybe you can add an integer amount which should also be valid.

src/UseCases/ApplyForMembership/MembershipApplicationValidator.php
JeroenDeDauw: I found that I had used a constant I should not have used at some place, and decided to specify my intend properly rather than wait for a new version of the language. When 7.1 comes out, we can go through all the constants and make then private if appropriate in little time.

tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
JeroenDeDauw: It's not money in the sense that you cannot give someone -1 EUR, which is a thing our Euro class checks.

tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php
JeroenDeDauw: I did not put a lot of thought into this yet, though was writing this code as if localization stuff was already taken care off before. Though if it was, we would presumably be dealing with Euro instances here already. That raises the question of where to do the localization stuff. I really don't want to have to deal with different kinds of notations and PHP types here. However we cannot just construct Euro instances beforehand, since those will already throw exceptions for input errors we want to have caught by the validator. This suggests doing localization stuff in the route that takes as input a string and outputs a string in the PHP float format. Do you see any problems with that?

JeroenDeDauw: The moment you realize AmountParser and AmountFormatter are in different directories

bark

JeroenDeDauw: We should also figure this out for the add donation case. There the AmountParser will end up with casting `false` to a float, and then that being given to Euro::newFromFloat. You'll later get an error cause the amount is too low, though that is not ideal.

I suggest to continue with the validator here as is for now, as these are two disjoint tasks.

gbirke: I've already encountered the `false` to float case while checking the donation. I then created a task for that: https://phabricator.wikimedia.org/T134864

PR #387, opened 2016-05-11
Mark application for moderation when amount is high

JeroenDeDauw:
https://phabricator.wikimedia.org/T130202

### DIFF
--- a/src/Domain/Model/MembershipPayment.php
+++ b/src/Domain/Model/MembershipPayment.php
@@ -14,10 +14,25 @@ class MembershipPayment {
 	private $amount;
 	private $bankData;
 
+	/**
+	 * @param int $intervalInMonths
+	 * @param Euro $amount
+	 * @param BankData $bankData
+	 *
+	 * @throws \InvalidArgumentException
+	 */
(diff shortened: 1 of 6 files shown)


src/Domain/Model/MembershipPayment.php
```diff
@@ -14,12 +14,27 @@ class MembershipPayment {
...
+		$this->assertIsValidInterval( $intervalInMonths );
+
 		$this->interval = $intervalInMonths;
 		$this->amount = $amount;
 		$this->bankData = $bankData;
 	}
 
+	private function assertIsValidInterval( int $intervalInMonths ) {
+		if ( !in_array( $intervalInMonths, [ 1, 2, 3, 4, 6, 12 ] ) ) {
+			throw new \InvalidArgumentException( 'Interval needs to be 1, 2, 3, 4, 6 or 12' );
```
JeroenDeDauw: We can do this right?

KaiNissen: sure, interval should neither be 2 nor 4, though. :)

PR #392, opened 2016-05-11
Refactor Mail functionality

gbirke:


gbirke:
Two questions:

@JeroenDeDauw: Is it ok to subclass `TemplateBasedMailer` and use the new derived class where needed? Or should I rather inject the `TemplateBasedMailer` into `DonationConfirmationMailer`?

@KaiNissen: I got rid of the `$needsmoderation` parameter. Please have a look if this is correct.

tests/Unit/Domain/Model/DonationTest.php
```diff
@@ -75,6 +75,13 @@ public function testGivenNewStatus_cancellationSucceeds() {
 		$this->assertSame( Donation::STATUS_CANCELLED, $donation->getStatus() );
 	}
 
+	public function testModerationStatusCanBeQueried() {
```
JeroenDeDauw: That's some weird name!

JeroenDeDauw:
First two commits look good.

About the inheritance thing... I'm not sure why you are extracting this code form the UC in the first place. Not saying it's bad - I can think of some advantages. Though it matters what you want to do with the new class. Do you want to use it from some other UC? Do you want to add dedicated tests for it? If both are no, then I'd just put the new class in the UC dir, switch to composition, and have the UC construct it itself, leaving this change invisible from the outside. Anyway, composition is very likely the better approach.

gbirke:
The refactoring is part of preparing for the task [Confirm donations by e-mail during callback handling triggered by payment providers]( https://phabricator.wikimedia.org/T134373 ). The new class will be used in at least 3 different use cases.

gbirke:
Will change the code form inheritance to composition, now that I realized that testing it will be easier.

tests/Integration/UseCases/AddDonation/AddDonationUseCaseTest.php
```diff
@@ -168,7 +168,7 @@ private function newInvalidDonationRequest(): AddDonationRequest {
 	public function testGivenInvalidRequest_noConfirmationEmailIsSend() {
 		$mailer = $this->newMailer();
 
-		$mailer->expects( $this->never() )->method( 'sendMail' );
+		$mailer->expects( $this->never() )->method( 'sendMailFromDonation' );
```
JeroenDeDauw: `$this->any()` binds to less stuff

`sendMailFromDonation` comes over as an odd name to me. You can say "from donation" when constructing a mail... in this case it kinda says the mail is coming from the donation tho. How about "for donation"? Or perhaps just `sendMail`, or `sendConfirmationMail`?

tests/Unit/Infrastructure/DonationConfirmationMailerTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+	public function testMailerExtractsEmailFromDonation() {
+		$donation = ValidDonation::newBankTransferDonation();
+		$templateMailer = $this->getMockBuilder( TemplateBasedMailer::class )->disableOriginalConstructor()->getMock();
+		$templateMailer->expects( $this->once() )
+			->method( 'sendMail' )
+			->with(
+				$this->equalTo( ValidDonation::DONOR_EMAIL_ADDRESS ),
+				$this->anything()
+			);
```
JeroenDeDauw: There is a mailer spy which is probably nicer here

tests/Unit/Infrastructure/DonationConfirmationMailerTest.php
```diff
@@ -0,0 +1,55 @@
...
+		$mailer = new DonationConfirmationMailer( $templateMailer );
+		$mailer->sendMailFromDonation( $donation );
+	}
+
+	public function testMailerAssemblesTemplateData() {
+		$donation = ValidDonation::newBankTransferDonation();
```
JeroenDeDauw: inline?

tests/Unit/Infrastructure/DonationConfirmationMailerTest.php
```diff
@@ -0,0 +1,55 @@
...
+		$expectedTemplateData = [
+			'recipient' => [
+				'lastName' => ValidDonation::DONOR_LAST_NAME,
+				'title' => ValidDonation::DONOR_TITLE,
+				'salutation' => ValidDonation::DONOR_SALUTATION
+			],
+			'donation' => [
+				'id' => $donation->getId(),
```
JeroenDeDauw: donation does not have an id, so this is null

tests/Unit/Infrastructure/DonationConfirmationMailerTest.php
```diff
@@ -0,0 +1,55 @@
...
+				'amount' => ValidDonation::DONATION_AMOUNT,
+				'interval' => ValidDonation::PAYMENT_INTERVAL_IN_MONTHS,
+				'needsModeration' => false,
+				'paymentType' => PaymentType::BANK_TRANSFER,
```
JeroenDeDauw: Not in ValidDonation?

PR #395, opened 2016-05-12
Improved DoctrineDonationRepositoryTest

JeroenDeDauw:
Includes a free TRACKING_TRACKING constant and DataData method!

### DIFF
--- a/src/DataAccess/DoctrineDonationRepository.php
+++ b/src/DataAccess/DoctrineDonationRepository.php
@@ -125,7 +125,7 @@ class DoctrineDonationRepository implements DonationRepository {
 		return array_merge(
 			$this->getDataFieldsFromTrackingInfo( $donation->getTrackingInfo() ),
 			$this->getDataFieldsForPaymentData( $donation->getPaymentMethod() ),
-			$this->getDataFieldsFromPersonalInfo( $donation->getDonor() )
+			$this->getDataFieldsFromDonor( $donation->getDonor() )
 		);
 	}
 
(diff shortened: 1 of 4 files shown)


manicki:
`TRACKING_TRACKING` is cool indeed. However, if alternative names were about to be considered, the value of it is supposed a combination of "campaign code" (I'd also call it a "test code") and "banner code" (campaign codes are skipped in some cases if I remember right, correct?). `TRACKING_CAMPAIGN_AND_BANNER_CODE` is getting verbose. Wouldn't calling it a `TRACKING_CODE` be going too far? That name doesn't explain much either.
So, as in the current donation campaign environment (that is rather unlikely too change I think) particular banners are always bound to campaigns (even if there are same-looking banners in multiple campaigns they have different codes/ids), maybe `TRACKING_BANNER_CODE`? or `TRACKING_BANNER_ID`?

Or maybe just let it be `TRACKING_TRACKING`, I dont know.

manicki:
Just realized that `TRACKING_BANNER_CODE` would not be specific enough. There are other things than banners that link to donation site and are tracked with a `tracking`. E.g. there is a "Spenden Sie jetzt!" link on https://www.wikimedia.de/wiki/Hauptseite and when a person follows it and donates, the donation gets a `tracking` "wikimedia_de_sidebar/Spendenaufruf".
So I am out of ideas. Sorry for elaborating and good luck @KaiNissen @gbirke with coming up with a name for tracking tracking!

gbirke:
Since the key in the data unfortunately really is `tracking` and applies to tracking data, I'm fine with `TRACKING_TRACKING` 😄

gbirke:
@KaiNissen told me he doesn't like it but doesn't care enough. So I'll just mergeMerge it.

PR #404, opened 2016-05-13
Refactored DonationConfirmationMailerTest

JeroenDeDauw:


tests/Unit/Infrastructure/DonationConfirmationMailerTest.php
```diff
@@ -4,52 +4,61 @@
...
+		);
+	}
+
+	private function newDonation(): Donation {
 		$donation = ValidDonation::newBankTransferDonation();
-		$templateMailer = $this->getMockBuilder( TemplateBasedMailer::class )->disableOriginalConstructor()->getMock();
-		$templateMailer->expects( $this->once() )
-			->method( 'sendMail' )
-			->with(
-				$this->equalTo( ValidDonation::DONOR_EMAIL_ADDRESS ),
```
JeroenDeDauw: Fun fact... this passes even though the argument is actually an `EmailAddress` object, as this check is not strict and casts the thing to string

PR #407, opened 2016-05-13
Fix address validation

gbirke:
Make client-side address validation fail if not all fields are filled in.


KaiNissen:
see a460e93 :)

Do you want me to remove that commit from the PR or should we just close this?

PR #411, opened 2016-05-16
Added logging decorator for mailer

JeroenDeDauw:
https://phabricator.wikimedia.org/T135135

### DIFF
--- a/src/Infrastructure/LoggingMailer.php
+++ b/src/Infrastructure/LoggingMailer.php
@@ -1,0 +1,42 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use WMDE\Fundraising\Frontend\Domain\Model\EmailAddress;
+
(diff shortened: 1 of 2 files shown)


src/Infrastructure/LoggingMailer.php
```diff
@@ -0,0 +1,42 @@
...
+
+	/**
+	 * @throws \RuntimeException
+	 */
+	public function sendMail( EmailAddress $recipient, array $templateArguments = [] ) {
+		try {
+			$this->mailer->sendMail( $recipient, $templateArguments );
+		}
+		catch ( \RuntimeException $ex ) {
+			$this->logger->log( $this->logLevel, $ex->getMessage(), [ self::CONTEXT_EXCEPTION_KEY => $ex ] );
```
JeroenDeDauw: This will just log `"Message delivery failed"` as message. Do we want to include something else as well? (If yes, creating the commit is not much more work then writing it here ;p)

gbirke: No, I think it's fine, since at the moment the error logger contains the the whole stack trace from the exception. If the error occurs, we can see which method(s) it comes from.

PR #417, opened 2016-05-16
Added MembershipApplicationTracker interface

JeroenDeDauw:
Start of https://phabricator.wikimedia.org/T130116

### DIFF
--- a/src/Infrastructure/MembershipApplicationTrackingInfo.php
+++ b/src/Infrastructure/MembershipApplicationTrackingInfo.php
@@ -1,0 +1,26 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class MembershipApplicationTrackingInfo {
+
+	private $totalImpressionCount;
+	private $singleBannerImpressionCount;
+	private $confirmationPage;
+	private $confirmationPageCampaign;
+
+	public function __construct( int $totalImpressionCount, int $singleBannerImpressionCount,
+		string $confirmationPage, string $confirmationPageCampaign ) {
+		$this->totalImpressionCount = $totalImpressionCount;
+		$this->singleBannerImpressionCount = $singleBannerImpressionCount;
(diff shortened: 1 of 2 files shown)


src/Infrastructure/MembershipApplicationTrackingInfo.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class MembershipApplicationTrackingInfo {
+
+	private $totalImpressionCount;
+	private $singleBannerImpressionCount;
+	private $confirmationPage;
+	private $confirmationPageCampaign;
```
JeroenDeDauw: I'm actually quite fuzzy about all four of these fields, so need some explanation to make sure they have sane names

KaiNissen: - `$singleBannerImpressionCount`: number of times the current banner was shown (before the user decided to donate)
- `$totalImpressionCount`: total number of times any banner was shown
- `$confirmationPageCampaign` and `$confirmationPage`: similar to the banner tracking campaign and keyword; confirmation pages may embed modified membership forms to a/b test against

The first two don't need to be tracked, as this is actually only relevant for the donation tracking. The latter two can/should be combined into one (`campaign/keyword`).

JeroenDeDauw: > The latter two can/should be combined into one (campaign/keyword).

Can we do that easily, or is that an improvement we better hold off until we have the app functional?

JeroenDeDauw: Ping pong @KaiNissen

KaiNissen: It's generated when confirming a donation and needs to be put into that embedded membership application form. Since I've been working a lot around the membership forms, I can do that.

JeroenDeDauw: Could you also modify the value objected added by this PR to reflect how you think it should be? (You can do that without doing the other thing first.)

KaiNissen: see #450

PR #419, opened 2016-05-17
Always dispatch validation finish action

gbirke:
The "null" state, originally introduced to avoid error messages showing
up, was too confusing and leads to the wrong value in `validity` when
a field is made empty after the form has been validated.

Incomplete forms that should not be sent to the server should return an
"INCOMPLETE" status (validity store checks for an "OK" status) and no
error messages.


JeroenDeDauw:
+1

PR #420, opened 2016-05-17
Fix LoggingCommentRepository

gbirke:
Was broken due to earlier refactoring (CommentFinder was moved).


JeroenDeDauw:
Derp. Ofc this happen when not writing any tests that execute the thing...

PR #422, opened 2016-05-17
Create Paypal child donations

gbirke:
This is for https://phabricator.wikimedia.org/T135232

Plz merge #423 first, I'll then rework this PR.

### DIFF
--- a/src/Domain/Model/PayPalData.php
+++ b/src/Domain/Model/PayPalData.php
@@ -28,6 +28,7 @@ class PayPalData {
 	private $paymentType = '';
 	private $paymentStatus = '';
 	private $paymentTimestamp = '';
+	private $childPayments = [];
 
 	public function __construct() {
 		$this->amount = Euro::newFromInt( 0 );
@@ -183,6 +184,14 @@ class PayPalData {
 		$this->assertIsWritable();
 		$this->paymentTimestamp = $paymentTimestamp;
 		return $this;
+	}
+
+	public function addChildPayment( string $paymentId, int $entityId ) {
+		$this->childPayments[$paymentId] = $entityId;
+		return $this;
+	}
+	public function hasChildPayment( string $paymentId ): bool {
+		return isset( $this->childPayments[$paymentId] );
 	}
 
 }
--- a/src/Domain/Model/PayPalData.php
+++ b/src/Domain/Model/PayPalData.php
@@ -194,4 +194,8 @@ class PayPalData {
 		return isset( $this->childPayments[$paymentId] );
 	}
 
+	public function getChildPaymentEntityId( string $paymentId ): int {
+		return $this->childPayments[$paymentId];
+	}
+
 }
--- a/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
+++ b/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
@@ -70,8 +70,8 @@ class HandlePayPalPaymentNotificationUseCase {
 		}
 
 		if ( $this->donationWasBookedWithDifferentTransactionId( $donation, $request ) ) {
-			$this->createChildDonation( $donation, $request );
-			return true;
+			$childDonation = $this->createChildDonation( $donation, $request );
+			return $childDonation !== null;
 		}
 
 		$donation->addPayPalData( $this->newPayPalDataFromRequest( $request ) );
@@ -158,6 +158,10 @@ class HandlePayPalPaymentNotificationUseCase {
 			return false;
 		}
 
+		if ( $payment->getPayPalData()->hasChildPayment( $request->getTransactionId() ) ) {
+			return false;
+		}
+
 		return true;
 	}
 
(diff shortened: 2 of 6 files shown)


src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -131,4 +143,51 @@ private function newPayPalDataFromRequest( PayPalNotificationRequest $request ):
...
+	}
+
+	private function createChildDonation( Donation $donation, PayPalNotificationRequest $request ) {
+		$childPaymentMethod = new PayPalPayment( $this->newPayPalDataFromRequest( $request ) );
+		$payment = $donation->getPayment();
+		$childDonation = new Donation(
+			null,
+			Donation::STATUS_EXTERNAL_BOOKED,
+			$donation->getDonor(),
+			new DonationPayment( $payment->getAmount(), $payment->getIntervalInMonths(), $childPaymentMethod ),
```
KaiNissen: The amount should be taken from `$request`, since it might have been changed by the donor.

src/Domain/Model/PayPalData.php
```diff
@@ -185,4 +186,16 @@ public function setPaymentTimestamp( string $paymentTimestamp ) {
 		return $this;
 	}
 
+	public function addChildPayment( string $paymentId, int $entityId ) {
```
KaiNissen: : self

src/Domain/Model/PayPalData.php
```diff
@@ -185,4 +186,16 @@ public function setPaymentTimestamp( string $paymentTimestamp ) {
...
+		$this->childPayments[$paymentId] = $entityId;
+		return $this;
+	}
+	public function hasChildPayment( string $paymentId ): bool {
+		return isset( $this->childPayments[$paymentId] );
+	}
+
+	public function getChildPaymentEntityId( string $paymentId ) {
```
KaiNissen: : int

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -131,4 +143,51 @@ private function newPayPalDataFromRequest( PayPalNotificationRequest $request ):
...
+			$donation->getOptsIntoNewsletter(), $donation->getTrackingInfo()
+		);
+		try {
+			$this->repository->storeDonation( $childDonation );
+		} catch ( StoreDonationException $ex ) {
+			return;
```
KaiNissen: Should we pass this exception back to the route? We should make PayPal resend the notification, right?

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -131,4 +143,51 @@ private function newPayPalDataFromRequest( PayPalNotificationRequest $request ):
...
+		}
+		/** @var PayPalPayment $paymentMethod */
+		$paymentMethod = $payment->getPaymentMethod();
+		$paymentMethod->getPayPalData()->addChildPayment( $request->getTransactionId(), $childDonation->getId() );
+		try {
+			$this->repository->storeDonation( $donation );
```
KaiNissen: Should we store the child and the parent donation in a single transaction?

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
gbirke: The old code [copies the values from the parent donation](https://github.com/wmde/fundraising/blob/master/spenden/paypal_handler.php#L227-L240), I just emulated that behavior. This is *not* for the recurring donations, where there will be  extra code for that, where interval is set to `0` and amount is taken from `$request`.

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
gbirke: So far, `DoctrineDonationRepository:storeDonation` does not support transactions with multiple donations. And in this case, that would not be useful IMHO, since we need to insert the child donation in the DB to generate its ID which then will be stored in the transaction data of the parent.

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
KaiNissen: Sorry, I mistook that for the handling of recurring donations.

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
KaiNissen: Right.

KaiNissen:
There you are: :+1: :wink:

I'm willing to merge, please rebase.

PR #425, opened 2016-05-24
Validity indicator

KaiNissen:
see [phab:T135823](https://phabricator.wikimedia.org/T135823)

### DIFF
--- a/app/js/lib/reducers/membership_input_validation.js
+++ b/app/js/lib/reducers/membership_input_validation.js
@@ -1,19 +1,25 @@
 'use strict';
 
-var inputValidationLib = require( './input_validation' ),
+var _ = require( 'lodash' ),
+	inputValidationLib = require( './input_validation' ),
+	objectFields = {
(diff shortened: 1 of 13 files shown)


app/js/lib/reducers/membership_input_validation.js
```diff
@@ -0,0 +1,33 @@
...
+	};
+
+module.exports = function membershipInputValidation( state, action ) {
+	if ( typeof state === 'undefined' ) {
+		state = initialState;
+	}
+	switch ( action.type ) {
+		default:
+			return inputValidationLib.inputValidation( state, action );
+	}
```
gbirke: You can remove the switch around the return statement, it does nothing.

KaiNissen: Oh, right.

gbirke:
Works fine. The per-country validation of postal codes is missing, but that can be done later.

PR #426, opened 2016-05-24
changed template parameter names

KaiNissen:
The previously used parameter names were sort of placeholders. I changed them to represent the model used in the donation confirmation page presenter.


JeroenDeDauw:
+1

PR #429, opened 2016-05-24
Handle Paypal payments without donation

gbirke:


src/Domain/Model/PersonName.php
```diff
@@ -29,11 +29,11 @@ private function __construct( string $nameType ) {
 		$this->personType = $nameType;
 	}
 
-	public static function newPrivatePersonName() {
+	public static function newPrivatePersonName(): PersonName {
```
JeroenDeDauw: `: self`

JeroenDeDauw:
```
FILE: ...alPaymentNotification/HandlePayPalPaymentNotificationUseCase.php

----------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

----------------------------------------------------------------------

 42 | ERROR | Function's cyclomatic complexity (11) exceeds allowed

    |       | maximum of 10

    |       | (Generic.Metrics.CyclomaticComplexity.MaxExceeded)

----------------------------------------------------------------------
```

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -190,4 +198,44 @@ private function createChildDonation( Donation $donation, PayPalNotificationRequ
...
+	}
+
+	private function newDonationFromRequest( PayPalNotificationRequest $request ): Donation {
+		$payment = new DonationPayment( $request->getAmountGross(), 0, new PayPalPayment() );
+		$donation = new Donation(
+			null,
+			Donation::STATUS_EXTERNAL_BOOKED,
+			$this->newDonorFromRequest( $request ),
+			$payment,
+			false,
```
JeroenDeDauw: `Donation::DOES_NOT_OPT_INTO_NEWSLETTER`

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -190,4 +198,44 @@ private function createChildDonation( Donation $donation, PayPalNotificationRequ
...
+			$this->newPhysicalAddressFromRequest( $request ),
+			$request->getPayerEmail()
+		);
+	}
+
+	private function newPersonNameFromRequest( PayPalNotificationRequest $request ): PersonName {
+		$name = PersonName::newPrivatePersonName();
+		$name->setFirstName( $request->getPayerFirstName() );
+		$name->setLastName( $request->getPayerLastName() );
+		$name->freeze();
```
JeroenDeDauw: Also `->assertNoNullFields`? (And yeah I know, PhpStorm broke itself in one of the last updates when it comes to autocompleting that)

tests/Integration/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCaseTest.php
```diff
@@ -288,6 +277,67 @@ public function testGivenTransactionIdInBookedChildDonation_noNewDonationIsCreat
...
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_EMAIL, $donation->getDonor()->getEmailAddress() );
+
+		$address = $donation->getDonor()->getPhysicalAddress();
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_ADDRESS_STREET, $address->getStreetAddress() );
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_ADDRESS_CITY, $address->getCity() );
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_ADDRESS_POSTAL_CODE, $address->getPostalCode() );
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_ADDRESS_COUNTRY_CODE, $address->getCountryCode() );
+
+		$payment = $donation->getPayment();
+		$this->assertEquals( ValidPayPalNotificationRequest::AMOUNT_GROSS_CENTS, $payment->getAmount()->getEuroCents() );
```
JeroenDeDauw: All of those assertEquals can be put into a more high level assertion method, so the details don't clutter the test method

Also don't forget that when comparing scalar values, `assertSame` is better

tests/Integration/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCaseTest.php
```diff
@@ -288,6 +277,67 @@ public function testGivenTransactionIdInBookedChildDonation_noNewDonationIsCreat
...
+
+		/** @var PayPalData $paypalData */
+		$paypalData = $payment->getPaymentMethod()->getPaypalData();
+		$this->assertEquals( ValidPayPalNotificationRequest::PAYER_ADDRESS_NAME, $paypalData->getAddressName() );
+
+	}
+
+	public function testWhenNotificationIsForNonExistingDonation_confirmationMailIsSent() {
+		$repositorySpy = new DonationRepositorySpy();
```
JeroenDeDauw: Can be inlined

tests/Integration/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCaseTest.php
```diff
@@ -288,6 +277,67 @@ public function testGivenTransactionIdInBookedChildDonation_noNewDonationIsCreat
...
+				$repositorySpy,
+			new SucceedingDonationAuthorizer(),
+			$this->getMailer(),
+			new NullLogger()
+		);
+
+		$useCase->handleNotification( $request );
+
+		$storeDonationCalls = $repositorySpy->getStoreDonationCalls();
+		$this->assertCount( 1, $storeDonationCalls, 'Donation is stored' );
```
JeroenDeDauw: This assert stuff can also be extracted

gbirke:
The big bad method could be split into three parts with three submethods
- check if this notification is handled at all (or just logged).
- handle notifications without existing donation
- handle notification with existing donation.

That would't reduce the cyclomatic complexity of the overall class but it would make the flow  in `handleNotification` more clear.

What do you think, @JeroenDeDauw ?

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
gbirke: You mean to catch the unlikely case that a `null` value comes from the request? Is that even possible withe all the type checking? 

Or should I put `->assertNoNullFields` in to guard against some change in the future where the internal values of newly instantiated person objects are no longer empty strings and `null` values instead?

I's a bit reluctant to do it since that means another try/catch block ...

JeroenDeDauw:
If you see a sane way to split it, and sounds like you do, then go for it

src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
JeroenDeDauw: The `assertNoNullFields` is part of normal `FreezableValueObject` initialization. As those do not have a real constructor, and instead require you to call setters, it's easy to forget to call a setter. And one can add a new field, and then not have the existing code that is wrongly not calling the new setter still work, though incorrectly.

If there is a field that can be null for valid reasons, then this does not work. Should probably do something else in that case, but I figure we have more pressing matters to get to.

JeroenDeDauw:
I'm fine with merging as is and extracting the test stuff afterwards. Preference?

JeroenDeDauw:
Let's address the remaining comments after merge

PR #436, opened 2016-05-26
Membership application confirmation

KaiNissen:


JeroenDeDauw:
+1

app/routes.php
```diff
@@ -392,7 +392,7 @@ function( Application $app, Request $httpRequest ) use ( $ffFactory ) {
 
 		$response = $ffFactory->newApplyForMembershipUseCase()->applyForMembership( $request );
 
-		return $response->isSuccessful() ? 'TODO success' : 'TODO fail'; // TODO
+		return $response->isSuccessful() ? $ffFactory->newMembershipApplicationConfirmationHtmlPresenter()->present( $response ) : 'TODO fail'; // TODO
```
gbirke: Handling failed here is a bit tricky since the request can come from the membership page or form the donation confirmation page. The right page needs to be displayed here, right? Or should we just display the pre-filled membership page? Then we'd still have to disable the "Aktive Mitglieder". 

But handling failure is outside the scope of this PR, right?

PR #442, opened 2016-05-27
Membership fail presenter

KaiNissen:


src/UseCases/ApplyForMembership/ApplyForMembershipRequest.php
```diff
@@ -45,6 +46,15 @@ public function setMembershipType( string $membershipType ) {
 		$this->membershipType = $membershipType;
 	}
 
+	public function getShowMembershipTypeOption(): bool {
```
JeroenDeDauw: This is presentation specific and (thus) not used by the UC, so should not be in here

KaiNissen: Right. This is directly passed to the presenter now.

src/Presentation/Presenters/MembershipFormViolationPresenter.php
```diff
@@ -0,0 +1,53 @@
...
+
+	public function __construct( TwigTemplate $template ) {
+		$this->template = $template;
+	}
+
+	public function present( array $violations, ApplyForMembershipRequest $request, bool $showMembershipTypeOption ): string {
+		return $this->template->render(
+			[
+				'initialFormValues' => $this->getMembershipFormArguments( $request ),
+				'violatedFields' => $violations,
```
JeroenDeDauw: Like this the template binds to stuff in the Validators NS. Kinda iffy

JeroenDeDauw:
+1

KaiNissen:
removed `violatedFields`, rebased

PR #447, opened 2016-05-29
Decrease PHPMD failures

JeroenDeDauw:
The fatal PHPMD error occurs only for the tests, with this
we can more easily run it against just src/.

### DIFF
--- a/phpmd.xml
+++ b/phpmd.xml
@@ -19,7 +19,13 @@
 
     <rule ref="rulesets/naming.xml">
         <exclude name="ShortVariable" />
+        <exclude name="LongVariable" />
     </rule>
 
     <rule ref="rulesets/unusedcode.xml" />
+
+    <exclude-pattern>*Validator</exclude-pattern>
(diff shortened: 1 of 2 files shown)


phpmd.xml
```diff
@@ -19,7 +19,13 @@
...
         <exclude name="ShortVariable" />
+        <exclude name="LongVariable" />
     </rule>
 
     <rule ref="rulesets/unusedcode.xml" />
+
+    <exclude-pattern>*Validator</exclude-pattern>
+    <exclude-pattern>*Factory.php</exclude-pattern>
+    <exclude-pattern>ConfigReader.php</exclude-pattern>
+    <exclude-pattern>ListCommentsUseCase.php</exclude-pattern>
```
JeroenDeDauw: Works though perhaps not correct: https://github.com/phpmd/phpmd/issues/373#issuecomment-222363580

gbirke: Why do you exclude the validators and the `ListCommentsUseCase`?

PR #450, opened 2016-05-30
membership application tracking info

KaiNissen:


src/Infrastructure/MembershipApplicationTrackingInfo.php
```diff
@@ -7,20 +7,19 @@
...
-		$this->singleBannerImpressionCount = $singleBannerImpressionCount;
-		$this->confirmationPage = $confirmationPage;
-		$this->confirmationPageCampaign = $confirmationPageCampaign;
+	public function __construct( string $campaignCode, string $keyword ) {
+		$this->campaignCode = $campaignCode;
+		$this->keyword = $keyword;
 	}
 
+	public function getInfo() {
+		return implode( '/', array_filter( [ $this->campaignCode, $this->keyword ] ) );
```
JeroenDeDauw: This class is meant to be a value object. This transformation is either something that belongs into the presentation layer or the data access one, I'm not sure which one.

PR #455, opened 2016-06-01
Persist membership tracking

KaiNissen:
based on #450 and thus #417

### DIFF
--- a/src/DataAccess/DoctrineMembershipApplicationRepository.php
+++ b/src/DataAccess/DoctrineMembershipApplicationRepository.php
@@ -20,6 +20,7 @@ use WMDE\Fundraising\Frontend\Domain\Model\PhysicalAddress;
 use WMDE\Fundraising\Frontend\Domain\Repositories\GetMembershipApplicationException;
 use WMDE\Fundraising\Frontend\Domain\Repositories\MembershipApplicationRepository;
 use WMDE\Fundraising\Frontend\Domain\Repositories\StoreMembershipApplicationException;
+use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationTrackingInfo;
 
 /**
  * @license GNU GPL v2+
--- a/src/DataAccess/DoctrineMembershipApplicationTracker.php
+++ b/src/DataAccess/DoctrineMembershipApplicationTracker.php
@@ -1,0 +1,67 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\DataAccess;
+
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\ORMException;
+use WMDE\Fundraising\Entities\MembershipApplication;
+use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationTracker;
+use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationTrackingException;
+use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationTrackingInfo;
+
+/**
+ * @licence GNU GPL v2+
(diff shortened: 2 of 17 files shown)


src/DataAccess/DoctrineMembershipApplicationRepository.php
```diff
@@ -122,6 +124,16 @@ private function setPaymentFields( DoctrineApplication $application, MembershipP
...
 
+	private function setTrackingField( DoctrineApplication $application, MembershipApplicationTrackingInfo $trackingInfo ) {
+		if ( $trackingInfo->getKeyword() === '' ) {
+			$application->setTracking( '' );
+		} else {
+			$application->setTracking(
+				implode( '/', array_filter( [ $trackingInfo->getCampaignCode(), $trackingInfo->getKeyword() ] ) )
+			);
+		}
+	}
```
JeroenDeDauw: Since this is just a single field, and $application->setTracking is always called, I'd have gone with `$application->setTracking( getTrackingField( $trackingInfo ) )`. Is slightly simpler

src/Domain/Model/MembershipApplication.php
```diff
@@ -30,24 +32,27 @@ class MembershipApplication {
...
 
 	public function __construct( int $id = null, string $type, MembershipApplicant $applicant, MembershipPayment $payment,
-		bool $needsModeration, bool $isCancelled ) {
+		MembershipApplicationTrackingInfo $trackingInfo, bool $needsModeration, bool $isCancelled ) {
 
 		$this->id = $id;
 		$this->type = $type;
 		$this->applicant = $applicant;
 		$this->payment = $payment;
+		$this->trackingInfo = $trackingInfo;
```
JeroenDeDauw: You did not create the field. More importantly though, I'm surprised you are adding this here. I thought the plan was to have the tracking info handled by a service?

KaiNissen: Err, right, the `MembershipApplicationTracker` interface. I must have been daydreaming when I merged #417... While implementing the tracking data persistence, I wondered why we want a service for that. Just like any other data the tracking stuff comes as a request parameter and can be handled as such.

Maybe I am missing something here, what's the advantage in having this handled by a service?

KaiNissen: There it is: Tracking as a Service

src/Infrastructure/MembershipApplicationTracker.php
```diff
@@ -7,10 +7,13 @@
...
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
  */
 interface MembershipApplicationTracker {
 
-	// TODO: exception
-	public function trackApplication( MembershipApplicationTrackingInfo $trackingInfo );
+	/**
+	 * @throws MembershipApplicationTracker
```
JeroenDeDauw: MembershipApplicationTrackingException

src/DataAccess/DoctrineMembershipApplicationTracker.php
```diff
@@ -0,0 +1,67 @@
...
+		try {
+			$this->entityManager->persist( $application );
+			$this->entityManager->flush();
+		}
+		catch ( ORMException $ex ) {
+			throw new MembershipApplicationTrackingException( 'Failed to persist membership application' );
+		}
+	}
+
+	private function concatenate( MembershipApplicationTrackingInfo $trackingInfo ) {
```
JeroenDeDauw: : string

JeroenDeDauw:
Comments addressed in https://github.com/wmde/FundraisingFrontend/pull/458

PR #460, opened 2016-06-02
Made empty value check more strict

JeroenDeDauw:
Not 100% sure if it is too strict now. For instance, is this code intended to
handle null as an empty value? I can't think of any places where we not always
put a string into this...

The old behaviour was definitly not good tho. It would for instance identify
the string "0" as being empty. (yay PHP!)

### DIFF
--- a/src/Validation/RequiredFieldValidator.php
+++ b/src/Validation/RequiredFieldValidator.php
@@ -11,7 +11,7 @@ namespace WMDE\Fundraising\Frontend\Validation;
 class RequiredFieldValidator {
 
 	public function validate( $value ): ValidationResult {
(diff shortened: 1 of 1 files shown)


KaiNissen:
I see your point and agree that we should change it. This requires us to explicitely set values in a request model to an empty string if e. g. missing in a HTTP request, though. What about checking for string emptiness like you propose but also for `is_string()`? That would eliminate the behaviour of `null` resulting in a successful validation.

JeroenDeDauw:
`is_string` is not needed, as `===` checks the type as well.

The request should indeed get empty string set explicitly. That's already the case though, as we either have normal constructors or use `assertNoNullFields`

PR #463, opened 2016-06-03
CC payment handler route

KaiNissen:
https://phabricator.wikimedia.org/T130191


src/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCase.php
```diff
@@ -0,0 +1,73 @@
...
+use WMDE\Fundraising\Frontend\Domain\Repositories\GetDonationException;
+use WMDE\Fundraising\Frontend\Infrastructure\DonationAuthorizer;
+use WMDE\Fundraising\Frontend\Infrastructure\DonationConfirmationMailer;
+use WMDE\Fundraising\Frontend\Infrastructure\DonationEventLogger;
+
+/**
+ * @license GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class HandleCreditCardPaymentNotificationUseCase {
```
JeroenDeDauw: How about `CreditCardNotificationUseCase`? For the NS I'd leave `Payment` in, though am dubious about the `Handle` part. That's what all UCs do right, handle a request :)

src/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCase.php
```diff
@@ -0,0 +1,73 @@
...
+		if ( !$this->authorizationService->systemCanModifyDonation( $request->getDonationId() ) ) {
+			return false;
+		}
+
+		return $this->handleRequest( $request, $donation );
+	}
+
+	private function handleRequest( CreditCardPaymentNotificationRequest $request, Donation $donation ): bool {
+		$this->sendConfirmationEmail( $donation );
+		return true;
```
JeroenDeDauw: Why not simply return true in the parent function?

src/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCase.php
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		if ( !$this->authorizationService->systemCanModifyDonation( $request->getDonationId() ) ) {
+			return false;
+		}
+
+		return $this->handleRequest( $request, $donation );
+	}
+
+	private function handleRequest( CreditCardPaymentNotificationRequest $request, Donation $donation ): bool {
```
JeroenDeDauw: $request is unused

tests/Integration/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCaseTest.php
```diff
@@ -0,0 +1,145 @@
...
+class HandleCreditCardPaymentNotificationUseCaseTest extends \PHPUnit_Framework_TestCase {
+
+	public function testWhenRepositoryThrowsException_handlerReturnsFalse() {
+		$useCase = new HandleCreditCardPaymentNotificationUseCase(
+			new DoctrineDonationRepository( ThrowingEntityManager::newInstance( $this ) ),
+			new FailingDonationAuthorizer(),
+			$this->getMailer(),
+			new NullLogger(),
+			$this->getEventLogger()
+		);
```
JeroenDeDauw: I recommend following the pattern used here: https://github.com/wmde/FundraisingFrontend/blob/master/tests/Unit/UseCases/ApplyForMembership/MembershipApplicationValidatorTest.php#L27

* Have a setUp that initializes all services, each having a field
* Each service has minimal behaviour for the "success case"
* Individual test methods call a newWhateverThing method and don't know how the class under test is constructed
* Before calling newWhateverThing they set services as needed for the behaviour being tested (almost always a single one)

tests/Integration/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCaseTest.php
```diff
@@ -0,0 +1,145 @@
...
+
+		$this->assertTrue( $useCase->handleNotification( $request ) );
+	}
+
+	/**
+	 * @return DonationConfirmationMailer|\PHPUnit_Framework_MockObject_MockObject
+	 */
+	private function getMailer(): DonationConfirmationMailer {
```
JeroenDeDauw: Personally I always use `newFoo` when a new instance is returned

tests/EdgeToEdge/Routes/HandleCreditCardPaymentNotificationRouteTest.php
```diff
@@ -23,8 +39,67 @@ public function testRouteExists() {
...
+		$this->assertSame( $request['amount'], $ccData->getAmount()->getEuroCents() );
+		$this->assertSame( $request['country'], $ccData->getCountryCode() );
+		$this->assertSame( $request['auth'], $ccData->getAuthId() );
+		$this->assertSame( $request['title'], $ccData->getTitle() );
+		$this->assertSame( $request['sessionId'], $ccData->getSessionId() );
+		$this->assertSame( $request['transactionId'], $ccData->getTransactionId() );
+		#$this->assertSame( $request['status'], $ccData->getTransactionStatus() );
+		$this->assertSame( $request['customerId'], $ccData->getCustomerId() );
+		#$this->assertSame( $request['mcp_cc_expiry_date'], $ccData->getExpiryDate() );
+		#$this->assertSame( $request['ext_payment_timestamp'], $ccData->getTransactionTimestamp() );
```
JeroenDeDauw: What's the deal with those? Is this commit finished, or do you intend to amend?

JeroenDeDauw:
+2 to all but the last commit (which also looks good)

src/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCase.php
KaiNissen: This one not just handles a request, but a payment notification request!! Sure, the names are becoming hilarious. Let me change this.

src/UseCases/HandleCreditCardPaymentNotification/HandleCreditCardPaymentNotificationUseCase.php
KaiNissen: There was still something missing here.

tests/EdgeToEdge/Routes/HandleCreditCardPaymentNotificationRouteTest.php
KaiNissen: did this in another commit

src/Domain/Model/CreditCardTransactionData.php
```diff
@@ -90,6 +92,23 @@ public function setCustomerId( string $customerId ): self {
...
+	 */
+	public function getCardExpiry() {
+		return $this->cardExpiry;
+	}
+
+	/**
+	 * @param CreditCardExpiry|null $cardExpiry
+	 * @return self
+	 */
+	public function setCardExpiry( $cardExpiry ) {
```
JeroenDeDauw: Does this need to be nullable?

src/Infrastructure/CreditCardExpiry.php
```diff
@@ -22,6 +22,15 @@ public function __construct( int $month, int $year ) {
 		$this->year = $year;
 	}
 
+	public static function newFromString( string $expirationDate ) {
+		$dateParts = explode( '/', $expirationDate );
+		if ( count( $dateParts ) === 2 ) {
+			return new self( (int)$dateParts[0], (int)$dateParts[1] );
+		}
+
+		return null;
```
JeroenDeDauw: Do we really want to silently ignore invalid expiration dates and introduce null values into the system?

That question depends on context. Ignoring the context, I think it's nicer to throw exceptions in alternative constructors when they cannot construct an instance. Ofc the context question is more important.

KaiNissen: We might not receive that data from the payment provider's service API. Not making it nullable means that we can never set the transaction data without it.

JeroenDeDauw: Ok :( Evil payment provider API!

PR #464, opened 2016-06-05
Bump PHPUnit req to 5.4

JeroenDeDauw:
Older versions of PHPUnit will fail after recent changes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,19 +48,20 @@
 	"require-dev": {
 		"ext-pdo_sqlite": "*",
 
-		"phpunit/phpunit": "~5.2",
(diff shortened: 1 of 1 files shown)


gbirke:
What about the 54 deprecated calls to `getMock` (which should be converted to calls to `createMock`)?

KaiNissen:
Already tackled in an [earlier commit](https://github.com/wmde/FundraisingFrontend/pull/462/commits/e38a5ab7ff9ac3ffa08af7dc9e55bef198bbb652).

gbirke:
Ah, forget it, I saw that the calls were already converted and the remaining 54 were to mockBuilder instances.

PR #466, opened 2016-06-07
redirect from old URLs

KaiNissen:
[phab:T123054](https://phabricator.wikimedia.org/T123054)

### DIFF
--- a/app/RouteHandlers/RouteRedirectionHandler.php
+++ b/app/RouteHandlers/RouteRedirectionHandler.php
@@ -1,0 +1,36 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
+1

gbirke:
Do we need to support any GET/POST parameters for the the `/spenden/?foo==bar` and `/spenden?foo=bar` cases, e.g. for tracking? I know it's legacy support, but still ...

KaiNissen:
I reworked this a little and added support for preserving the query string.

gbirke:
When I tested on my local VM (that uses Nginx) and on the test server, it turned out that we still have to do some server-side rewrite, otherwise the Nginx rules will try to access a nonexisting PHP file.

Did a commit for it in the infrastucture repo.

PR #469, opened 2016-06-07
Created CreditCardNotificationHandler

JeroenDeDauw:
Part of https://phabricator.wikimedia.org/T130185


JeroenDeDauw:
Well not really needed now... since we won't add new stuff to it. So not sure if this improves things or not

PR #471, opened 2016-06-07
Log to donation when booking externally

gbirke:
When a donation is booked through an external provider, add the booking
event to its event log.

Refactored tryToLogDonationEvent to its own class. Otherwise there would
be duplicate code in CreditCardNotificationUseCase and
HandlePayPalPaymentNotificationUseCase.

This is for https://phabricator.wikimedia.org/T135528


JeroenDeDauw:
Looks good. +1. Kinda tired so not merging at present

src/Factories/FunFunFactory.php
```diff
@@ -354,7 +355,10 @@ public function getEntityManager(): EntityManager {
 	}
 
 	private function newDonationEventLogger(): DonationEventLogger {
-		return new DoctrineDonationEventLogger( $this->getEntityManager() );
+		return new BestEffortDonationEventLogger(
```
JeroenDeDauw: Think this name is fine. I probably would have gone with FallbackDonationEventLogger

JeroenDeDauw:
+2 when follow up is merged: https://github.com/wmde/FundraisingFrontend/pull/481

JeroenDeDauw:
Could already have merged this :)

PR #477, opened 2016-06-08
Fix AddDonationresponse::getAccessToken

gbirke:
The response returned the wrong token.

### DIFF
--- a/src/UseCases/AddDonation/AddDonationResponse.php
+++ b/src/UseCases/AddDonation/AddDonationResponse.php
@@ -83,7 +83,7 @@ class AddDonationResponse {
 	 * @return string|null
 	 */
 	public function getAccessToken() {
-		return $this->updateToken;
+		return $this->accessToken;
 	}
 
 }


gbirke:
I’d really like to fix the test that should have cought this (`testWhenAdditionWorks_successResponseContainsTokens`), but it’s hard: We are using `FixedTokenGenerator` which can only return one token value. The options for testing I can come up with aren’t good:

1) Create a mock token generator or `SequenceTokenGenerator` Fixture that returns different tokens. Problem: we are tying the test to the sequence of the calls in `AddDonationUseCase::addDonation` which should be a private implementation detail.

2) two different token generators for accessToken and updateToken. Problem: This would be just for testing and would blow up the already huge constructor for `AddDonationUseCase`.

3) Separate token generators into different variables in the  `AddDonationUseCase` constructor and add a setter for them that is used in testing. Problem: Funny setter that is confusing because it’s only used for testing.


Any better ideas? If not, I’d go with number 3.

JeroenDeDauw:
Uuuuuuugh. I knew we where using the same token in the test and found it too hard to change that, figuring this is unlikely a thing to get f*cked up anyway. Ofc its broken :)

I will look at this again later today.

KaiNissen:
argh...

gbirke:
Even when this is merged, there should be some unit test coverage (which is not the case at the moment).

JeroenDeDauw:
Actually given that we can't think of a good way to go about this, I'm inclined to leave it as-is. A getter like this can have a copy-paste error at the beginning, but once it is correct, it is very unlikely to become incorrect.

PR #478, opened 2016-06-08
Per country validation

KaiNissen:
https://phabricator.wikimedia.org/T136107


app/js/lib/country_specifics_repository.js
```diff
@@ -0,0 +1,68 @@
...
+			email: {
+				placeholder: 'z. B. name@domain.ch'
+			}
+		}
+	};
+
+module.exports = {
+	getCountrySpecifics: function ( countryCode ) {
+		if ( _.has( countrySpecifics, countryCode ) ) {
+			return _.extend( {}, countrySpecifics.generic, countrySpecifics[ countryCode ] );
```
JeroenDeDauw: Why is the `{}` needed?

app/js/lib/country_specifics_repository.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+var _ = require( 'underscore' ),
+	countrySpecifics = {
+		generic: {
```
JeroenDeDauw: This could be outside countrySpecifics (and it is not actually country specific)

app/js/lib/reducers/country_specifics.js
```diff
@@ -0,0 +1,22 @@
...
+	if ( typeof state !== 'object' ) {
+		newState = getCountrySpecifics( 'DE' );
+	} else {
+		newState = Object.assign( {}, state );
+	}
+
+	switch ( action.type ) {
+		case 'UPDATE_ELEMENT_ATTRIBUTES':
+			newState = getCountrySpecifics( action.payload.countryCode );
+			return newState;
```
JeroenDeDauw: return getCountrySpecifics( action.payload.countryCode );

app/js/lib/reducers/country_specifics.js
```diff
@@ -0,0 +1,22 @@
...
+		default:
+			return newState;
```
JeroenDeDauw: The first code before the switch block only gets used here. So why not have a guard statement that returns the stuff from the previous comment, followed by this code before the switch block? Looks like you don't need a single assignment in this function.

app/js/lib/view_handler/attribute_value_handler.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+	_ = require( 'underscore' ),
+
+	AttributeValueHandler = {
+		'post-code': null,
+		city: null,
+		email: null,
```
JeroenDeDauw: That's not very consistent

app/js/lib/view_handler/attribute_value_handler.js
```diff
@@ -0,0 +1,33 @@
...
+			} );
+		}
+	};
+
+module.exports = {
+	createAttributeValueHandler: function ( postCodeElement, cityElement, emailElement ) {
+		return objectAssign( Object.create( AttributeValueHandler ), {
+			'post-code': postCodeElement,
+			city: cityElement,
+			email: emailElement
```
JeroenDeDauw: Same here

JeroenDeDauw:
+1

app/js/lib/country_specifics_repository.js
gbirke: To create a new object instead of overriding/extending `countrySpecifics.generic`.

gbirke: Another question: why use `extend` at all? All the fields in the country-specific data override the generic data, there is no need for defaults.

app/js/lib/view_handler/attribute_value_handler.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+	_ = require( 'underscore' ),
+
+	AttributeValueHandler = {
```
gbirke: The code is very generic, making `AttributeValueHandler` an appropriate name. However the constructor is very specific to the fields affected by country code. Please change either the name of the class or the constructor.

app/js/lib/view_handler/attribute_value_handler.js
```diff
@@ -0,0 +1,33 @@
...
+		'post-code': null,
+		city: null,
+		email: null,
+
+		update: function ( elements ) {
+			var self = this;
+			_.each( elements, function ( attributes, elementId ) {
+				self.setAttributesForElement( self[ elementId ], attributes );
```
gbirke: If you change the constructor to accept a jQuery object, you can replace `self[ elementId ]` with something like 

    self.elements.find( elementId );

It'll make the unit test a bit more complicated, but you'll manage.

app/js/lib/view_handler/attribute_value_handler.js
```diff
@@ -0,0 +1,33 @@
...
+
+var objectAssign = require( 'object-assign' ),
+	_ = require( 'underscore' ),
+
+	AttributeValueHandler = {
+		'post-code': null,
+		city: null,
+		email: null,
+
+		update: function ( elements ) {
```
gbirke: I found this parameter name confusing because you aren't passing in actual elements but attribute data for elements. Maybe rename it to `elementAttributeData`?

app/js/lib/actions.js
```diff
@@ -114,6 +114,15 @@ module.exports = {
 			type: 'FINISH_SEPA_CONFIRMATION_VALIDATION',
 			payload: validationResult
 		};
+	},
+
+	newCountrySpecificsUpdateAction: function ( countryCode ) {
```
gbirke: Is this new action really needed? Couldn't the reducer just listen to the `CHANGE_CONTENT` action of the country field?

app/templates/MembershipApplication.html.twig
```diff
@@ -52,7 +52,8 @@
 				WMDE.Components.createValidatingTextComponent( store, $( '#street' ), 'street' ),
 				WMDE.Components.createValidatingTextComponent( store, $( '#post-code' ), 'postcode' ),
 				WMDE.Components.createValidatingTextComponent( store, $( '#city' ), 'city' ),
-				WMDE.Components.createRadioComponent( store, $( '#country' ), 'country' ),
+				WMDE.Components.createSelectMenuComponent( store, $( '#country' ), 'country' ),
+				WMDE.Components.createCountrySpecificsUpdateTriggerComponent( store, $( '#country' ), 'country' ),
```
gbirke: Add multiple "components" for the same element is not how I imagined components to work.

JeroenDeDauw:
Follow up at https://github.com/wmde/FundraisingFrontend/pull/483

PR #480, opened 2016-06-08
Add preliminary state graph

gbirke:
Added state graph as Graphviz file.

The description of states is not complete and also needs the
additional actions when moderated/deleted donations are booked
externally and the guard clauses when a state is restored in the backend (see
https://github.com/wmde/fundraising-backend/blob/bf379296aef2a5ee702aacdc4c91261db6bfc065/application/model/Donation.php#L50-L73
for that).

### DIFF
--- a/doc/donation_state.dot
+++ b/doc/donation_state.dot
@@ -1,0 +1,15 @@
+digraph G {
+  "New" -> "Moderation" [label="Add comment with bad words"]
+  "New" -> "Cancelled" [label="Cancelled by user or in backend"]
+  "Promise" -> "Moderation" [label="Add comment with bad words"]
+  "External_Incomplete"  -> "Moderation" [label="Add comment with bad words"]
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
+1. How to best view without installing fancy stuff?

gbirke:
http://www.webgraphviz.com/

JeroenDeDauw:
Cool.

External Incomplete is a possible starting state right? Perhaps the starting states should be indicated somehow? Cause now it kinda looks like you cannot get to any of the external states without first going through moderation.

gbirke:
Added double lines to mark initial states

JeroenDeDauw:
Merged it now. Probably good if @KaiNissen does a double check though.

KaiNissen:
It is actually the result of @gbirke and me talking about this. I do approve.

PR #482, opened 2016-06-08
Allow external booking of canelled donations and those in moderation

JeroenDeDauw:
Part of https://phabricator.wikimedia.org/T137328


gbirke:
Can you add the deleted status as well? I'd rather not have that in two PRs.

JeroenDeDauw:
Sure. Ill poke at it now

PR #487, opened 2016-06-10
Make comment private when changing status from moderation or cancelled

JeroenDeDauw:
Finishes https://phabricator.wikimedia.org/T137328

Please verify the logic is correct. What I did is check if the status is either
moderation or cancelled, as opposed to checking if the status is not external
incomplete as described in the ticket. I suspect these are equivalent, though
am not sure as the later is not very explicit.

### DIFF
--- a/src/Domain/Model/Donation.php
+++ b/src/Domain/Model/Donation.php
@@ -103,7 +103,7 @@ class Donation {
 	}
 
 	public function addComment( DonationComment $comment ) {
-		if ( $this->comment !== null ) {
+		if ( $this->hasComment() ) {
 			throw new RuntimeException( 'Can only add a single comment to a donation' );
 		}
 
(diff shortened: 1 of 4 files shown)


src/Domain/Model/Donation.php
```diff
@@ -149,9 +149,25 @@ public function confirmBooked() {
...
+			$this->makeCommentPrivate();
+		}
+
 		$this->status = self::STATUS_EXTERNAL_BOOKED;
 	}
 
+	private function makeCommentPrivate() {
+		$this->comment = new DonationComment(
+			$this->comment->getCommentText(),
+			false,
```
KaiNissen: Should we use constants here?

JeroenDeDauw: Maybe. The context here is very small and specific though. Anyway, feel free to make PR.

Another possibility for this code is to move it into a `createPrivateCopy` method in `DonationComment`.

KaiNissen: Right, actually the method name says it all.

PR #490, opened 2016-06-10
 	Made evil comments private and their donations marked for moderation

JeroenDeDauw:
Finishes https://phabricator.wikimedia.org/T137307

### DIFF
--- a/src/UseCases/AddComment/AddCommentUseCase.php
+++ b/src/UseCases/AddComment/AddCommentUseCase.php
@@ -9,6 +9,7 @@ use WMDE\Fundraising\Frontend\Domain\Repositories\DonationRepository;
 use WMDE\Fundraising\Frontend\Domain\Repositories\GetDonationException;
 use WMDE\Fundraising\Frontend\Domain\Repositories\StoreDonationException;
 use WMDE\Fundraising\Frontend\Infrastructure\DonationAuthorizer;
+use WMDE\Fundraising\Frontend\Validation\TextPolicyValidator;
 
 /**
  * @license GNU GPL v2+
@@ -18,10 +19,14 @@ class AddCommentUseCase {
 
 	private $donationRepository;
 	private $authorizationService;
+	private $textPolicyValidator;
 
-	public function __construct( DonationRepository $repository, DonationAuthorizer $authorizationService ) {
+	public function __construct( DonationRepository $repository, DonationAuthorizer $authorizationService,
+		TextPolicyValidator $textPolicyValidator ) {
+
 		$this->donationRepository = $repository;
 		$this->authorizationService = $authorizationService;
+		$this->textPolicyValidator = $textPolicyValidator;
 	}
 
 	public function addComment( AddCommentRequest $addCommentRequest ): AddCommentResponse {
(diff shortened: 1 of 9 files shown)


src/UseCases/AddComment/AddCommentUseCase.php
```diff
@@ -46,6 +51,10 @@ public function addComment( AddCommentRequest $addCommentRequest ): AddCommentRe
 
 		$donation->addComment( $this->newCommentFromRequest( $addCommentRequest ) );
 
+		if ( !$this->commentTextPassesValidation( $addCommentRequest->getCommentText() ) ) {
```
KaiNissen: In cases where users want to publish their comment the validation is being carried out twice.

src/UseCases/AddComment/AddCommentUseCase.php
```diff
@@ -63,9 +72,18 @@ private function requestIsAllowed( AddCommentRequest $addCommentRequest ): bool
...
 			$request->getCommentText(),
-			$request->isPublic(),
+			$this->commentCanBePublic( $request ),
 			$request->getAuthorDisplayName()
 		);
 	}
 
+	private function commentCanBePublic( AddCommentRequest $request ): bool {
+		return $request->isPublic()
+			&& $this->commentTextPassesValidation( $request->getCommentText() );
```
KaiNissen: Again, this removes the original intent of users whether or the comment should be public. What about we add another filter for moderated (and deleted/cancelled) donations in the `CommentRepository` to prevent that stuff from ending up in the public list of comments?

src/UseCases/AddComment/AddCommentUseCase.php
JeroenDeDauw: I know. I did not see a way to do it only once and still have both checks happen, without adding mutable state. My thinking here is that the code is slightly simpler without the state, and that the performance difference is not relevant (no expensive resources are used). If you disagree with the later, or see a way to do it once without making the code more complicated, I'd love to hear it.

src/UseCases/AddComment/AddCommentUseCase.php
JeroenDeDauw: Works for me. I suggest to create a ticket for that under New Stuff. One question I have for it is: does the backend need to be modified in any way to match such a change?

src/UseCases/AddComment/AddCommentUseCase.php
KaiNissen: We could store the validation result in a private field and that if the validation was already done. True, the use case object wouldn't be immutable anymore (at least not internally).

I'm just saying, because the validator also checks DNS entries for URLs included in a comment, which would increase the server response time if executed twice. Not really opposing this PR as is, though.

JeroenDeDauw: As far as I understood the code, the DNS entries are not checked. It's checking URLs, but not doing the DNS check.

gbirke:
Described how this can be solved in https://phabricator.wikimedia.org/T137704
@KaiNissen please check if the task description is correct.

src/UseCases/AddComment/AddCommentUseCase.php
KaiNissen: Right. The old application [did](https://github.com/wmde/fundraising/blob/master/src/CommentHandler.php#L42), though. As far as I remember, we introduced that a while back, because there were false positives produced by punctuation mistakes. What looks like a URL doesn't necessarily have to be [one.is](one.is) what I mean.

PR #497, opened 2016-06-13
Added page parameter to list comments JSON route

JeroenDeDauw:
Fixes https://phabricator.wikimedia.org/T126202


JeroenDeDauw:
Note that the task is not done. I am apparently blind, and did this for the JSON route instead of the HTML one.

PR #499, opened 2016-06-13
Add access denied page

gbirke:
Add an application-level "Access Denied" exception.
Add a special error handler that catches it.
Use the exception for donation confirmation.


gbirke:
Not sure if the Exception belongs in the application layer, please comment on that.

PR #500, opened 2016-06-16
Different sucess messages when adding comment

gbirke:
Depending on moderation status we need different sucesss messages.

Made the messages translatable.

### DIFF
--- a/src/UseCases/AddComment/AddCommentResponse.php
+++ b/src/UseCases/AddComment/AddCommentResponse.php
@@ -7,11 +7,12 @@ namespace WMDE\Fundraising\Frontend\UseCases\AddComment;
 /**
  * @license GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
  */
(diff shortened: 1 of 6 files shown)


gbirke:
Not sure about the test and the translation keys as strings in the use case. Should they be constants in the response class? the constant names would just be uppercase versions of the translation keys. But it would improve the tests if we could check individual constants.

src/UseCases/AddComment/AddCommentResponse.php
```diff
@@ -7,21 +7,24 @@
...
 
 	public static function newFailureResponse( string $errorMessage ): self {
 		return new self( $errorMessage );
 	}
 
 	private $errorMessage;
+	private $successMessage;
 
-	private function __construct( string $errorMessage = '' ) {
+	private function __construct( string $errorMessage = '', string $successMessage = '' ) {
```
JeroenDeDauw: Now there are 3 empty strings while we only need two. I suggest making the constructor parameters required.

JeroenDeDauw:
The variable and method names are deceptive like this, as they do not deal with messages, but message keys.

PR #503, opened 2016-06-17
Credit card payments

KaiNissen:


app/config/config.test.json
```diff
@@ -40,7 +40,8 @@
 		"project-id": "myprojectid",
 		"background-color": "000000",
 		"skin": "default",
-		"theme": "default"
+		"theme": "default",
+		"testmode": true
```
gbirke: This should also be added to `config.dist.json` which is meant as a kind of "template" for the real production files. if It's missing, there will be fatal type errors in `Presentation/CreditCardConfig.php`

JeroenDeDauw:
+1

JeroenDeDauw:
I'm not fond of the method name `isTestMode`. It's not valid English. `isInTestMode` makes more sense on its own. Though taken together with the class it perhaps does not, as it's not the config that is in test mode.

Took me looking at the code for a bit to realize `CreditCardConfig` is just a wrapper around config for `CreditCardUrlGenerator`. `CreditCardUrlConfig` would avoid such confusion.

tests/EdgeToEdge/Routes/CreditCardPaymentNotificationRouteTest.php
```diff
@@ -61,13 +62,17 @@ public function testGivenValidRequest_applicationIndicatesSuccess() {
...
 				'/handle-creditcard-payment-notification',
 				$this->newRequest()
 			);
 
 			$this->assertSame( 200, $client->getResponse()->getStatusCode() );
-			$this->assertSame( 'successful', $client->getResponse()->getContent() );
+			$this->assertContains( 'status=ok', $client->getResponse()->getContent() );
+			$this->assertContains(
+				'url=http://my.donation.app/show-donation-confirmation?donationId=1&accessToken=my_secret_access_token',
+				$client->getResponse()->getContent()
```
JeroenDeDauw: Not sure, but is it nice to bind to the argument order in the URL here? Would doing multiple asserts for the ket/value pairs not be better?

JeroenDeDauw:
Merging but please beware of comments

PR #505, opened 2016-06-17
Refactor donation validation

gbirke:
Validate request instead of domain object.

~~TODO: Low amount test for validator.~~
~~TODO: Move moderation checks to another class~~
~~TODO: Change amount validator to use string instead of Euro class.~~
~~TODO: Make other tests pass~~


JeroenDeDauw:
> TODO: Change amount validator to use string instead of Euro class.

mhhh... why? Passing around money strings and various code interpreting them is kinda scary.

JeroenDeDauw:
My thinking here was that while the thing validating the request would need to deal with a string, you can still have a validator for an amount, which does not also have the responsibility of determine what it gets is actually an amount.

Muh. Forgot to ask Sebastian about having a string in the request model and associated validator to begin with. This is presentation specific to some degree. Imagine you get a JSON thing, and do json_decode, you'll presumably have a numeric type.

Actually, what about the localization? That is definitely a presentation thing. And for euro amounts, we'd already end up having to deal with invalid values there right? If so, then the request model could just use the Euro object, avoiding the earlier mentioned problems.

tests/Integration/UseCases/AddDonation/AddDonationValidatorTest.php
```diff
@@ -0,0 +1,145 @@
...
+	public function setUp() {
+		$this->donationValidator = $this->newDonationValidator();
+	}
+
+	public function testGivenValidDonation_validationIsSuccessful() {
+		$request = ValidAddDonationRequest::getRequest();
+		$this->assertEmpty( $this->donationValidator->validate( $request )->getViolations() );
+	}
+
+	public function testGivenAnonymousDonor_validatorReturnsTrue() {
```
KaiNissen: Actually, it returns an empty array.

tests/Integration/UseCases/AddDonation/AddDonationValidatorTest.php
```diff
@@ -0,0 +1,145 @@
...
+
+		$this->assertConstraintWasViolated(
+			$this->donationValidator->validate( $request ),
+			AddDonationValidationResult::SOURCE_PAYMENT_TYPE
+		);
+	}
+
+	public function testGivenUnsupportedPaymentType_validatorReturnsFalse() {
+		$request = ValidAddDonationRequest::getRequest();
+		$request->setPaymentType( 'KaiCoin' );
```
KaiNissen: such easter egg

tests/Fixtures/SucceedingStringLengthValidator.php
```diff
@@ -0,0 +1,22 @@
...
+namespace WMDE\Fundraising\Frontend\Tests\Fixtures;
+
+use WMDE\Fundraising\Frontend\Validation\StringLengthValidator;
+use WMDE\Fundraising\Frontend\Validation\ValidationResult;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class SucceedingStringLengthValidator extends StringLengthValidator {
```
KaiNissen: This is never used. I assume, this slipped into the commit while you were already working on another task?

src/UseCases/AddDonation/AddDonationPolicyValidator.php
```diff
@@ -0,0 +1,70 @@
...
+		if ( $request->donorIsAnonymous() ) {
+			return [];
+		}
+
+		return array_merge(
+			$this->getPolicyViolationsForField( $request->getDonorFirstName(), Result::SOURCE_DONOR_FIRST_NAME ),
+			$this->getPolicyViolationsForField( $request->getDonorLastName(), Result::SOURCE_DONOR_LAST_NAME ),
+			$this->getPolicyViolationsForField( $request->getDonorCompany(), Result::SOURCE_DONOR_COMPANY ),
+			$this->getPolicyViolationsForField( $request->getDonorStreetAddress(), Result::SOURCE_DONOR_STREET_ADDRESS ),
+			$this->getPolicyViolationsForField( $request->getDonorPostalCode(), Result::SOURCE_DONOR_POSTAL_CODE ),
```
KaiNissen: Just a remark/question, out of scope of this PR: On the client side we restrict input to numeric strings. Do you think, we should do this in the server side validation as well? This would make policy checks on this field obsolete.

tests/Fixtures/SucceedingStringLengthValidator.php
gbirke: Yes. I started working on [field length validation](https://phabricator.wikimedia.org/T136604) when I realized that I need to refactor first.

tests/Integration/UseCases/AddDonation/AddDonationValidatorTest.php
gbirke: Taken from `DonationValidator`, written and hidden by you ;-)

KaiNissen: No, it was Jeroen, of course!

src/UseCases/AddDonation/AddDonationPolicyValidator.php
gbirke: You're right. I've restricted the postal code to numeric input and removed the policy check.

PR #510, opened 2016-06-20
Moved token generation to doctrine event listner thing

JeroenDeDauw:


src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -76,52 +71,16 @@ public function addDonation( AddDonationRequest $donationRequest ): AddDonationR
...
-		$updateToken = $this->tokenGenerator->generateToken();
-
-		$this->authorizationUpdater->allowModificationViaToken(
-			$donationId,
-			$updateToken,
-			$this->tokenGenerator->generateTokenExpiry()
+		return AddDonationResponse::newSuccessResponse(
+			$donation,
+			$tokens->getUpdateToken(),
+			$tokens->getAccessToken()
```
KaiNissen: We could pass the instance of `DonationTokens` instead of two strings here.

JeroenDeDauw:
Meh. So now the CI breaks whenever we do a composer update apparently... @gbirke how did you fix this when it happened last time? @KaiNissen just ignore for this PR please. Tests run locally, and there is no indication on TravisCI anything from this PR is wrong either.

src/UseCases/AddDonation/AddDonationUseCase.php
JeroenDeDauw: Yeah. I consciously did not include that in this PR though.

KaiNissen:
Yeah, already tried the tests locally.

PR #520, opened 2016-06-24
fix cancellation of membership applications

KaiNissen:
fixes [phab:T138567](https://phabricator.wikimedia.org/T138567)

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -417,15 +417,15 @@ $app->post(
 	}
 );
 
-$app->post(
+$app->get(
 	'cancel-membership-application',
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Is there not some security reason to have this work via post? I have the notion write actions should generally be post, though am not sure where it is coming from

KaiNissen:
I wasn't quite satisfied with the current situation of a form sitting in the sidebar of the confirmation page and pretending to be a link. Although I consider your comment that write actions should be POST, but I think using the updateToken should suffice.

PR #528, opened 2016-06-28
Trim values

KaiNissen:
Solves [phab:T138086](https://phabricator.wikimedia.org/T138086)

### DIFF
--- a/tests/EdgeToEdge/TrimValuesTest.php
+++ b/tests/EdgeToEdge/TrimValuesTest.php
@@ -1,0 +1,51 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\EdgeToEdge;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
+2 to PHP, +1 to JS

tests/EdgeToEdge/TrimValuesTest.php
```diff
@@ -0,0 +1,51 @@
...
+			'var2' => 'val2 ',
+			'var3' => ' val3 ',
+			'var4' => '   val4  ',
+			'var5' => 0,
+			'var6' => 1234.56,
+		];
+	}
+
+	private function trimArray( array $params ) {
+		array_walk( $params, [ self::class, 'trimValue' ] );
```
gbirke: Not sure if this is too much "logic" in the code and there should be a `trimmedParams` function instead, that returns a hand-trimmed version of the array from `newParams`. But if @JeroenDeDauw gives +2 to PHP, I don't object.

JeroenDeDauw: I was also thinking about this and would have gone with hand trimming them. The main value of this test seems to be verifying that the trimming is actually happening, not that the trim function used is correct.

PR #529, opened 2016-06-28
Added pagination to comment list html template

JeroenDeDauw:
I touched HTML! Brb, going to wash my hands and eyes with bleach

https://phabricator.wikimedia.org/T126202


gbirke:
See #534 for slight improvements, otherwise +2

PR #536, opened 2016-06-29
Form submission limit

KaiNissen:


tests/templates/System_Message.twig
```diff
@@ -0,0 +1 @@
+{$ message | trans() | raw $}
```
gbirke: I know the old application just displayed the message, but shouldn't we be a bit more user friendly and display it in the site layout?

KaiNissen: This is a stubbed template used in the route tests. There is a [nicer version](https://backend.wikimedia.de/wiki/index.php?title=Web:Spendenseite-HK2013/rewrite/System_Message) in the wiki.

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
```diff
@@ -178,4 +180,50 @@ public function testGivenValidRequest_requestIsRedirected() {
 		} );
 	}
 
+	public function testWhenApplicationGetsPersisted_timestampIsStoredInCookie() {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ) {
+			$factory->setNullMessenger();
+			$factory->setCookieHandler( new FixedCookieHandler( [ 'memapp_timestamp' => self::FIXED_TIMESTAMP ] ) );
```
gbirke: Maybe you could use a `CookieHandlerSpy` instead.

tests/Fixtures/FixedCookieHandler.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace WMDE\Fundraising\Frontend\Tests\Fixtures;
+
+use WMDE\Fundraising\Frontend\Infrastructure\CookieHandler;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class FixedCookieHandler implements CookieHandler {
```
gbirke: This class confuses me a bit because from the name I expected it to always return the same value for the cookies, some kind of "immutable cookie jar". Instead, it ignores the values that are set and instead always returns the fixed values. If you further use this class (see my other comments on usage), could you please mention how this class works in a doc comment?

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
```diff
@@ -178,4 +180,50 @@ public function testGivenValidRequest_requestIsRedirected() {
...
+		}
+
+		return false;
+	}
+
+	public function testWhenMultipleMembershipFormSubmissions_requestGetsRejected() {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ) {
+			$factory->setNullMessenger();
+
+			$cookieHandler = new FixedCookieHandler( [ 'memapp_timestamp' => self::FIXED_TIMESTAMP ] );
```
gbirke: Would't it be better to use a timestamp 10 seconds ago? Otherwise the tests will break in 2020.

app/bootstrap.php
```diff
@@ -33,6 +38,21 @@ function ( Request $request, Application $app ) {
...
 
+$app->before( function( Request $request ) use ( $ffFactory ) {
+	if ( $ffFactory->getCookieHandler() === null ) {
+		$ffFactory->setCookieHandler( new BrowserCookieHandler( $request->cookies ) );
+	}
+}, Application::EARLY_EVENT );
+
+$app->after(
+	function ( Request $request, Response $response ) use ( $ffFactory ) {
+		$cookieHandler = $ffFactory->getCookieHandler();
```
gbirke: I struggled with the `after` function a bit, maybe you can add the comment that explains what is does, e.g. "Store Cookie values that were set in the route handlers"

tests/Fixtures/FixedCookieHandler.php
KaiNissen: Now the tests are using a mock instead.

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
KaiNissen: Right. When using a spy the test cannot check the actual timestamp stored, but this is covered in the other test anyway.

JeroenDeDauw:
Needs rebase

KaiNissen:
rebased

src/Factories/FunFunFactory.php
```diff
@@ -1099,4 +1101,28 @@ private function newDonationAmountPolicyValidator(): AmountPolicyValidator {
...
+	 */
+	public function getCookieHandler() {
+		return $this->cookieHandler;
+	}
+
+	public function setCookieHandler( CookieHandler $cookieHandler ) {
+		$this->cookieHandler = $cookieHandler;
+	}
+
+	public function getDonationTimeframeLimit() {
```
JeroenDeDauw: type string?

src/Factories/FunFunFactory.php
```diff
@@ -1099,4 +1101,28 @@ private function newDonationAmountPolicyValidator(): AmountPolicyValidator {
...
+		return $this->config['donation-timeframe-limit'];
+	}
+
+	public function newSystemMessageResponse( string $message ) {
+		$test = $this->getIncludeTemplate( 'System_Message.twig' );
+		return $test->render( [ 'message' => $message ] );
+	}
+
+	public function getMembershipApplicationTimeframeLimit() {
```
JeroenDeDauw: return type

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
```diff
@@ -178,4 +181,52 @@ public function testGivenValidRequest_requestIsRedirected() {
...
+				'POST',
+				'/apply-for-membership',
+				$this->newValidHttpParameters()
+			);
+
+			$this->assertContains( 'Sie haben vor sehr kurzer Zeit bereits', $client->getResponse()->getContent() );
+		} );
+	}
+
+	private function getPastTimestamp() {
```
JeroenDeDauw: return type

JeroenDeDauw:
Putting the cookies in the factory is a bit iffy

PR #537, opened 2016-06-30
do not trim non-string values

KaiNissen:


JeroenDeDauw:
+1. Dunno if that is how to check for a string in JS

KaiNissen:
It is, but now that you say that, I think we could also use underscore's `_.isString()`.

PR #550, opened 2016-07-05
fix timeframe limit

KaiNissen:
- actually use donation timeframe limit for donations
- actually create a past timestamp in tests


app/RouteHandlers/AddDonationHandler.php
```diff
@@ -178,7 +178,7 @@ private function isSubmissionAllowed( Request $request ) {
 		}
 
 		$minNextTimestamp = \DateTime::createFromFormat( self::TIMESTAMP_FORMAT, $lastSubmission )
-			->add( new \DateInterval( $this->ffFactory->getMembershipApplicationTimeframeLimit() ) );
+			->add( new \DateInterval( $this->ffFactory->getDonationTimeframeLimit() ) );
```
JeroenDeDauw: oops? :D

PR #554, opened 2016-07-06
Refactording of bad/white word retrieval

JeroenDeDauw:
https://phabricator.wikimedia.org/T138964

This allows us to add caching in a sane way. Will do that in a separate PR. Thoughts on where to cache? We approach is in the data access layer, which would be `ApiBasedPageRetriever`. Or perhaps at the other end, and just cache the word lists. The later approach will not have caching for other users of `ApiBasedPageRetriever`, though perhaps that is not desired. (The only user now is the twig templating, which already has caching.)

### DIFF
--- a/src/Validation/TextPolicyValidator.php
+++ b/src/Validation/TextPolicyValidator.php
@@ -4,14 +4,17 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\Validation;
 
+use WMDE\Fundraising\Frontend\Infrastructure\ArrayBasedStringList;
+use WMDE\Fundraising\Frontend\Infrastructure\StringList;
+
 /**
  * @licence GNU GPL v2+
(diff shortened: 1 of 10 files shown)


src/Validation/TextPolicyValidator.php
```diff
@@ -21,6 +24,25 @@ class TextPolicyValidator {
 	// Could use an URL validation strategy
 	const CHECK_URLS_DNS = 2;
 
+	public function __construct( StringList $badWords = null, StringList $whiteWords = null ) {
```
JeroenDeDauw: Could have used `Iterator` as well... that is more fuzzy, though has a lot of basic implementations already, avoiding the need for things such as `ArrayBasedStringList`. #SighPHP

gbirke:
I've investigated the template caching a saw that the cache is needed somewhere else. See this: https://phabricator.wikimedia.org/T139468

PR #555, opened 2016-07-07
Added CachingPageRetriever

JeroenDeDauw:
Part of https://phabricator.wikimedia.org/T138964
and https://phabricator.wikimedia.org/T139468

### DIFF
--- a/src/Infrastructure/CachingPageRetriever.php
+++ b/src/Infrastructure/CachingPageRetriever.php
@@ -1,0 +1,35 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+use Doctrine\Common\Cache\Cache;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class CachingPageRetriever implements PageRetriever {
+
+	private $pageRetriever;
+	private $cache;
+
+	public function __construct( PageRetriever $pageRetriever, Cache $cache ) {
+		$this->pageRetriever = $pageRetriever;
+		$this->cache = $cache;
+	}
+
+	public function fetchPage( string $pageName, string $fetchMode ): string {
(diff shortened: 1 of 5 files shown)


src/Infrastructure/CachingPageRetriever.php
```diff
@@ -0,0 +1,35 @@
...
+	private $cache;
+
+	public function __construct( PageRetriever $pageRetriever, Cache $cache ) {
+		$this->pageRetriever = $pageRetriever;
+		$this->cache = $cache;
+	}
+
+	public function fetchPage( string $pageName, string $fetchMode ): string {
+		if ( $this->cache->contains( $pageName ) ) {
+			return $this->cache->fetch( $pageName );
```
JeroenDeDauw: Note that I am ignoring $fetchMode on purpose. See the comment in the interface. I am not sure we still need this thing, so did not spend the time on supporting it here for now.

gbirke: You mean you don't have a default value for `$fetchMode`? That's fine, as far as I remember nearly all calls to it have a fetch mode.

JeroenDeDauw: I mean that this decorator does not hold  it into account as well. So if you fetch page A with mode FOO, and then with mode BAR, you will get the mode FOO version as it got cached.

KaiNissen: That should also be fine. Whether content should be fetched in raw mode depends on the content (e. g. plain text mails, such word lists), not the case of use.

JeroenDeDauw: Sure... though this limitation should be documented in case we keep $fetchMode. Can you explain why it is needed? One example of each case will do.

KaiNissen: The mode `render` is mostly used for retrieving displayable content. When using plain text mails or word lists, we don't want MediaWiki's parser to mess up the content. Word lists and also e. g. configuration files in JSON would break, the content of text mails would be "enriched" by HTML tags.

PR #556, opened 2016-07-07
Avoid infinite loops on server error

gbirke:
The deferred object returned by a jQuery AJAX call is not a proper
Promise which lead to endless loops on failure because the first
parameter of the rject method was a promise instead of a failure
message.

This solves https://phabricator.wikimedia.org/T139472

### DIFF
--- a/app/js/lib/form_validation.js
+++ b/app/js/lib/form_validation.js
@@ -2,6 +2,7 @@
 
 var jQuery = require( 'jquery' ),
 	objectAssign = require( 'object-assign' ),
+	Promise = require( 'promise' ),
 	_ = require( 'underscore' ),
 
 	ValidationStates = {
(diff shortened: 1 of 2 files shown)


app/js/lib/form_validation.js
```diff
@@ -2,8 +2,23 @@
...
+	 * @param jQueryDeferredObject
+	 * @returns {Promise}
+	 */
+	jQueryDeferredToPromise = function ( jQueryDeferredObject ) {
+		return new Promise( function ( resolve, reject ) {
+			jQueryDeferredObject.then( resolve, function ( xhrObject, statusCode, statusMessage ) {
+				reject( statusMessage );
+			} );
+		} );
+	},
```
JeroenDeDauw: matrix

JeroenDeDauw:
Needs rebase

PR #558, opened 2016-07-08
client-side form validation

KaiNissen:
This is a slight refactoring of the validation stuff happening on the client side.

I'm not so sure about all the logic that happens in the form setup script (`donationForm.js`). This seems necessary to reflect the dependency of required fields for different address types. I didn't want to combine the validation and the form content stuff, though. @gbirke, please have a look at that, maybe you can come up with a suggestion.

The renderer of the field components use `val()` to update the field's value. Since programmatic changes (like emptying the field's value) do not trigger a change event, which the input validation would handle to set the validity indicator, I introduced a trigger for that event in the `render()` method, which also does not seem very nice. @gbirke, please also have a closer look at that.

This actually fixes the tasks [phab:T139048](https://phabricator.wikimedia.org/T139048) and [phab:T139057](https://phabricator.wikimedia.org/T139057)


gbirke:
In general your approach is fine, I'd give it +2 when it's finished. I made some improvements in #559 that hint where this PR should go IMHO.

Regarding the setup for address-type specific validation: I refactored the AddressValidator a bit, but I di'd leave it like that. I think the duplication of field names in `triggerFieldValidity` and  `AddressValidator` is better than adding another layer of nesting to `DefaultRequiredFieldsForAddressType`, like this:

      DefaultRequiredFieldsForAddressType = {
		person: {
                     required: [ 'salutation', 'firstName', 'lastName', 'street', 'postcode', 'city', 'email' ],
                     neutral: [ 'companyName' ]
                },
            // ... etc
	},
Maybe @JeroenDeDauw has a clever idea.

app/js/lib/form_components.js
```diff
@@ -42,7 +42,10 @@ var objectAssign = require( 'object-assign' ),
 		onChange: null,
 		validator: null,
 		render: function ( formContent ) {
-			this.element.val( formContent[ this.contentName ] );
+			if ( this.element.val() !== formContent[ this.contentName ] ) {
+				this.element.val( formContent[ this.contentName ] );
+				this.element.change();
+			}
```
gbirke: While I agree that triggering a DOM event during rendering feels somewhat icky, it's the correct behavior none the less. Well done.

app/js/lib/reducers/input_validation.js
```diff
@@ -22,6 +22,19 @@ function inputValidation( validationState, action ) {
 				isValid: inputIsValid( action.payload.value, action.payload.pattern )
 			};
 			return newValidationState;
+		case 'VALIDATE_FIELDSET':
```
gbirke: I'd rather call this `MARK_EMPTY_FIELD_AS_INVALID` or `MARK_UNCHANGED_FIELDS_AS_INVALID` or something, because no real validation is taking place here.

gbirke:
What's left after merging #559 is to implement updating the validation messages for the 1st and 3rd page, as outlined in my TODOs. And testing if the first page is sent invalid data via POST (from a fake "banner"), it is displaying the correct error message (because in this case I think displaying the error message when clicking ion the button is too late).

All in all you're doing a great job managing this mad mix of background validation triggered by field changes and serverless client-side validation triggered by submit buttons. I'm starting to think that implementing the spinning indicator from https://phabricator.wikimedia.org/T134166 would make things much more easy, but that refactoring is "new stuff" unfortunately.

PR #562, opened 2016-07-11
Cache requests to the CMS wiki on the filesystem

JeroenDeDauw:
Fixes https://phabricator.wikimedia.org/T138964
and https://phabricator.wikimedia.org/T139468

Tested manually. Not sure how to write sane automated tests for this.

The cache is only used when the app gets initialized via the non-dev
entry point. Currently the twig cache is used by the dev-entry point
unless you disable it in your config (which is also used by the non-
dev one). We presumably want to also change that.

Twig cache was moved from var/cache to var/cache/twig.

Profiler cache was moved from /app/cache/profiler to var/cache/profiler.
Might want to change this to var/profiler. Not really a cache is it?

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -4,6 +4,9 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\Factories;
 
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\FilesystemCache;
+use Doctrine\Common\Cache\VoidCache;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\ORM\EntityManager;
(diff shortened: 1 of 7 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -81,7 +81,7 @@
 use WMDE\Fundraising\Frontend\Infrastructure\Repositories\LoggingSubscriptionRepository;
 use WMDE\Fundraising\Frontend\Infrastructure\TemplateBasedMailer;
 use WMDE\Fundraising\Frontend\Infrastructure\TokenGenerator;
-use WMDE\Fundraising\Frontend\Infrastructure\TwigCachePurger;
+use WMDE\Fundraising\Frontend\Infrastructure\AllOfTheCachePurger;
```
KaiNissen: Funny name, but, err, maybe something like `EntireCachePurger` or `CompleteCachePurger`? Not so sure, it sounds like the purger itself is entire or complete. Merging it anyway.

JeroenDeDauw: Yeah, I also can't think of something good that is not open to misinterpretation

PR #563, opened 2016-07-11
Catch mailer exceptions in cancel donation UC

JeroenDeDauw:
This is the non-presentation part of https://phabricator.wikimedia.org/T133161

### DIFF
--- a/src/UseCases/CancelDonation/CancelDonationUseCase.php
+++ b/src/UseCases/CancelDonation/CancelDonationUseCase.php
@@ -59,17 +59,31 @@ class CancelDonationUseCase {
 
 		$this->donationLogger->log( $donation->getId(), self::LOG_MESSAGE_FOR_BACKEND );
 
-		$this->sendConfirmationEmail( $donation );
+		try {
+			$this->sendConfirmationEmail( $donation );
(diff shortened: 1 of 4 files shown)


src/UseCases/CancelDonation/CancelDonationUseCase.php
```diff
@@ -59,17 +59,31 @@ public function cancelDonation( CancelDonationRequest $cancellationRequest ): Ca
 
 		$this->donationLogger->log( $donation->getId(), self::LOG_MESSAGE_FOR_BACKEND );
 
-		$this->sendConfirmationEmail( $donation );
+		try {
+			$this->sendConfirmationEmail( $donation );
+		}
+		catch ( \RuntimeException $ex ) {
+			return new CancelDonationResponse(
```
KaiNissen: We extracted creating the other response objects into a private method. Any specific reason why this is not the case here?

JeroenDeDauw: My reasoning was that there is just one caller. Same goes for the success case. So yeah, kinda inconsistent. Guess extracting it is a bit nicer.

PR #564, opened 2016-07-12
Cancellation mail failure

KaiNissen:


src/Presentation/Presenters/CancelDonationHtmlPresenter.php
```diff
@@ -22,7 +22,8 @@ public function __construct( TwigTemplate $template ) {
 	public function present( CancelDonationResponse $response ): string {
 		return $this->template->render( [
 			'donationId' => $response->getDonationId(),
-			'cancellationSuccessful' => $response->cancellationSucceeded()
+			'cancellationSuccessful' => $response->cancellationSucceeded(),
+			'mailDeliveryFailed' => !$response->mailDeliveryFailed()
```
JeroenDeDauw: Should that not be `succeeded` then?

JeroenDeDauw:
I take it we have a template on CMS wiki that uses this?

src/Presentation/Presenters/CancelDonationHtmlPresenter.php
KaiNissen: Dang, it was `mailDeliverySuccessful` first, then I last-minute-equalized, but forgot to remove the negation. Good, that this line of code is not tested. Fixed that, added a route test.

KaiNissen:
Yes, there is a [template in the wiki](https://backend.wikimedia.de/wiki/index.php?title=Web:Spendenseite-HK2013/rewrite/Donation_Cancellation_Confirmation). I removed the one in the repo.

PR #566, opened 2016-07-13
Actually save changes after donation cancellation

JeroenDeDauw:
Fixes https://phabricator.wikimedia.org/T140126

Includes refactoring of the test class: 660ms => 170ms runtime :)

Also added missing database failure tests

And had to fix a lot of tests that failed after changing the FakeDonationRepository,
as they where relying on being able to modify already stored objects via reference.


JeroenDeDauw:
I'll add tests for the repository failure cases, which appears to be missing, in a follow up

PR #573, opened 2016-07-19
[DNM] Trying out code sharing approach for application modification

JeroenDeDauw:
Based on https://github.com/wmde/FundraisingStore/pull/90

---

Goal: avoid boilerplate code in services that need to modify a membership application identified by ID, and simplify testing them.

### DIFF
--- a/src/DataAccess/DoctrineMembershipApplicationPiwikTracker.php
+++ b/src/DataAccess/DoctrineMembershipApplicationPiwikTracker.php
@@ -4,11 +4,11 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\DataAccess;
 
-use Doctrine\ORM\EntityManager;
-use Doctrine\ORM\ORMException;
 use WMDE\Fundraising\Entities\MembershipApplication;
 use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationPiwikTracker;
 use WMDE\Fundraising\Frontend\Infrastructure\MembershipApplicationPiwikTrackingException;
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
@KaiNissen I added both the callable approach and the individual methods to the class in the store component. Opted to go with the callable here, as then testing the failure cases is real easy: just one test in which the modifyApplication throws an exception. When using the individual methods, you need to test an exception for each, and need to verify the second one actually gets called.

src/DataAccess/DoctrineMembershipApplicationPiwikTracker.php
```diff
@@ -4,55 +4,35 @@
...
-	private function persistApplication( MembershipApplication $application ) {
 		try {
-			$this->entityManager->persist( $application );
-			$this->entityManager->flush();
+			$this->applicationModifier->modifyApplication(
+				$applicationId,
+				function( MembershipApplication $application ) use ( $trackingString ) {
+					$application->setTracking( $trackingString );
+				}
+			);
```
JeroenDeDauw: Should also call `$this->applicationModifier->flush()` the way it is currently implemented

JeroenDeDauw:
Any thoughts on this?

JeroenDeDauw:
Ping @gbirke @KaiNissen @wiese <o>

timEulitz:
This PR is almost 2 years old, is there any action that could be taken in the near future / should we pursue this further?

gbirke:
Was implemented by https://github.com/wmde/fundraising-memberships/pull/5

PR #575, opened 2016-07-19
[DNM] The road to hell is paved with good intentions

JeroenDeDauw:

PR #576, opened 2016-07-19
Added profiling decorators

JeroenDeDauw:
Only for CommentFinder, PageRetriever and Guzzle Client so far.
The later two are presumably the most important ones to profile.

Note that the generic decorator building approach relies on PHPUnits
mocking API, which internally builds a string and evals it. So this
is presumably something we do not want to use in the production code
when it is not initialized by the dev entry point. (This PR makes
sure that does not happen.)

Follow up things to do:

* Decorate ALL the interesting things with profilers
* Do something about the global access in the factory
* Move the generic decorator building code elsewhere (own package?)
* Thoroughly unit-test the decorator building 

Can already be merged as far as I'm concerned though. Will take care of the follow-ups.

suchprofile

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
 		"symfony/translation": "^3.0",
 		"symfony/twig-bridge": "^3.0",
 		"twig/extensions": "^1.3",
-		"micropayment-de/service-client": "~1.19.0"
+		"micropayment-de/service-client": "~1.19.0",
+		"phpunit/phpunit": "~5.4"
 	},
 	"repositories": [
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
This is how it started out https://github.com/wmde/FundraisingFrontend/pull/575 :)

JeroenDeDauw:
@sebastianbergmann Should I take care to avoid you at conferences now for my own safety?

KaiNissen:
may contain traces of magic

PR #580, opened 2016-07-20
Moved logger construction out of factory

JeroenDeDauw:
Only behaviour change for now is that the tests will no longer use a logger.
We can set one up in TestEnvironment if we want to though.

Like this we can set up different logging for production and for the dev
entry point. Presumably we want those to write to different loggs. Also,
right now the non-dev entry point is logging every request, which we presumably
do not want to do.


JeroenDeDauw:
@KaiNissen Looks like every request should have all pages retrieved from the wiki logged already. `ApiBasedPageRetriever` has 

```php
$this->logger->debug( __METHOD__ . ': pageTitle', [ $pageTitle ] );
```

This info ought to be helpful for finding out why things are still fetched over the network when they should be changed already.

JeroenDeDauw:
Should also look at how we can have the logs show up in the profiler. That'd be way more convenient

logs

PR #582, opened 2016-07-21
Added fun services panel to profiler UI

JeroenDeDauw:
Based on and including https://github.com/wmde/FundraisingFrontend/pull/580


JeroenDeDauw:
Please beware that this has security implications: we are exposing parameter values, which might contain sensitive information such as passwords (ie logging into the wiki). So care needs to be taken to not have those displayed. AFAIK, no such information is currently recorded. Yet such errors can easily arise, so I'd be good to make the dev entry point not publicly accessible. That probably is true as well without this PR, as exception info, config info, etc, is also not something you want attackers to access.

PR #584, opened 2016-07-26
Fix payment summary

KaiNissen:
This fixes the payment summary section on the second donation form page by adding the payment type.


gbirke:
+2 with the changes from #587 and a rebase on a fix for https://phabricator.wikimedia.org/T141381

PR #585, opened 2016-07-26
add donation form presenter

KaiNissen:
This introduces a donation form presenter. Data that was already entered using a banner is being passed to it and made available to the template. Currently, it only processes payment type, amount and payment interval.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -24,8 +24,10 @@ use WMDE\Fundraising\Frontend\Domain\Model\Iban;
 use WMDE\Fundraising\Frontend\Domain\Model\PersonName;
 use WMDE\Fundraising\Frontend\Domain\Model\PhysicalAddress;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+use WMDE\Fundraising\Frontend\Infrastructure\AmountParser;
 use WMDE\Fundraising\Frontend\Infrastructure\CreditCardPaymentHandlerException;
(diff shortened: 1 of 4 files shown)


app/routes.php
```diff
@@ -455,11 +457,15 @@ function ( Application $app, Request $request ) use ( $ffFactory ) {
 	}
 );
 
-$app->get( '/', function ( Application $app ) {
-	return $app->handle(
-		Request::create( '/page/DonationForm', 'GET' ),
-		HttpKernelInterface::SUB_REQUEST
-	);
+$app->get( '/', function ( Application $app, Request $request ) use ( $ffFactory ) {
```
gbirke: Is this change really needed? Would't it be handled by the subrequest to add donation?

KaiNissen: The `page` route can't pass those values into the template. Also, a subrequest does not automatically pass on variables sent by `GET` or `POST`.

app/routes.php
```diff
@@ -455,11 +457,15 @@ function ( Application $app, Request $request ) use ( $ffFactory ) {
...
+	$params = $request->query;
+	$amount = Euro::newFromFloat( ( new AmountParser( 'de_DE' ) )->parseAsFloat( $params->get( 'betrag', '' ) ) );
+	$donationRequest = new AddDonationRequest();
```
JeroenDeDauw: Why use the request model for the AddDonation UC here?

PR #587, opened 2016-07-26
Fix the fixes

gbirke:


web/res/js/donationForm.js
```diff
@@ -174,10 +174,10 @@ $( function () {
...
 					},
 					{
-						'BEZ': 'Lastschrift',
-						'UEB': 'Überweisung',
-						'MCP': 'Kreditkarte',
-						'PPL': 'PayPal'
+						'BEZ': ' per Lastschrift',
+						'UEB': ' per Überweisung',
+						'MCP': ' per Kreditkarte',
+						'PPL': ' per PayPal'
```
KaiNissen: Sorry, I should have mentioned that: the "per" was [added to the template](https://backend.wikimedia.de/wiki/index.php?title=Web:Spendenseite-HK2013/rewrite/DonationForm_PersonalData&curid=1100&diff=7186&oldid=7181). I think it's better to have just the terms here.

PR #588, opened 2016-07-26
Fix IBAN conversion validation

gbirke:
Fixes https://phabricator.wikimedia.org/T141381


KaiNissen:
This reintroduces [phab:T140476](https://phabricator.wikimedia.org/T140476), which should have been fixed by #568, but apparently wasn't.

JeroenDeDauw:
I supposed Kai's comment has been addressed since the commit is more recent now... +1 as I don't see anything obviously wrong, though then I don't understand the code well enough to catch a lot of things.

If this is something that was wrong and then not fully fixed and reintroduced, having a test for it seems real nice. Is that difficult to do nicely in this case?

app/js/lib/form_validation.js
```diff
@@ -97,6 +100,10 @@ var jQuery = require( 'jquery' ),
 				};
 				validationUrl = this.validationUrlForNonSepa;
 			}
+			// avoid sending incomplete data to the server
+			if ( _.find( data, isEmptyString ) !== undefined ) {
```
JeroenDeDauw: It took me a good number of seconds to fully interpret what this is doing, promoting me to create https://github.com/wmde/FundraisingFrontend/pull/591/commits/f6c2b0298ee96b83292b59116ebca81efa523740. As it is here, you need to deal with not relevant low level details such as the behaviour of the _.find function.

As I found out there, there is no need for such a general check, or for the local variables in this method. https://github.com/wmde/FundraisingFrontend/pull/591/commits/646dcc922e0ff9b4f0c974a2c6868ed27c4c299b

PR: https://github.com/wmde/FundraisingFrontend/pull/591

gbirke:
I think for testing this behavior, you really need a browser test, that simulates a user clicking in the interface.

KaiNissen:
There is one tiny bit missing: If I switch from direct debit to any other payment method, the bank data fields are empty and receive a green validity indicator in all cases (were valid before, were invalid before, were already empty).

gbirke:
Addressed comment from @KaiNissen 

@JeroenDeDauw Could you merge this?

PR #589, opened 2016-07-28
Piwik Tracking

KaiNissen:
This PR adds the following behaviour to the application:

- The Piwik script snippet is included into all pages
- Predefined "goals" are tracked both on the donation form and the confirmation page
- Custom variables that represent the initially intended and actual payment type, amount and interval are sent to Piwik after creating a donation

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -14,6 +14,7 @@ use WMDE\Fundraising\Frontend\Domain\Model\Iban;
 use WMDE\Fundraising\Frontend\Domain\Model\PaymentType;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\Infrastructure\AmountParser;
+use WMDE\Fundraising\Frontend\Infrastructure\PiwikVariableCollector;
 use WMDE\Fundraising\Frontend\Presentation\SelectedConfirmationPage;
 use WMDE\Fundraising\Frontend\UseCases\AddDonation\AddDonationRequest;
 use WMDE\Fundraising\Frontend\UseCases\AddDonation\AddDonationResponse;
@@ -36,7 +37,7 @@ class AddDonationHandler {
 		$this->app = $app;
 	}
 
-	public function handle( Request $request ): Response {
+	public function handle( Request $request, array $sessionTrackingData ): Response {
 		if ( !$this->isSubmissionAllowed( $request ) ) {
 			return new Response( $this->ffFactory->newSystemMessageResponse( 'donation_rejected_limit' ) );
 		}
@@ -48,17 +49,18 @@ class AddDonationHandler {
 			return new Response( $this->ffFactory->newDonationFormViolationPresenter()->present( $responseModel->getValidationErrors(), $addDonationRequest ) );
 		}
 
-		return $this->newHttpResponse( $responseModel, $this->ffFactory->getDonationConfirmationPageSelector()->selectPage() );
+		return $this->newHttpResponse( $responseModel, $this->ffFactory->getDonationConfirmationPageSelector()->selectPage(), $sessionTrackingData );
 	}
 
-	private function newHttpResponse( AddDonationResponse $responseModel, SelectedConfirmationPage $selectedPage ): Response {
+	private function newHttpResponse( AddDonationResponse $responseModel, SelectedConfirmationPage $selectedPage, array $sessionTrackingData ): Response {
 		switch( $responseModel->getDonation()->getPaymentType() ) {
 			case PaymentType::DIRECT_DEBIT:
 			case PaymentType::BANK_TRANSFER:
 				$httpResponse = new Response( $this->ffFactory->newDonationConfirmationPresenter()->present(
 					$responseModel->getDonation(),
 					$responseModel->getUpdateToken(),
-					$selectedPage
+					$selectedPage,
+					PiwikVariableCollector::newForDonation( $sessionTrackingData, $responseModel->getDonation() )
 				) );
 				break;
 			case PaymentType::PAYPAL:
--- a/app/RouteHandlers/ShowDonationConfirmationHandler.php
+++ b/app/RouteHandlers/ShowDonationConfirmationHandler.php
@@ -4,11 +4,11 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\App\RouteHandlers;
 
-use Silex\Application;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Response;
 use WMDE\Fundraising\Frontend\App\AccessDeniedException;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+use WMDE\Fundraising\Frontend\Infrastructure\PiwikVariableCollector;
 use WMDE\Fundraising\Frontend\UseCases\ShowDonationConfirmation\ShowDonationConfirmationRequest;
 
 /**
@@ -23,7 +23,7 @@ class ShowDonationConfirmationHandler {
 		$this->ffFactory = $ffFactory;
 	}
 
-	public function handle( ParameterBag $params ): Response {
+	public function handle( ParameterBag $params, array $sessionTrackingData ): Response {
 		$useCase = $this->ffFactory->newShowDonationConfirmationUseCase( $params->get( 'accessToken', '' ) );
 
 		$responseModel = $useCase->showConfirmation( new ShowDonationConfirmationRequest(
@@ -36,7 +36,8 @@ class ShowDonationConfirmationHandler {
 				$this->ffFactory->newDonationConfirmationPresenter()->present(
 					$responseModel->getDonation(),
 					$params->get( 'updateToken', '' ),
-					$selectedConfirmationPage
+					$selectedConfirmationPage,
+					PiwikVariableCollector::newForDonation( $sessionTrackingData, $responseModel->getDonation() )
 				)
 			);
 		}
--- a/app/RouteHandlers/ShowDonationConfirmationHandler.php
+++ b/app/RouteHandlers/ShowDonationConfirmationHandler.php
@@ -4,7 +4,7 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\App\RouteHandlers;
 
-use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use WMDE\Fundraising\Frontend\App\AccessDeniedException;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
@@ -23,11 +23,11 @@ class ShowDonationConfirmationHandler {
 		$this->ffFactory = $ffFactory;
 	}
 
-	public function handle( ParameterBag $params, array $sessionTrackingData ): Response {
-		$useCase = $this->ffFactory->newShowDonationConfirmationUseCase( $params->get( 'accessToken', '' ) );
+	public function handle( Request $request, array $sessionTrackingData ): Response {
+		$useCase = $this->ffFactory->newShowDonationConfirmationUseCase( $request->get( 'accessToken', '' ) );
 
 		$responseModel = $useCase->showConfirmation( new ShowDonationConfirmationRequest(
-			(int)$params->get( 'donationId', '' )
+			(int)$request->get( 'donationId', '' )
 		) );
 
 		if ( $responseModel->accessIsPermitted() ) {
@@ -35,7 +35,7 @@ class ShowDonationConfirmationHandler {
 			return new Response(
 				$this->ffFactory->newDonationConfirmationPresenter()->present(
 					$responseModel->getDonation(),
-					$params->get( 'updateToken', '' ),
+					$request->get( 'updateToken', '' ),
 					$selectedConfirmationPage,
 					PiwikVariableCollector::newForDonation( $sessionTrackingData, $responseModel->getDonation() )
 				)
--- a/app/routes.php
+++ b/app/routes.php
@@ -427,7 +427,7 @@ $app->match(
 	'show-donation-confirmation',
 	function( Application $app, Request $request ) use ( $ffFactory ) {
 		return ( new ShowDonationConfirmationHandler( $ffFactory ) )->handle(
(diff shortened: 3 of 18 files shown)


app/templates/Piwik.twig
```diff
@@ -0,0 +1,26 @@
...
+			g = d.createElement( 'script' ),
+			s = d.getElementsByTagName( 'script' )[ 0 ];
+		g.type = 'text/javascript';
+		g.async = true;
+		g.defer = true;
+		g.src = u + 'piwik.js';
+		s.parentNode.insertBefore( g, s );
+	} ) ();
+</script>
+<noscript><p><img src="{$ piwik.baseUrl $}piwik.php?idsite={$ piwik.siteId $}" style="border:0;" alt="" /></p></noscript>
```
gbirke: Why the `<p>` tag? won't this create an unnecessary line break?

KaiNissen: Not sure, if it creates a link break, but it's definitely not needed. I copied it from the piwik installation's gimme-the-snippet-thingy.

app/RouteHandlers/ShowDonationConfirmationHandler.php
```diff
@@ -36,12 +39,42 @@ public function handle( ParameterBag $params ): Response {
...
+			$piwikTracking->setCustomVariable(
+				3,
+				'Interval',
+				$sessionTrackingData['paymentInterval'] . '/' . $donation->getPaymentIntervalInMonths(),
+				'visit'
+			);
+		}
+		return $piwikTracking;
+	}
+
```
gbirke: Oh noes, duplicated code! Maybe refactor it to a `PiwikVariableCollector`?

tests/Integration/Presentation/Presenters/DonationConfirmationHtmlPresenterTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @covers WMDE\Fundraising\Frontend\Presentation\Presenters\DonationConfirmationHtmlPresenter
+ *
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DonationConfirmationHtmlPresenterTest extends \PHPUnit_Framework_TestCase {
+
+	public function testWhenSomethingHappens_itMustBeGood() {
```
gbirke: `testWhenPresenterRenders_itPassedParamsToTemplate`?

gbirke:
+1 from me, but would also like a comment from @JeroenDeDauw 

If you implement a `PiwikVariableCollector` you could also test it to ensure the variable values are concatenated correctly (silly, I know, but still ...). 

Maybe you can even figure out a way to do an edge-to-edge test that changes the predefined amount to assert that the variable contains the original and the user-selected value.

app/RouteHandlers/ShowDonationConfirmationHandler.php
JeroenDeDauw: How come we need this on both places?

app/routes.php
```diff
@@ -413,7 +414,8 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
 	'show-donation-confirmation',
 	function( Application $app, Request $request ) use ( $ffFactory ) {
 		return ( new ShowDonationConfirmationHandler( $ffFactory ) )->handle(
-			$request->getMethod() === Request::METHOD_GET ? $request->query : $request->request
+			$request->getMethod() === Request::METHOD_GET ? $request->query : $request->request,
+			$app['session']->get( 'piwikTracking', [] )
```
JeroenDeDauw: I got real confoosed by this indenting and the trinary for a bit :)

app/routes.php
```diff
@@ -455,7 +457,18 @@ function ( Application $app, Request $request ) use ( $ffFactory ) {
 	}
 );
 
-$app->get( '/', function ( Application $app ) {
+$app->get( '/', function ( Application $app, Request $request ) {
+	$app[ 'session' ]->set( 'piwikTracking', array_filter(
```
JeroenDeDauw: `$app['session']`

app/routes.php
```diff
@@ -455,7 +457,18 @@ function ( Application $app, Request $request ) use ( $ffFactory ) {
...
+		[
+			'paymentType' => $request->get( 'zahlweise', '' ),
+			'paymentAmount' => $request->get( 'betrag', '' ),
+			'paymentInterval' => $request->get( 'periode', '' )
+		],
+		function ( $value ) {
```
JeroenDeDauw: Can type hint `string`

src/Presentation/Presenters/DonationConfirmationHtmlPresenter.php
```diff
@@ -50,7 +54,8 @@ private function getConfirmationPageArguments( Donation $donation, string $updat
 			],
 			'person' => $this->getPersonArguments( $donation ),
 			'bankData' => $this->getBankDataArguments( $donation->getPaymentMethod() ),
-			'initialFormValues' => $this->getInitialMembershipFormValues( $donation )
+			'initialFormValues' => $this->getInitialMembershipFormValues( $donation ),
+			'piwikTracking' => $piwikTracking->getEvents()
```
JeroenDeDauw: There is some naming mismatch here. Is it correct to say that `PiwikCustomTracking` collects "piwik events"? Then perhaps it can be renamed to `PiwikEvents`, and `'piwikTracking'` here would become `'piwikEvents'` as well, assuming we can make such a rename.

src/Infrastructure/PiwikCustomTracking.php
```diff
@@ -0,0 +1,33 @@
...
+		$this->events[] = [ self::EVENT_SET_CUSTOM_VARIABLE, $variableId, $name, $value, $scope ];
+	}
+
+	public function trackGoal( int $goalId ) {
+		$this->events[] = [ self::EVENT_TRACK_GOAL, $goalId ];
+	}
+
+	public function getEvents(): array {
+		return $this->events;
+	}
```
JeroenDeDauw: There is some not entirely obvious binding here. `getEvents` only user is a thing where the result gets shoved into a template. When first looking at this class, that was not clear to me at all. Callers of the two event recording methods are thus using the same format as the template. That's very suspicious.

src/Infrastructure/PiwikCustomTracking.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Simple wrapper for tracking methods in Piwik's API
+ *
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class PiwikCustomTracking {
+
+	const SCOPE_VISIT = 'visit';
+	const SCOPE_PAGE = 'page';
```
JeroenDeDauw: Did you intend to use these in the route handlers? As far as I can tell they are unused now.

src/Infrastructure/PiwikCustomTracking.php
```diff
@@ -0,0 +1,33 @@
...
+
+	const EVENT_SET_CUSTOM_VARIABLE = 'setCustomVariable';
+	const EVENT_TRACK_GOAL = 'trackGoal';
+
+	private $events = [];
+
+	public function setCustomVariable( int $variableId, string $name, string $value, string $scope ) {
```
JeroenDeDauw: "set" is misleading, as this always adds something, and never updates

tests/Integration/Presentation/Presenters/DonationConfirmationHtmlPresenterTest.php
```diff
@@ -0,0 +1,78 @@
...
+			'piwikTracking' => [
+				[ 'setCustomVariable', 1234, 'Custom Variable', 'some value', 'visit' ],
+				[ 'trackGoal', 4095 ]
+			]
+		];
+	}
+
+	/**
+	 * @return PiwikCustomTracking
+	 */
```
JeroenDeDauw: Such PHP 5!

tests/Integration/Presentation/Presenters/DonationConfirmationHtmlPresenterTest.php
```diff
@@ -0,0 +1,78 @@
...
+			ValidDonation::newBookedAnonymousPayPalDonation(),
+			'update_token',
+			$pageSelector,
+			$this->newPiwikCustomTracking()
+		);
+	}
+
+	/**
+	 * @return array
+	 */
```
JeroenDeDauw: MAW

app/RouteHandlers/AddDonationHandler.php
```diff
@@ -183,4 +186,33 @@ private function isSubmissionAllowed( Request $request ) {
 		return true;
 	}
 
+	private function collectPiwikTrackingEvents( array $sessionTrackingData, Donation $donation ) {
+		$piwikTracking = new PiwikCustomTracking();
+		if ( isset( $sessionTrackingData['paymentType'] ) ) {
+			$piwikTracking->setCustomVariable(
+				1,
```
JeroenDeDauw: Where do these IDs come from? I'm guessing something will break if they get changed around here. If so, some indication of what that is, and how to verify correctness here would be nice. As far as I can tell no tests will break if they are changed here (just tried, and indeed, they don't). Perhaps this is a case where adding such tests makes little sense. However the combination of not realizing there is something that can be messed up and no detection of that happening short of the fun people coming into the room shouting at you is not good.

gbirke: I was thinking the same thing when I reviewed this. The IDs and their names are defined in Piwik. What the ID means is expressed in the next parameter, the name of the variable so it would probably be silly to have

```
$piwikTracking->setCustomVariable(
    self::VARIABLE_ID_AMOUNT,
    'Amount'
```

The meaning (and the IDs) of the variables is unlikely to change in the future.

tests/Integration/Presentation/Presenters/DonationConfirmationHtmlPresenterTest.php
KaiNissen: `testWhenUsingBogusTestMethodNames_iTendToForgetToChangeThemLater()`?

app/RouteHandlers/AddDonationHandler.php
KaiNissen: I also thought about what @gbirke wrote, but found it silly, too. The class now has constants for the variable IDs and a mapping array. It will throw an exception if `setCustomVariable` is called with an undefined ID.

src/Infrastructure/PiwikCustomTracking.php
KaiNissen: I intended a simple wrapper for Piwik's tracking API, thus calling the methods the same. I changed it to `trigger*()` to make it clearer.

src/Infrastructure/PiwikCustomTracking.php
KaiNissen: Yes, changed that.

src/Infrastructure/PiwikCustomTracking.php
KaiNissen: Would it be clearer, if I moved it into the `Presentation` namespace?

app/routes.php
KaiNissen: Unnecessary confusion, added another commit to remove that.

KaiNissen:
Addressed (almost) all of your comments. There is also an edge-to-edge test for them custom variables.

app/RouteHandlers/ShowDonationConfirmationHandler.php
KaiNissen: Mmmh. Actually the `AddDonationUseCase` should forward to the `ShowDonationConfirmationUseCase`. If we do that, collecting the custom variables is only needed there (unless we do more tracking in other use cases).

app/routes.php
```diff
@@ -455,7 +457,18 @@ function ( Application $app, Request $request ) use ( $ffFactory ) {
...
-$app->get( '/', function ( Application $app ) {
+$app->get( '/', function ( Application $app, Request $request ) {
+	$app['session']->set( 'piwikTracking', array_filter(
+		[
+			'paymentType' => $request->get( 'zahlweise', '' ),
+			'paymentAmount' => $request->get( 'betrag', '' ),
+			'paymentInterval' => $request->get( 'periode', '' )
+		],
+		function ( string $value ) {
+			return $value !== '';
```
gbirke: Please replace with 

    return $value !== '' && strlen($value) < 20;

Otherwise a potential DDOS opportunity would be opened up by posting "War and Peace" into the three fields and generating large session files rapidly, filling up server disk space.

20 characters should be enough for large sums of money ;-)

KaiNissen: good point

src/Infrastructure/PiwikCustomTracking.php
JeroenDeDauw: As I understand it, this is not a presentation class...

src/Infrastructure/PiwikEvents.php
```diff
@@ -0,0 +1,62 @@
...
+
+	const EVENT_SET_CUSTOM_VARIABLE = 'setCustomVariable';
+	const EVENT_TRACK_GOAL = 'trackGoal';
+
+	const CUSTOM_VARIABLE_PAYMENT_TYPE = 1;
+	const CUSTOM_VARIABLE_AMOUNT = 2;
+	const CUSTOM_VARIABLE_PAYMENT_INTERVAL = 3;
+
+	private $variableNames = [
+		1 => 'Payment',
```
JeroenDeDauw: I'd have used the constants here

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -504,4 +505,98 @@ public function testWhenTrackingCookieExists_valueIsPersisted() {
...
+				[
+					'betrag' => '5,00',
+					'periode' => 3,
+					'zahlweise' => 'Eh bien, mon prince. Gênes et Lucques ne sont plus que des apanages, des поместья, de la ' .
+						'famille Buonaparte. Non, je vous préviens que si vous ne me dites pas que nous avons la guerre, si ' .
+						'vous vous permettez encore de pallier toutes les infamies, toutes les atrocités de cet Antichrist ' .
+						'(ma parole, j’y crois) — je ne vous connais plus, vous n’êtes plus mon ami, vous n’êtes plus мой ' .
+						'верный раб, comme vous dites. Ну, здравствуйте,' .
+						'здравствуйте. Je vois que je vous fais peur, ' .
+						'садитесь и рассказывайте.'
```
JeroenDeDauw: What's with the random French and Russian(?)?

PR #590, opened 2016-07-28
Move donation form to new route

gbirke:
IMHO the "root" route can support posting params, but banners should rather post to the "pre-filled donation form" route to allow for different root route content in the future.

--

PR into https://github.com/wmde/FundraisingFrontend/pull/585


JeroenDeDauw:
+1, though concerns for the base PR remain

PR #591, opened 2016-07-29
Reduced state and made checks more explicit in BankDataValidator

JeroenDeDauw:
PR into #588

### DIFF
--- a/app/js/lib/form_validation.js
+++ b/app/js/lib/form_validation.js
@@ -88,6 +88,7 @@ var jQuery = require( 'jquery' ),
 					status: 'OK'
 				};
 			}
+
 			if ( formValues.debitType === 'sepa' ) {
 				data = {
 					iban: formValues.iban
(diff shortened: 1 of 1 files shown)


app/js/lib/form_validation.js
```diff
@@ -82,29 +82,40 @@ var jQuery = require( 'jquery' ),
...
+				}
+
+				return this.sendFunction(
+					this.validationUrlForSepa,
+					{
+						iban: formValues.iban
+					},
+					null,
+					'json'
+				);
```
JeroenDeDauw: It took me some digging to realize that the signature here is the one of `jQuery.get`. I'm quite dubious about the fact that we make web requests show up here, especially in a manner so prominent when reading the code here.

JeroenDeDauw: Huh... if `sendFunction` really is just jQuery.get, and all it does is decode JSON, then whatever is calling the `validate` method here directly consumes the format our API returns.

JeroenDeDauw:
Huh... if sendFunction really is just jQuery.get, and all it does is decode JSON, then whatever is calling the validate method here directly consumes the format our API returns.

--

I tried indicating the new methods are not public by prefixing them with an underscore, but I got the CI shouting at me. Do we have a way to indicate privateness?

gbirke:
Ah, JavaScript and "private". It will probably take some years until there is a language construct to be specified that doesn't need memory-hogging closure acrobatics to achieve this. For the time being, we could just mark them as private with [JSDoc comments](http://usejsdoc.org/tags-private.html) 😷

manicki:
Adding a `@private` doc might be the "best" thing to do here. Other way could be disabling JSCS's [disallowDanglingUnderscores](http://jscs.info/rule/disallowDanglingUnderscores) but this is definitely not a perfect solution.

PR #594, opened 2016-07-31
Moved files to reflect contexts

JeroenDeDauw:
https://phabricator.wikimedia.org/T139913

TODO:

* [ ] Duplicate PersonName and PhysicalAddress in donation and membership contexts
* [ ] Rename ApplicationContext as it is not a real context (I think)
* [ ] Finish going through Infrastructure/
* [ ] Look at ResponseModel/
* [ ] Look at Validation/
* [ ] Second look at Presentation/ (most things here are not part of the contexts; no domain binding)
* [ ] Move tests to match production code
* [ ] Fix FQNs created by FailStorm
* [ ] Simplify class names where now possible, ie MembershipApplicant to Applicant
* [ ] Investigate PaymentContext; perhaps PayPalContext + CreditCardContext, etc make more sense


JeroenDeDauw:
Note that I moved the Euro class into it's own little package (saved the trouble of making a place for it here ;p): https://github.com/wmde/Euro

This can be merged already as far as I'm concerned. Pretty much anything PHP that is done based off current master is going to conflict with this :)

PR #598, opened 2016-08-02
Scrub wiki comments from templates

gbirke:
The app output should not contain comments that were generated by the
Mediawiki CMS.

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -602,7 +602,12 @@ class FunFunFactory {
 
 	private function newPageContentModifier(): PageContentModifier {
 		return new PageContentModifier(
-			$this->getLogger()
+			$this->getLogger(),
+			[
+				'/<!--\s*NewPP limit report.*?-->/s' => '',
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
I'd be inclined to create some MediaWikiStuffStrippingContentModifierOrWhateverThinghy class that holds this config rather than putting it in the factory. Won't know if the regex is wrong now due to no test. I'm kinda muh-uh about it tho, so +1. Will look at this again tomorrow when I'm actually awake.

gbirke:
Since the `ContentModifier` is only used for MediaWiki pages, I'd rather use that class instead of creating a new one. But you are right, there are some missing tests which I should have written instead of massaging the regexes until the browser showed the correct result. 😢  

I'll modify this PR.

JeroenDeDauw:
Note that there is a TODO in `PageContentModifier`. I'm not entirely sure it is correct... Would also help to clarify if `PageRetriever` is MediaWiki specific or not.

gbirke:
Superseded by #611

PR #600, opened 2016-08-03
Duplicated PersonName to avoid dependency between contexts

JeroenDeDauw:
https://phabricator.wikimedia.org/T139913

Now DonorName and ApplicantName


gbirke:
At first I was against this change but Jeroen convinced me, that the two name objects are actually different - there is no possibility of anonymous memberships, only private and company.

PR #605, opened 2016-08-03
Use stable version of the Store component

JeroenDeDauw:
And fixed silly upper bound of the Euro one


gbirke:
I think you need to do a `composer update`, otherwise the deployment will probably complain.

JeroenDeDauw:
Done

PR #611, opened 2016-08-03
Refactor Wiki page content modification

gbirke:
Add tests for untested code and remove `ModifyingPageRetriever` class

PR #615, opened 2016-08-04
Re-add code that presumably got lost

JeroenDeDauw:
This behaviour was there before in PageContentModifier, which is no longer
used since https://github.com/wmde/FundraisingFrontend/pull/611

Not sure the trim is correct here

And not sure the drop of the code was not intentional. This PR is initially to
keep track of this. Tests are missing.


gbirke:
That code was left out intentionally - the result coming from the server is always `json_decode`d by `Mediawiki\Api\MediawikiApi::postRequest`, so it's impossible to have HTML in `$content`. When HTML is returned, `json_decode` will return `null` which is already caught by `ApiBasedPageRetriever` and the behavior is tested.

PR #620, opened 2016-08-04
Add license file

gbirke:


JeroenDeDauw:
I suggest to copy the one from the store component. Our inline docs and composer.json say v2+, not v3

gbirke:
Done.

JeroenDeDauw:
Any reason for this file name? It's different than what we have elsewhere

PR #626, opened 2016-08-09
Make deployment atomic

gbirke:
Change deployment scripts, divide into local build step and remote
deployment step.
Update deployment documentation.


KaiNissen:
+1

PR #627, opened 2016-08-10
Validate configuration

gbirke:
Validate the configuration with JSON Schema

### DIFF
--- a/app/config/schema.json
+++ b/app/config/schema.json
@@ -1,0 +1,593 @@
+{
+  "type": "object",
+  "title": "Fundraising app configuration",
+  "properties": {
+    "db": {
+      "type": "object",
+      "title": "Database settings",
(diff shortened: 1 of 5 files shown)


app/config/schema.json
```diff
@@ -0,0 +1,593 @@
...
+          "title": "Host name",
+          "description": "Database host name",
+          "default": ""
+        },
+        "port": {
+          "type": "integer",
+          "title": "Database port",
+          "default": ""
+        },
+        "driverOptions": {
```
KaiNissen: This is not defined in `config.dist.json`. Maybe we want to specify more clearly what this is and how it can be used?

gbirke: I've edited and expanded the description and explained what to put in there and why. Now that I look at this I think this should not be configurable but should be placed in the factory, depending on the driver name. But that's a different PR.

PR #628, opened 2016-08-12
Add robots.txt

gbirke:
The old application disallowed crawling the comments, the new app does
the same.


gbirke:
@KaiNissen @WMDE-Fisch Any Idea why the comments where blocked? Privacy reasons or performance reasons?

KaiNissen:
Privacy reasons, some people didn't want that to show up in search engines

PR #632, opened 2016-08-15
Validate payment type

gbirke:
When validating the amount, also check the payment type.

Fixes [phab:T142967](https://phabricator.wikimedia.org/T142967)

### DIFF
--- a/src/DonatingContext/UseCases/AddDonation/AddDonationValidator.php
+++ b/src/DonatingContext/UseCases/AddDonation/AddDonationValidator.php
@@ -16,7 +16,7 @@ use WMDE\Fundraising\Frontend\Validation\EmailValidator;
  * @author Gabriel Birke < gabriel.birke@wikimedia.de >
  */
 class AddDonationValidator {
-	private $amountValidator;
+	private $paymentDataValidator;
 	private $bankDataValidator;
 	private $emailValidator;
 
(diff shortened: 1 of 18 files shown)


web/res/js/donationForm.js
```diff
@@ -331,8 +332,7 @@ $( function () {
 
 	function paymentDataIsValid() {
 		var currentState = store.getState();
-		return currentState.validity.amount ||
-			( currentState.donationFormContent.amount && currentState.donationFormContent.paymentType ) ;
+		return currentState.validity.amount;
```
KaiNissen: This might be misleading. It actually is the validity of payment data (including the payment type).

src/DonatingContext/UseCases/AddDonation/AddDonationValidator.php
```diff
@@ -46,7 +46,7 @@ class AddDonationValidator {
 		Result::SOURCE_TRACKING_SOURCE => 250
 	];
 
-	public function __construct( AmountValidator $amountValidator, BankDataValidator $bankDataValidator,
+	public function __construct( PaymentDataValidator $amountValidator, BankDataValidator $bankDataValidator,
```
KaiNissen: argument name should be changed, too

tests/EdgeToEdge/WebRouteTestCase.php
```diff
@@ -151,7 +151,13 @@ protected function assertErrorJsonResponse( Response $response ) {
 		$responseData = $this->getJsonFromResponse( $response );
 		$this->assertArrayHasKey( 'status', $responseData );
 		$this->assertEquals( $responseData['status'], 'ERR' );
-		$this->assertArrayHasKey( 'message', $responseData );
+		$this->assertThat(
+			$responseData,
+			$this->logicalOr(
+				$this->arrayHasKey( 'message' ),
```
KaiNissen: Is this array key still used somewhere?

gbirke: `mesage` is still used in the `validate-fee` and `add-comment` JSON response. Do you think the assertion is too ugly?

KaiNissen: No, not too ugly, I just didn't see it used.

PR #633, opened 2016-08-16
Improve configuration documentation

KaiNissen:
fixed typos, clarified token length parameter


app/config/schema.json
```diff
@@ -80,7 +80,7 @@
       "type": "string",
       "minLength": 1,
       "title": "Password",
-      "description": "Passworf for the CMS wiki API",
+      "description": "Password for the CMS wiki API",
```
gbirke: Passworf

PR #635, opened 2016-08-16
Revert "Comment out cache profiling until bug is fixed"

JeroenDeDauw:
Reverts wmde/FundraisingFrontend#625

~~Blocked on https://github.com/JeroenDeDauw/GenericDecorator/issues/1~~


gbirke:
You'd probably want to check in the new `composer.lock` version with updated hash for the GenericDecorator release at the same time

PR #636, opened 2016-08-17
Force validation of default configuration as part of the CI

gbirke:
Refactored configuration validation.
Changed composer CI to validate config


app/ConfigValidation/ValidateConfigApplication.php
```diff
@@ -0,0 +1,45 @@
...
+namespace WMDE\Fundraising\Frontend\App\ConfigValidation;
+
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class ValidateConfigApplication extends Application {
```
JeroenDeDauw: `ConfigValidatingApplication` (needs to be a thing, not an action)

app/ConfigValidation/ValidateConfigApplication.php
```diff
@@ -0,0 +1,45 @@
...
+
+	protected function getCommandName( InputInterface $input )
+	{
+		return ValidateConfigCommand::NAME;
+	}
```
JeroenDeDauw: Such wrong bracket positions

app/ConfigValidation/ValidateConfigApplication.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\ConfigValidation;
+
+use Symfony\Component\Console\Application;
```
JeroenDeDauw: Took me some time to realize that the Application keyword here is from the Console component. I first thought it was the Silex one.

JeroenDeDauw: So this thing is completely independent from the Fundraising Frontent app PHP right? If so, I think it should not be in the app/ folder. How about a new sub folder in build/? Or perhaps a new folder altogether.

JeroenDeDauw: The new folder on the top level (ie cli/) makes more sense to me in case we see this as a general entry point to the application (ie a counterpart to the web one in app/) as opposed to a script just part of the build/CI.

JeroenDeDauw:
Looks reasonable to me. Some stuff that could be changed, though that can be done in follow ups.

You can avoid having the Application subclass by just creating an instance of Application and calling `add`, `setName` and `setVersion` on it. Have a look at what I did in https://github.com/JeroenDeDauw/Replicator

gbirke:
The reason why i subclassed `Application` was that I wanted to have [a single CLI script without having to specify commands](http://symfony.com/doc/current/components/console/single_command_tool.html). However, the `cli` folder sounds like a good idea. Then we can have other CLI commands in the future and ditch the application. I'll do a rebase and move stuff around.

JeroenDeDauw:
Thanks for explaining, I did not realize that part

PR #640, opened 2016-08-19
disable fee options also based on address type

KaiNissen:
fixes [phab:T143305](https://phabricator.wikimedia.org/T143305)

### DIFF
--- a/web/res/js/membershipForm.js
+++ b/web/res/js/membershipForm.js
@@ -150,12 +150,8 @@ $( function () {
 				stateKey: 'membershipFormContent.membershipType'
 			},
 			{
-				viewHandler: WMDE.View.createFeeOptionSwitcher( $( '#amount-1' ), 1 ),
-				stateKey: 'membershipFormContent.paymentIntervalInMonths'
-			},
-			{
-				viewHandler: WMDE.View.createFeeOptionSwitcher( $( '#amount-2' ), 6 ),
-				stateKey: 'membershipFormContent.paymentIntervalInMonths'
+				viewHandler: WMDE.View.createFeeOptionSwitcher( [ $( '#amount-1' ), $( '#amount-2' ), $( '#amount-3' ), $( '#amount-4' ), $( '#amount-5' ), $( '#amount-6' ), $( '#amount-7' ) ], { person: 24, firma: 100 } ),
+				stateKey: 'membershipFormContent'
 			},
 			{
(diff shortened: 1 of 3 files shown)


web/res/js/membershipForm.js
```diff
@@ -150,12 +150,8 @@ $( function () {
 				stateKey: 'membershipFormContent.membershipType'
 			},
 			{
-				viewHandler: WMDE.View.createFeeOptionSwitcher( $( '#amount-1' ), 1 ),
-				stateKey: 'membershipFormContent.paymentIntervalInMonths'
-			},
-			{
-				viewHandler: WMDE.View.createFeeOptionSwitcher( $( '#amount-2' ), 6 ),
-				stateKey: 'membershipFormContent.paymentIntervalInMonths'
+				viewHandler: WMDE.View.createFeeOptionSwitcher( [ $( '#amount-1' ), $( '#amount-2' ), $( '#amount-3' ), $( '#amount-4' ), $( '#amount-5' ), $( '#amount-6' ), $( '#amount-7' ) ], { person: 24, firma: 100 } ),
```
JeroenDeDauw: Such amount field naming :)

web/res/js/membershipForm.js
```diff
@@ -150,12 +150,8 @@ $( function () {
...
+				stateKey: 'membershipFormContent'
```
JeroenDeDauw: So the amounts are hardcoded here... If I'm not mistaken in the PHP they come from the config. Is there a way in the JS to get those values yet?

JeroenDeDauw:
+1

JeroenDeDauw:
> module.js:340
    throw err;
          ^
Error: Cannot find module 'lodash'
    at Function.Module._resolveFilename (module.js:338:15)

web/res/js/membershipForm.js
manicki: should clearly run from 'amount=0'

web/res/js/membershipForm.js
KaiNissen: These limits are hardcoded as a constant in `MembershipFeeValidator`. In some cases it would be nice to let client-side and server-side code share some configuration. Also regarding the messages (we keep them in JSON anyway). But we have to be careful, since there is stuff in the config we wouldn't want the client know about.

gbirke:
+2 from me with improvements in #642

PR #646, opened 2016-09-02
Moved Context folders out of src/ into contexts/

JeroenDeDauw:


JeroenDeDauw:
So far I've only moved the tests that I already moved once before. Those belonging to contexts but not organized as such still need moving (with or without this PR). Furthermore, this PR makes the namespaces of the newly moved tests mismatch with the directories. I'm leaving both of these things out for now, so we can do them after committing to this approach, ie: last chance to rage about this :)

gbirke:
I'm not sure I understand what this is all about. You're trying to create [Bounded Contexts](http://martinfowler.com/bliki/BoundedContext.html), right? From my understanding, this is necessary when the domain becomes very big and domain objects with the same name get different meanings. I'm not sure if our domain is big enough to justify that but I'm willing to go along.

Ideally, bounded contexts would not share domain objects, maybe share some interfaces, right? 
If that's true, then we need to look at the PaymentContext whose domain objects are used by the paypal notification and create donation use cases in the DonatingContext.

gbirke:
Apart from my general lack of understanding: I like the new `contexts` folder and the organization. +2 for that.

JeroenDeDauw:
What this change does is reorganize the files as we talked about before I went to SoCraTes, see the "Test directory structure" email. The organization into contexts is in the current code and is not changed by this PR.

> From my understanding, this is necessary when the domain becomes very big and domain objects with the same name get different meanings. I'm not sure if our domain is big enough to justify that but I'm willing to go along.

Let's talk about that outside of this PR

PR #651, opened 2016-09-10
Removed nonsensical DisplayPage UC

JeroenDeDauw:
Long overdue refactoring... we changed how this stuff worked in the first few weeks
after which the UC no longer made any sense and then never got rid of it.

So much stuff for what can simply be done with 3 lines of code :)


KaiNissen:
nice :)

PR #652, opened 2016-09-10
Speed up build by not waiting on coverage generation

JeroenDeDauw:


JeroenDeDauw:
I confirmed this works. Things green here while the coverage was still running on TravisCI

PR #662, opened 2016-09-26
fix validity handling when passing payment details

KaiNissen:


web/res/js/donationForm.js
```diff
@@ -409,6 +409,13 @@ $( function () {
 	store.dispatch( actions.newAddPageAction( 'personalData' ) );
 	store.dispatch( actions.newAddPageAction( 'bankConfirmation' ) );
 
+	// override validity
```
JeroenDeDauw: If a comment is needed, can't you extract the block into a function?

app/routes.php
```diff
@@ -493,7 +495,7 @@ function ( string $value ) {
 		),
 		HttpKernelInterface::SUB_REQUEST
 	);
-} );
+} )->method( 'GET|POST' );
```
JeroenDeDauw: What made you add this?

app/routes.php
```diff
@@ -380,15 +380,17 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
...
 // Show a donation form with pre-filled payment values, e.g. when coming from a banner
-$app->get( 'donation/new', function ( Request $request ) use ( $ffFactory ) {
-	$amount = Euro::newFromFloat( ( new AmountParser( 'de_DE' ) )->parseAsFloat( $request->get( 'betrag', '' ) ) );
+$app->match( 'donation/new', function ( Application $app, Request $request ) use ( $ffFactory ) {
+	$amount = $request->get( 'betrag', $request->get( 'betrag_auswahl', '' ) );
+	$validationResult = ( $ffFactory->newPaymentDataValidator() )->validate( $amount, (string) $request->get( 'zahlweise', '' ) );
 
 	return new Response( $ffFactory->newDonationFormPresenter()->present(
-		$amount,
+		Euro::newFromFloat( ( new AmountParser( 'de_DE' ) )->parseAsFloat( $amount ) ),
```
JeroenDeDauw: I was going to say "but what about the exception!", though then I realized you just moved this part. We still have some TODOs around the presentation specific parsing and formatting of values scattered around...

app/routes.php
```diff
@@ -380,15 +380,17 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
 );
 
 // Show a donation form with pre-filled payment values, e.g. when coming from a banner
-$app->get( 'donation/new', function ( Request $request ) use ( $ffFactory ) {
-	$amount = Euro::newFromFloat( ( new AmountParser( 'de_DE' ) )->parseAsFloat( $request->get( 'betrag', '' ) ) );
+$app->match( 'donation/new', function ( Application $app, Request $request ) use ( $ffFactory ) {
+	$amount = $request->get( 'betrag', $request->get( 'betrag_auswahl', '' ) );
+	$validationResult = ( $ffFactory->newPaymentDataValidator() )->validate( $amount, (string) $request->get( 'zahlweise', '' ) );
```
JeroenDeDauw: What's with the brackets around `$ffFactory->newPaymentDataValidator()`?

web/res/js/donationForm.js
KaiNissen: just improved the code together with glorious @gbirke which included removing these lines

app/routes.php
KaiNissen: they are not needed :P

app/routes.php
KaiNissen: just thought it would be nice to be able to use this route as the target of banner forms. we can, of course, use `donation/new` if you object this.

KaiNissen:
I think, this can only be covered by browser tests. Or maybe @gbirke has an idea?

gbirke:
Our JavaScript unit tests only test components in isolation and mock the jQuery elements. 

We *could* introduce some "Integration testing" testing where we check the DOM with fixed templates (to avoid errors resulting from faulty PHP code), but I that's so close to browser testing that I think browsers tests are better suited to that and catch more errors (errors that come from PHP/Javascript interaction).

PR #663, opened 2016-09-27
Avoid exception with negative donation amount

gbirke:
Since the Euro class is used for building the requests and it does not
allow for negative values, we initialize it with 0 if the amount is
lower than 0.

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -168,8 +168,12 @@ class AddDonationHandler {
 
 	private function getEuroAmountFromString( string $amount ) {
 		$locale = 'de_DE'; // TODO: make this configurable for multilanguage support
+		try {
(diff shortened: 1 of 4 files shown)


gbirke:
@JeroenDeDauw Not sure if this is the most optimal solution. Maybe Euro is a domain class that should not be used on the framework level?

JeroenDeDauw:
I'm not sure about using `Euro` in the request models either, as it indeed means you need to deal with the exception coming from negative numbers in the presentation layer. Still, I'm leaning towards this approach because you need to handle errors in the presentation layer anyway. If you have an int in your request model, the user can still provide `foo`, which will fail to be parsed and need to be handled. In addition, it's not clear "needs to be unsigned int" is actually more of a domain thing than "needs to be int", as PHP could just as well have native support for the former type of number.

JeroenDeDauw:
I'd have expected to abort the request in the framework layer and not invoke the UC at all if there are parse errors. Perhaps this approach is better though, at least in this case - it certainly seems simple enough.

PR #660, opened 2016-09-23
 	Reworked the PurgeCacheUseCase into an application service

JeroenDeDauw:
Replaces https://github.com/wmde/FundraisingFrontend/pull/653

PR #666, opened 2016-09-28
fix class constant names

KaiNissen:
#660 broke the route. I'm introducing a simple route test just for the failure case to make sure the route exists. Everything else is covered in the integration test.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -518,9 +518,9 @@ $app->get( '/purge-cache', function( Request $request ) use ( $ffFactory ) {
 
 	return new Response(
 		[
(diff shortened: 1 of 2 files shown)


KaiNissen:
sorry @manicki, there is no way to reserve a PR number for someone.

JeroenDeDauw:
\o/ thanks @KaiNissen

manicki:
@KaiNissen there are plenty 666s waiting in other repos so I'll be fine.
You could have at least set `INVALID_SECRET` in the test to some black metal lyric verse, though.

PR #668, opened 2016-09-28
Add legacy routing

gbirke:
After several hours of trying to get Nginx rewriting to work (together
with having the correct value for REQUEST_URI which is needed for our PHP
routing) I gave up and implemented the legacy routing in PHP.

When
https://serverfault.com/questions/805881/nginx-populate-request-uri-with-rewritten-url
is answered, this commit can be reverted.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -496,9 +496,25 @@ $app->get( '/', function ( Application $app, Request $request ) {
 	);
 } );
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I know rite, webserver config is such great fun, esp when its routing

PR #670, opened 2016-09-29
Changed wrong field name for company

gbirke:
initialValues expects `companyName` instead of `company`.

Also fixed the wrong field name in the wiki template.


KaiNissen:
gosh

PR #672, opened 2016-09-29
provide update token to donation confirmation page

KaiNissen:


JeroenDeDauw:
Looks good to me but I'm kinda tired so just +1

PR #676, opened 2016-09-30
Make TextPolicyValidatorTest not fail when ISP is derp

JeroenDeDauw:


tests/System/TextPolicyValidatorTest.php
```diff
@@ -69,14 +71,28 @@ public function testWhenGivenHarmlessComment_validatorReturnsTrue( $commentToTes
...
 			[ 'Bei Wikipedia kann man eine Menge zum Thema Hamster finden. Hamster fressen voll viel Zeug alter!' ],
 			[ 'Manche Seiten haben keinen Inhalt, das finde ich sch...e' ], // this also tests the domain detection
 		];
 	}
 
+	public function testHamlessContentWithDns() {
+		$this->skipIfNoInternet();
+
+		if ( checkdnsrr( 'some-non-existing-domain-drfeszrfdaesr.sdferdyerdhgty', 'A' ) ) {
+			$this->markTestSkipped( 'Sure, blame it on your ISP!' );
```
gbirke: I'd rather have a more informative message like "Your DNS/ISP provider gives results for impossible host names."

JeroenDeDauw: BUT BUT BUT

https://www.youtube.com/watch?v=HGBOeLdm-1s

JeroenDeDauw: Well we can only cater to the uneducated so much... :)

Amended now

PR #679, opened 2016-10-04
send all contact form field values

KaiNissen:


src/ApplicationContext/UseCases/GetInTouch/GetInTouchUseCase.php
```diff
@@ -18,13 +18,14 @@
...
+	private $confirmingMailer;
 
-	public function __construct( GetInTouchValidator $validator, Messenger $messenger, TemplateBasedMailer $templateMailer ) {
+	public function __construct( GetInTouchValidator $validator, TemplateBasedMailer $forwardingMailer,
+								 TemplateBasedMailer $confirmingMailer ) {
 		$this->validator = $validator;
-		$this->messenger = $messenger;
-		$this->templateBasedMailer = $templateMailer;
+		$this->forwardingMailer = $forwardingMailer;
+		$this->confirmingMailer = $confirmingMailer;
```
JeroenDeDauw: The method in the factory is called `newContactConfirmationMailer`. `confirmationMailer` seems more correct to me than `confirmingMailer`. Not sure about `forwardingMailer` as I don't get what it means. What do you think?

KaiNissen: @gbirke said something similar and is working on a follow-up

tests/Fixtures/TemplateBasedMailerSpy.php
```diff
@@ -25,6 +25,10 @@ public function sendMail( EmailAddress $recipient, array $templateArguments = []
 		$this->sendMailCalls[] = func_get_args();
 	}
 
+	public function sendMailToOperator( EmailAddress $replyToAddress, array $templateArguments = [] ) {
+		$this->sendMailCalls[] = func_get_args();
```
JeroenDeDauw: But those are not calls to `sendMail`

src/Infrastructure/TemplateBasedMailer.php
```diff
@@ -36,4 +36,17 @@ public function sendMail( EmailAddress $recipient, array $templateArguments = []
 		);
 	}
 
+	/**
+	 * @throws \RuntimeException
+	 */
+	public function sendMailToOperator( EmailAddress $replyToAddress, array $templateArguments = [] ) {
```
JeroenDeDauw: IIRC I introduced this TemplateBasedMailer in part to not expose this method to all its users, which generally don't need it.

What made you add it here?

KaiNissen: What made me add this method is the lack of support for `Messenger::sendMessageToOperator()` in this class. We want to use a template for the message content, but use the fixed operator's address as the recipient and add a `Reply-To` header for convenience.

We could instead introduce another parameter for the value of the `Reply-To` header. That would force us to make all parameters default to `null` (since `Messenger::sendMessageToOperator()` does not expect the recipient to be passed). Or we could pass a flag to let the method know what the passed e-mail address should be used as.

But I don't prefer any of the above approaches to adding another method. What do you think?

JeroenDeDauw: I also think that what you did is nicer than the alternatives you described

JeroenDeDauw: I had a proper look at this in my IDE now.

The instance on which `sendMailToOperator` is being called only has that method being called upon it. And all other instances of `TemplateBasedMailer` do not have `sendMailToOperator` called on it. That means there is no reason to put them in the same interface. (See also: https://en.wikipedia.org/wiki/Interface_segregation_principle)

How about creating an `OperatorMailer` or `TemplateBasedOperatorMailer` that is like the existing `TemplateBasedMailer` but only has the operator method? An alternative approach is to keep them in the same class but have two interfaces, though this is way more disruptive, as you'd need to update existing type hints to one of the interfaces.

Btw, I'm not so sure about the `TemplateBased` part in the name. The implementation sure depends on a twig template, but from the user perspective, the interface just takes a bunch of arguments. I'd be nicer for the name to reflect the second part. `ArgumentTakingMailer` is somehow weird tho.

JeroenDeDauw: I was not able to do this now yet... can do it tomorrow. If you agree with what I propose, feel free to merge already, then we can do this change in a follow up.

PR #681, opened 2016-10-04
Improve naming of mailers

gbirke:
Change names of mailers according to the people who receive mails.
Add a test to check if the user mail is actually sent.
Improve TemplateBasedMailerSpy to accept calls without template params
(func_get_args doesn't get default params, only the real params of the
function call).


JeroenDeDauw:
> func_get_args doesn't get default params, only the real params of the function call

php

PR #682, opened 2016-10-05
Prefix asset files

gbirke:
Goal: Add a cache-busting prefix to asset files.
This solves https://phabricator.wikimedia.org/T147399

The prerequisite Nginx configuration from the [infrastructure repo](
https://github.com/wmde/fundraising-infrastructure/pull/62/commits/eff4e55f5696cd2628e75c62fc4016e760f1825b) is already deployed on the servers.

In this PR:
- Add Twig extension that adds a file prefix, if one is configured.
- Generate a prefix value when deploying

### DIFF
--- a/deployment/tasks/deploy_to_server.yml
+++ b/deployment/tasks/deploy_to_server.yml
@@ -29,6 +29,9 @@
 - name: Create var/cache dir
   file: path={{ app_dir }}/{{ release_prefix }}{{ release_name }}/var/cache owner={{app_owner}} group={{app_group}} state=directory mode=775
 
+- name: Create cache-busting prefix
+  copy: dest={{ app_dir }}/{{ release_prefix }}{{ release_name }}/var/file_prefix.txt content="{{ release_name|hash('md5') }}" owner={{app_owner}} group={{app_group}}
+
 - name: Create symlink to log dir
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
Comment on Phab

deployment/tasks/deploy_to_server.yml
```diff
@@ -29,6 +29,9 @@
 - name: Create var/cache dir
   file: path={{ app_dir }}/{{ release_prefix }}{{ release_name }}/var/cache owner={{app_owner}} group={{app_group}} state=directory mode=775
 
+- name: Create cache-busting prefix
+  copy: dest={{ app_dir }}/{{ release_prefix }}{{ release_name }}/var/file_prefix.txt content="{{ release_name|hash('md5') }}" owner={{app_owner}} group={{app_group}}
```
JeroenDeDauw: Perhaps I should RTFM... what is `{{ release_name }}` here anyway?

gbirke: `release_name` is a variable passed in from the "parent" playbook `deployment.yml`. It contains a timestamp in format YYYYMMDDHHMMSS. 

There is not much manual to read, I've tried to improve that situation with #684

PR #686, opened 2016-10-10
Upgraded to Silex 2.x

JeroenDeDauw:
Includes updating of various dependencies to match, most notably
Pimple from 1.x to 3.x

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -9,22 +9,23 @@
 declare( strict_types = 1 );
 
 use Silex\Application;
+use Silex\Provider\RoutingServiceProvider;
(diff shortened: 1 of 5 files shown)


gbirke:
Looking at all the assignments I'm wondering if we really need Pimple if we're using the factory pattern. 

+2 Apart from that.

JeroenDeDauw:
Well that is a +2 then? Since I did not change our pimple usage, and if we do that, it's better to do that in a different PR.

The benefit Pimple provides is the caching/sharing of the service instances. We could also have a field for each and a bunch of if statements to see if the thing is null and if so initiliaze. Or perhaps an array, much like Pimple, still with the if stuff tho.

PR #690, opened 2016-10-10
Display warning for recurring PPL donations

gbirke:
This is for https://github.com/wmde/fundraising/issues/1078

### DIFF
--- a/app/js/lib/view_handler/animator.js
+++ b/app/js/lib/view_handler/animator.js
@@ -1,0 +1,62 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+
+	ElementAnimator = {
+		el: null,
+		showElement: function () {
+			throw new Error( 'This is an abstract class!' );
+		},
+		hideElement: function () {
+			throw new Error( 'This is an abstract class!' );
+		}
+	},
+
+	SlidingElementAnimator = objectAssign( Object.create( ElementAnimator ), {
+		// internal fields
+		slideSpeed: 600,
+
+		showElement: function () {
+			this.el
+				.slideDown( this.slideSpeed )
+				.animate(
+					{ opacity: 1 },
+					{ queue: false, duration: this.slideSpeed }
+				);
+		},
+		hideElement: function () {
+			this.el
+				.slideUp( this.slideSpeed )
+				.animate(
+					{ opacity: 0 },
+					{ queue: false, duration: this.slideSpeed }
+				);
+		}
+	} ),
+
+	SimpleElementAnimator = objectAssign( Object.create( ElementAnimator ), {
+		showElement: function () {
+			this.el.show();
+		},
+		hideElement: function () {
+			this.el.hide();
+		}
+	} );
+
+module.exports = {
+	createSimpleElementAnimator: function ( element ) {
+		return objectAssign(
+			Object.create( SimpleElementAnimator ),
+			{ el: element }
+		);
+	},
+
+	createSlidingElementAnimator: function ( element ) {
+		return objectAssign(
+			Object.create( SlidingElementAnimator ),
+			{ el: element }
+		);
+	},
+
+	ElementAnimator: ElementAnimator
+};
--- a/app/js/lib/view_handler/ppl_recurrent_warning.js
+++ b/app/js/lib/view_handler/ppl_recurrent_warning.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+
+	/**
+	 * View Handler for showing and hiding elements if the update value matches a regular expression
+	 * @class
+	 */
+	PPLRecurrentWarning = {
+		isHidden: null,
+		animator: null,
+
+		update: function ( formContent ) {
+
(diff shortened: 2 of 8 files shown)


app/js/lib/view_handler/ppl_recurrent_warning.js
```diff
@@ -0,0 +1,38 @@
...
+	PPLRecurrentWarning = {
+		element: null,
+		isHidden: null,
+
+		update: function ( formContent ) {
+
+			if ( formContent.paymentType === 'PPL' && formContent.paymentIntervalInMonths > 0 ) {
+				if ( this.isHidden === false ) {
+					return;
+				}
```
JeroenDeDauw: Somehow it seems weird to use a guard clause here

app/js/lib/view_handler/ppl_recurrent_warning.js
```diff
@@ -0,0 +1,38 @@
...
+				this.element.show();
+				this.isHidden = false;
+			} else {
+				if ( this.isHidden === true ) {
+					return;
+				}
+				this.element.hide();
+				this.isHidden = true;
+			}
```
JeroenDeDauw: Is there an issue with calling `show` or `hide` on an element that already has the respective state? Like this

```js
this.isHidden = formContent.paymentType !== 'PPL' || formContent.paymentIntervalInMonths === 0
if ( this.isHidden ) {
    this.element.hide();
}
else {
    this.element.show();
}
```

Of course its even simpler if `this.element` has some `setVisibility`.

app/js/tests/view_handlers/test_ppl_recurrent_warning.js
```diff
@@ -0,0 +1,61 @@
...
+		hide: sinon.spy()
+	};
+}
+
+test( 'When interval is recurring, warning is shown', function ( t ) {
+	var element = createElement(),
+		formValues = {
+			paymentType: 'PPL',
+			paymentIntervalInMonths: 12
+	    },
```
JeroenDeDauw: Indent is foobar

app/js/tests/view_handlers/test_ppl_recurrent_warning.js
```diff
@@ -0,0 +1,61 @@
...
+		handler = createPPLRecurrentWarningHandler( element );
+
+	handler.update( formValues );
```
JeroenDeDauw: Why did you create a local var for `formValues`? You can just pass it in here no? More local...

app/js/lib/view_handler/ppl_recurrent_warning.js
gbirke: I copied that code from the `VisibilitySwitcher` class where it is needed to avoid unneccessary animations. I need to refactor this class to use the `Animator` interface.

app/js/lib/view_handler/animator.js
```diff
@@ -0,0 +1,62 @@
...
+
+	ElementAnimator = {
+		el: null,
+		showElement: function () {
+			throw new Error( 'This is an abstract class!' );
+		},
+		hideElement: function () {
+			throw new Error( 'This is an abstract class!' );
+		}
+	},
```
JeroenDeDauw: Don't we already have something like this?

Anyway, why did you create this "interface"? Is it "type hinted" against anywhere?

JeroenDeDauw:
+1... no objection against merging. Ping @KaiNissen 

Somehow I'm not entirely reassured by the tests, as we only have a unit test here. If something is wrong with the wiring, or the unit test includes wrong assumptions, we won't notice.

gbirke:
@JeroenDeDauw I agree, I'm also not assured. I've added it as a subtask for the browser tests: https://phabricator.wikimedia.org/T147941

JeroenDeDauw:
> But that should not prevent this from being merged.

Yeah sure. Else I would not have +1'd. As I tried indicating, I'd like Kai to also have a look and merge if he's also fine with it.

app/js/lib/view_handler/ppl_recurrent_warning.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+
+	/**
+	 * View Handler for showing and hiding elements if the update value matches a regular expression
+	 * @class
+	 */
+	PPLRecurrentWarning = {
```
KaiNissen: It becomes pretty clear what this thing does and under what circumstances when looking at the code. The name might not be self-explanatory, though. Maybe something like `PaymentTypeNotice` or at least `PayPalRecurrentWarning`? Not a big thing, though. You might as well ignore my comment.

JeroenDeDauw: +1 to using `PayPal` instead of `PPL`

app/js/lib/view_handler/ppl_recurrent_warning.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var objectAssign = require( 'object-assign' ),
+
+	/**
+	 * View Handler for showing and hiding elements if the update value matches a regular expression
```
gbirke: Done

PR #693, opened 2016-10-12
Fixed validation bug in donation/new route

JeroenDeDauw:
https://phabricator.wikimedia.org/T147826

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -374,7 +374,7 @@ $app->post(
 	'donation/add',
 	function( Application $app, Request $request ) use ( $ffFactory ) {
 		return ( new AddDonationHandler( $ffFactory, $app ) )
-			->handle( $request, $app['session']->get( 'piwikTracking', [] ) );
+			->handle( $request );
 	}
 );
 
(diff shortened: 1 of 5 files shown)


app/routes.php
```diff
@@ -387,12 +387,16 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
 	}
 	$validationResult = $ffFactory->newPaymentDataValidator()->validate( $amount, (string) $request->get( 'zahlweise', '' ) );
 
-	return new Response( $ffFactory->newDonationFormPresenter()->present(
-		$amount,
-		$request->get( 'zahlweise', '' ),
-		intval( $request->get( 'periode', 0 ) ),
-		$validationResult->isSuccessful()
-	) );
+	// TODO: don't we want to use newDonationFormViolationPresenter when !$validationResult->isSuccessful()?
```
JeroenDeDauw: I tried creating an edge-to-edge test for the route but found that the output is the same for both successful and unsuccessful requests. And that a bunch of the output is not getting the values it expects.

```
    <main id="main">
        <div class="container">
                Amount: 5,51
Payment type: BEZ
Interval: 0
IBAN: 
BIC: 
Bank name: 
Address type: 
Salutation: 
Title: 
Company: 
First name: 
Last name: 
Street: 
Postal code: 
City: 
Country code: 
Email address: 

Violations:
        </div>
    </main>
```

I looked at this stuff a bit and am not entirely sure what is wrong, if anything, but it seems highly suspicious.

By all means do already merge this PR; it does not need to fix all the issues in one go.

JeroenDeDauw: @gbirke any idea about what I wrote here?

PR #694, opened 2016-10-12
Fix paypal handler

gbirke:
Various code changes to make the handling behave more like the old application.

Based on the exceptions that were thrown on the server, both either stemming from `item_name` or `payment_status`

### DIFF
--- a/src/Infrastructure/LoggingPaymentNotificationVerifier.php
+++ b/src/Infrastructure/LoggingPaymentNotificationVerifier.php
@@ -30,8 +30,13 @@ class LoggingPaymentNotificationVerifier implements PaymentNotificationVerifier 
 		try {
 			$this->verifier->verify( $request );
 		} catch ( PayPalPaymentNotificationVerifierException $exception ) {
-			$this->logger->log( $this->logLevel, $exception->getMessage(), [ self::CONTEXT_EXCEPTION_KEY => $exception ] );
-			$this->logger->log( LogLevel::DEBUG, 'Paypal request data', $request );
+			if ( $exception->getCode() == PayPalPaymentNotificationVerifierException::ERROR_UNSUPPORTED_STATUS ) {
+				$this->logger->log( LogLevel::INFO, 'Unsupported payment_status', $request );
+			} else {
+				$this->logger->log( $this->logLevel, $exception->getMessage(), [ self::CONTEXT_EXCEPTION_KEY => $exception ] );
+				$this->logger->log( LogLevel::DEBUG, 'Paypal request data', $request );
+			}
+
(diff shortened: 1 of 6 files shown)


src/Infrastructure/LoggingPaymentNotificationVerifier.php
```diff
@@ -30,8 +30,13 @@ public function verify( array $request ) {
 		try {
 			$this->verifier->verify( $request );
 		} catch ( PayPalPaymentNotificationVerifierException $exception ) {
-			$this->logger->log( $this->logLevel, $exception->getMessage(), [ self::CONTEXT_EXCEPTION_KEY => $exception ] );
-			$this->logger->log( LogLevel::DEBUG, 'Paypal request data', $request );
+			if ( $exception->getCode() == PayPalPaymentNotificationVerifierException::ERROR_UNSUPPORTED_STATUS ) {
+				$this->logger->log( LogLevel::INFO, 'Unsupported payment_status', $request );
```
gbirke: With our current logging settings, these will be dropped. Should we log this as warning?

tests/Unit/Infrastructure/LoggingPaymentNotificationVerifierTest.php
```diff
@@ -67,6 +71,38 @@ private function assertExceptionLoggedAsCritical( string $expectedExceptionType,
...
+	}
+
+	private function assertUnsupportedMethodWasLogged( string $expectedMethodName, LoggerSpy $logger ) {
+		$logCalls = $logger->getLogCalls();
+
+		$this->assertCount( 1, $logCalls, 'There should only be one log call' );
+		$logCall = $logCalls[0];
+
+		$this->assertSame( LogLevel::INFO, $logCall[0] );
+		$this->assertArrayHasKey( 'payment_status', $logCall[2], 'the log context should contain an exception element' );
```
JeroenDeDauw: payment_status element

JeroenDeDauw:
Follow up https://github.com/wmde/FundraisingFrontend/pull/695

src/Infrastructure/LoggingPaymentNotificationVerifier.php
gbirke: The IPN notification endpoint is expected to at least receive all `payment_status` messages from PPL and return a "200 OK". It does not need to do something with them. Until the FR people request more status handled, we just ignore them.

The exception thing is a bit weird, I agree. The alternative would be to check the status in the route handler and create a Response before calling the veryfier. That way, we could even implement status-based routing later if we need it (I know, YAGNI). Go ahead if you like.

PR #700, opened 2016-10-13
switch amount parser locale to en_EN

KaiNissen:
The banners are converting the amount input into a parsable float before sending POSTing. To make sure older banners are still working, I changed the locale the `AmountParser` is using to `en_EN`.

see [phab:T148039](https://phabricator.wikimedia.org/T148039)

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -381,7 +381,7 @@ $app->post(
 // Show a donation form with pre-filled payment values, e.g. when coming from a banner
 $app->get( 'donation/new', function ( Request $request ) use ( $ffFactory ) {
 	try {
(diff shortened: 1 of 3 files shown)


gbirke:
Ugh, this is inconsistent with the locale in the "Add donation". I fear that using different locales just to be backwards compatible to old banners will come back to haunt us. I'll merge this quick fix, but we really need to find another solution.

JeroenDeDauw:
Hmm... it's in the presentation layer, so I don't think it's that much of an issue. You can imagine having different languages in different presentation stuffs and presentation stuffs that support multiple by means of some parameter.

PR #701, opened 2016-10-14
Log potential fatal errors

gbirke:
There are some cases where the paypal IPN fails. The error log shows
that this happens when the `titel` data field is not set. This should
never happen, but apparently it does. Added logging code to see which
donations are affected.

This is for https://phabricator.wikimedia.org/T147785


gbirke:
I know this is ugly and should be reverted as soon as we found out something. I'm totally clueless how to solve this any other way and I'm open for suggestions.

JeroenDeDauw:
Why not use a real logger?

gbirke:
I did not use a real logger because the `DoctrineDonationRepository` doesn't have one and I hope to remove the code I inserted in this request soon. No I realize I could have injected one in the same commit and then just reverted. Oh well ...

PR #704, opened 2016-10-15
Use a more descriptive and short name for the logger

JeroenDeDauw:


web/index.php
```diff
@@ -30,7 +30,7 @@
 $ffFactory->enablePageCache();
 
 $ffFactory->setLogger( call_user_func( function() use ( $ffFactory ) {
-	$logger = new Logger( 'WMDE Fundraising Frontend logger' );
+	$logger = new Logger( 'index.php logger' );
```
gbirke: The actual logging is not done in index.php but at other places that use the logger, so calling it index.php is a bit too confusing fir my taste.

The name given here is used as the "channel" designation in the log messages. So the `logger` suffix can be dropped IMO. Why not just call it `Fundraising Frontend` ?

gbirke:
I've tried out more descriptive names for the channel but they all were way too long. So I've created #705 that has even shorter names and doesn't use file names.

PR #707, opened 2016-10-17
Change configuration deployment

gbirke:
Change config deploy mechanism from symlinks in target servers to config files on deployment server.

Please merge #706 first


JeroenDeDauw:
+1

PR #710, opened 2016-10-18
Fix paypal donation update

gbirke:
This solves two problems:

1) Loading of exported donations failed because their donor information
is purged. The email address was checked for a null value, which was not
a good indicator (since it's set to empty string by the backup script)
Now we handle the two cases where there is no donor information -
address type is "anonym" and doctrine entity is marked as "exported"
(the export purges the data).

2) Writing back information of donations with no donor information
caused the full name to be overwritten with `Anonym` which is not
correct. Now we only write the name when the doctrine donation is new.


KaiNissen:
code looks good. @JeroenDeDauw?

JeroenDeDauw:
+1 looks good to me, but I'm too tired to +2. Pong @KaiNissen  :)

PR #713, opened 2016-10-19
Compare tokens using hash_equals

JeroenDeDauw:
Perhaps timing attacks where possible. Perhaps they still are, though this
ought to mittegate at least something :)


WMDE-Fisch:
Good to know, just googled it.

PR #716, opened 2016-10-19
Ignore missing PayPal address status

gbirke:
Ignore missing 'paypal_address_status' when constructing the paypal data
from the data blob.

Should be either "confirmed" or "unconfirmed", but is sometimes not
written by the IPN callback, so we need to handle entities that don't
have it.

This is for https://phabricator.wikimedia.org/T147785


JeroenDeDauw:
> is sometimes not written by the IPN callback

Why not? Because we don't get the info?

gbirke:
Yes, not all info is provided by the callback. The new applications guards against that by using empty strings as default in `PayPalNotificationHandler::newUseCaseRequestFromPost` but recurrent donations just don't have the fields in their data blob.

PR #717, opened 2016-10-20
Improve IBAN handling and BIC field checking

gbirke:
This is for fixing https://phabricator.wikimedia.org/T148620

### DIFF
--- a/contexts/PaymentContext/src/Domain/Model/Iban.php
+++ b/contexts/PaymentContext/src/Domain/Model/Iban.php
@@ -13,7 +13,7 @@ class Iban {
 	private $iban;
 
 	public function __construct( string $iban ) {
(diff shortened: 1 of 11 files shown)


contexts/PaymentContext/src/Domain/Model/Iban.php
```diff
@@ -13,7 +13,7 @@ class Iban {
 	private $iban;
 
 	public function __construct( string $iban ) {
-		$this->iban = preg_replace( '/\s+/', '', $iban );
+		$this->iban = strtoupper( preg_replace( '/\s+/', '', $iban ) );
```
JeroenDeDauw: oops :D

JeroenDeDauw:
CR+2 for PHP, CR+1 for JS. Such follow up https://github.com/wmde/FundraisingFrontend/pull/719

PR #723, opened 2016-10-24
Fix legacy tracking data handling

gbirke:
Be more lenient when loading tracking data from old entries.

Refactor ValidDoctrineDonation to create the valid legacy entries.

This is for the newly found errors in https://phabricator.wikimedia.org/T147785


gbirke:
Incorporated into #724

PR #725, opened 2016-10-24
handle IPNs for membership applications

KaiNissen:


contexts/MembershipContext/src/DataAccess/DoctrineApplicationRepository.php
```diff
@@ -142,13 +169,17 @@ private function getDoctrineStatus( Application $application ): int {
 			$status += DoctrineApplication::STATUS_CANCELED;
 		}
 
-		if ( $status === DoctrineApplication::STATUS_NEUTRAL ) {
+		if ( $status === DoctrineApplication::STATUS_NEUTRAL && $this->isDirectDebitPayment( $application ) ) {
 			$status += DoctrineApplication::STATUS_CONFIRMED;
```
gbirke: So this means that direct debit payments are "autoconfirmed"? Isn't there a confirmation that has to be set in the backend?

contexts/MembershipContext/src/DataAccess/DoctrineApplicationRepository.php
```diff
@@ -236,7 +268,7 @@ private function newPaymentMethod( DoctrineApplication $application ): PaymentMe
 		}
 
 		if ( $application->getPaymentType() === PaymentType::PAYPAL ) {
-			return new PayPalPayment();
+			return new PayPalPayment( $this->newPayPalData( $application ) );
```
gbirke: This will fail with a TypeError when `newPaypalData` returns `null`.

contexts/MembershipContext/src/Domain/Model/Application.php
```diff
@@ -103,4 +113,46 @@ public function needsModeration(): bool {
 		return $this->needsModeration === self::NEEDS_MODERATION;
 	}
 
+	public function isConfirmed(): bool {
+		return $this->isConfirmed === self::IS_CONFIRMED;
+	}
+
+	public function setConfirmed() {
```
gbirke: maybe call it `confirm` or `confirmMembership` instead, similar to `confirmBooked` in the `Donation` Domain model?

contexts/MembershipContext/src/UseCases/HandlePayPalNotification/HandlePayPalMembershipApplicationNotificationUseCase.php
```diff
@@ -0,0 +1,135 @@
...
+		if ( $this->transactionIsSubscriptionRelatedButNotAPayment( $request ) ) {
+			return true;
+		}
+
+		$this->logUnhandledStatus( $request );
+		return false;
+	}
+
+	private function handleRequestForMembershipApplication( PayPalNotificationRequest $request,
+															Application $application ): bool {
```
gbirke: I know this is copied 1:1 from the donation use case, but I just realized that some logging (probably on notice level) whenever we return `false` would be really helpful. Each `false` means we did not handle an IPN but have no record of it not being handled.

contexts/MembershipContext/src/DataAccess/DoctrineApplicationRepository.php
KaiNissen: Direct debits were first persisted with a neutral status and updated to confirmed once the SEPA checkbox form was sent.

contexts/MembershipContext/src/DataAccess/DoctrineApplicationRepository.php
KaiNissen: You sure? The constructor's parameter is nullable.

gbirke: Ok, using a `null` value for a constructor with default `null` will throw no type error, I've checked.

gbirke:
I've made a followup PR with some small changes that caught my eye.

contexts/MembershipContext/tests/Data/ValidMembershipApplication.php
```diff
@@ -171,7 +178,7 @@ private function newPayPalPayment() {
 		return new Payment(
 			self::PAYMENT_PERIOD_IN_MONTHS,
 			Euro::newFromFloat( self::PAYMENT_AMOUNT_IN_EURO ),
-			new PayPalPayment()
+			new PayPalPayment( ( new PayPalData() )->freeze() )
```
KaiNissen: Right, totally unnecessary. I changed it (and also the `MembershipApplicationBuilder`).

PR #727, opened 2016-10-26
consistent naming for bank account number field

KaiNissen:
fixes wmde/fundraising#1119


KaiNissen:
build failing due to issues with the paypal sandbox

PR #728, opened 2016-10-26
Log all Paypal handling code paths

gbirke:
Add paypal logger
Add log calls to all cases when the IPN was not handled successfully.

This for https://phabricator.wikimedia.org/T149172

### DIFF
--- a/contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
+++ b/contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
@@ -4,8 +4,6 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\DonationContext\UseCases\HandlePayPalPaymentNotification;
 
-use Psr\Log\LoggerInterface;
-use Psr\Log\LogLevel;
 use WMDE\Fundraising\Frontend\DonationContext\Authorization\DonationAuthorizer;
 use WMDE\Fundraising\Frontend\DonationContext\Domain\Model\Donation;
 use WMDE\Fundraising\Frontend\DonationContext\Domain\Model\DonationPayment;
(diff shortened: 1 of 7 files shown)


contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/PaypalNotificationResponse.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\DonationContext\UseCases\HandlePayPalPaymentNotification;
+
+class PaypalNotificationResponse
+{
```
JeroenDeDauw: Such bracket placement

contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/PaypalNotificationResponse.php
```diff
@@ -0,0 +1,44 @@
...
+	private $wasHandled;
+	private $notificationFailed;
+	private $handlingContext;
+
+	private function __construct( bool $notficationWasHandled, bool $isError, array $handlingContext = [] )
+	{
```
JeroenDeDauw: maw

contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
```diff
@@ -61,82 +61,59 @@ public function handleNotification( PayPalNotificationRequest $request ): bool {
...
-	private function logUnhandledNonPayment( PayPalNotificationRequest $request ) {
-		$logContext = [
-			'transaction_type' => $request->getTransactionType(),
-			'txn_id' => $request->getTransactionId()
-		];
-		$this->logger->info( 'Unhandled PayPal subscription notification: ' . $request->getTransactionType(), $logContext );
+	private function createUnhandledResponse( string $reason, PayPalNotificationRequest $request ): PaypalNotificationResponse {
+		return PaypalNotificationResponse::newUnhandledResponse( [
+			'message' => $reason,
+			'request' => $request->toArray()
```
JeroenDeDauw: I find it a bit odd that this part of the response model format is determined by the request model.

gbirke: Agree. The intention behind that was to check what PayPal sent in its request, but now that I've looked at it again I think it would be much better to get rid of `toArray` and the `request` field in the response and add the parameters directly from the HTTP request when logging in `PayPalNotificationHandler`.

JeroenDeDauw: Sounds good

PR #731, opened 2016-10-31
Refactor tracking

gbirke:
Now all request handlers can access the tracking code and the tracking source from `Request` attributes. This is more consistent, especially when `Subscription` will also store the tracking information.


JeroenDeDauw:
CR+2

JeroenDeDauw:
Btw, if I CR+2, feel free to self-merge once the CI is fluff

PR #733, opened 2016-11-01
pass tracking data across forms

KaiNissen:
Certain tracking data that is set in banners has to be passed through the donation form. This PR sets ground for it. For each of the tracking parameters an `<input type="hidden" />` must be added to the [form template](https://fundraising-cms.wikimedia.de/index.php?title=Web:Spendenseite-DefaultSkin/test/DonationForm_PersonalData&diff=601&oldid=596).


JeroenDeDauw:
Looks good to me. Minor follow up: https://github.com/wmde/FundraisingFrontend/pull/734

PR #739, opened 2016-11-03
Move Paypal logging to handler

gbirke:
The Paypal notification Response object no longer contains the
PayPalNotificationRequest object. Instead, the POST params from Paypal
are logged in the handler. This enables better debugging because we can
see what data was sent by Paypal.
The test of the logging contents have been moved from the use case to
the route test.

Removed status check from PayPalPaymentNotificationVerifier because
status is automatically and implicitly handled by the use case.

Added logging in case the PayPalPaymentNotificationVerifier throws an
Exception.

Removed wrapped PayPalPaymentNotificationVerifier from test.


JeroenDeDauw:
Reminds me I should finish up this LoggerSpy stuff...

JeroenDeDauw:
Would be cool if you could reply to https://github.com/wmde/PsrLogTestDoubles/pull/1

PR #740, opened 2016-11-04
Add explicit 404 presenter

gbirke:
The error message was not localizable and contained the HTTP request
method and the word "route" which can be considered jargon.

Also, every 404 error was logged with a stack trace.

This patch adds a localizable general message and does not log 404
errors to the app error log (the errors will still show up in the web
server logs).

This is the first improvement for https://phabricator.wikimedia.org/T147931

### DIFF
--- a/app/templates/404message.html.twig
+++ b/app/templates/404message.html.twig
@@ -1,2 +1,5 @@
-Could not load main content!
-{# TODO translate this and add a link to main page #}
+{% extends 'BaseLayout.twig' %}
+
+{% block main %}
+    {$ message|default( 'page_not_found' )|trans()|raw $}
+{% endblock %}
--- a/src/Presentation/Presenters/PageNotFoundPresenter.php
+++ b/src/Presentation/Presenters/PageNotFoundPresenter.php
@@ -1,0 +1,25 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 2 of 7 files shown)


src/Presentation/Presenters/PageNotFoundPresenter.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace WMDE\Fundraising\Frontend\Presentation\Presenters;
+
+use WMDE\Fundraising\Frontend\Presentation\TwigTemplate;
+
+/**
+ * @license GNU GPL v2+
+ * @author Gabriel Birke < gabriel.birke@wikimedia.de >
+ */
+class PageNotFoundPresenter {
```
JeroenDeDauw: This class seems a bit silly. How about just inlining this code until there is more than just a single method call ?

app/templates/404message.html.twig
```diff
@@ -1,2 +1,5 @@
-Could not load main content!
-{# TODO translate this and add a link to main page #}
\ No newline at end of file
+{% extends 'BaseLayout.twig' %}
+
+{% block main %}
+    {$ message|default( 'page_not_found' )|trans()|raw $}
```
JeroenDeDauw: Where is the cat? https://www.entropywins.wtf/you-need-such-a-cat

PR #741, opened 2016-11-08
set expiry of cookie created after donation

KaiNissen:
as defined in wmde/fundraising#1122, the cookie duration has to be set to 180 days


JeroenDeDauw:
Build failure looks unrelated. I've restarted the build

gbirke:
The failure occurs because the test domain `alltheguys.com` has gone away and the request validation fails. 

Rebase after merging #742

JeroenDeDauw:
Ugggh why is there a "rebase and merge" button but not a "rebase" one? >_>

PR #746, opened 2016-11-10
Use PSR Log Test Doubles

JeroenDeDauw:


JeroenDeDauw:
Mawrrrggggg @gbirke @KaiNissen

PR #747, opened 2016-11-11
Server-side tracking for mobile paypal donations

gbirke:
Mobile paypal donations are anonymous and go straight to donation/add
route which redirects to paypal. That means we don't have Piwik tracking
data which is needed by the backend. That means we have to simulate the
piwik tracking on the server.

ServerSideTracker, its implementations and PageViewTracker were taken
from the old fundraising app.

This is for https://phabricator.wikimedia.org/T150525

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -63,8 +63,7 @@ class AddDonationHandler {
 		$campaign = $trackingCode[0];
 		$keyword = $trackingCode[1] ?? '';
 
-		$this->ffFactory->getServerSideTracker()->setIp( $request->getClientIp() );
-		$this->ffFactory->newPageViewTracker()->trackPaypalRedirection( $campaign, $keyword );
+		$this->ffFactory->getPageViewTracker()->trackPaypalRedirection( $campaign, $keyword, $request->getClientIp() );
 	}
 
 	private function newHttpResponse( AddDonationResponse $responseModel ): Response {
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -49,6 +49,8 @@ class AddDonationHandler {
 			);
 		}
 
+		$this->sendTrackingDataIfNeeded( $request, $responseModel );
+
 		return $this->newHttpResponse( $responseModel );
 	}
 
@@ -180,6 +182,17 @@ class AddDonationHandler {
 		$tracking->setTotalImpressionCount( intval( $request->get( 'impCount', 0 ) ) );
 
 		return $tracking;
+	}
+
+	private function sendTrackingDataIfNeeded( Request $request, AddDonationResponse $responseModel ) {
+		if ( $request->get( 'mbt', '' ) !== '1' || $responseModel->getDonation()->getPaymentType() !== PaymentType::PAYPAL ) {
+			return;
+		}
+		$trackingCode = explode( '/', $request->attributes->get( 'trackingCode' ) );
+		$campaign = $trackingCode[0];
+		$keyword = $trackingCode[1] ?? '';
+		$this->ffFactory->getServerSideTracker()->setIp( $request->getClientIp() );
+		$this->ffFactory->newPageViewTracker()->trackPaypalRedirection( $campaign, $keyword );
 	}
 
 }
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -416,10 +416,8 @@ class FunFunFactory {
 			return new VoidCache();
 		};
 
-		$this->pimple['server_side_tracker'] = function () {
-			return new PiwikServerSideTracker(
-				new \PiwikTracker( $this->config['piwik']['site_id'], 'https://' . $this->config['piwik']['baseUrl'] )
-			);
+		$this->pimple['page_view_tracker'] = function () {
+			return new PageViewTracker( $this->newServerSideTracker(), $this->config['piwik']['siteUrlBase'] );
 		};
 
 		return $pimple;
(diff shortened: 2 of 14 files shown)


gbirke:
@JeroenDeDauw I have a feeling that this code might need to be refactored, see my comment in `AddDonationRouteTest.php`. Thoughts?

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -657,4 +658,72 @@ public function testWhenInitiallyIntendedPaymentOptionsDifferFromActual_itIsRefl
...
+			$this->assertCount( 0, $trackerSpy->getPageViews(), 'The page view should not have been tracked' );
+		} );
+	}
+
+	// strictly speaking, we should also write the tests
+	// testWhenMobileTrackingIsRequested_piwikTrackerIsNotCalledForDirectDebitPayment
+	// testWhenMobileTrackingIsRequested_piwikTrackerIsNotCalledForBankTransferPayment
+	// but somehow that feels ugly, especially when thinking about add more payment methods
+	// Maybe there should be a payment-specific ServersideTrackingStrategy class/factory that can be tested outside of
+	// the edge-to-edge routing test? Other ideas?
```
JeroenDeDauw: I've not looked at all the code yet, but my first thought when reading this is to just have a 

`testWhenMobileTrackingIsRequested_piwikTrackerIsNotCalledForNonPaypalPayment`

app/RouteHandlers/AddDonationHandler.php
```diff
@@ -182,4 +184,15 @@ private function newTrackingInfoFromRequest( Request $request ): DonationTrackin
...
 
+	private function sendTrackingDataIfNeeded( Request $request, AddDonationResponse $responseModel ) {
+		if ( $request->get( 'mbt', '' ) !== '1' || $responseModel->getDonation()->getPaymentType() !== PaymentType::PAYPAL ) {
+			return;
+		}
+		$trackingCode = explode( '/', $request->attributes->get( 'trackingCode' ) );
+		$campaign = $trackingCode[0];
+		$keyword = $trackingCode[1] ?? '';
+		$this->ffFactory->getServerSideTracker()->setIp( $request->getClientIp() );
+		$this->ffFactory->newPageViewTracker()->trackPaypalRedirection( $campaign, $keyword );
```
JeroenDeDauw: This is really ick. How about giving the IP to the tracking method? As far as I can tell there is no reason to have this mutability, much less stuffing the state in the global factory.

src/Factories/FunFunFactory.php
```diff
@@ -1283,4 +1292,18 @@ public function newPageNotFoundHTMLPresenter(): PageNotFoundPresenter {
 		return new PageNotFoundPresenter( $this->getLayoutTemplate( '404message.html.twig' ) );
 	}
 
+	public function setServerSideTracker( ServerSideTracker $tracker ) {
+		$this->pimple['server_side_tracker'] = function () use ( $tracker )  {
+			return $tracker;
+		};
+	}
```
JeroenDeDauw: heh? :)

tests/Fixtures/ServerSideTrackerSpy.php
```diff
@@ -0,0 +1,44 @@
...
+	 */
+	public function getPageViews() {
+		return $this->pageViews;
+	}
+
+	public function setIp( string $ip ) {
+		$this->ips[] = $ip;
+	}
+
+	public function getIps(): array {
```
JeroenDeDauw: I'd prefer something like getSetIps, or getSetIpCalls. I'll leave this be for now, cause altering this will conflict with modifying or removing the setIp stuff

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -657,4 +658,72 @@ public function testWhenInitiallyIntendedPaymentOptionsDifferFromActual_itIsRefl
 		$this->assertContains( '12/0', $responseContent );
 	}
 
+	public function testWhenMobileTrackingIsRequested_piwikTrackerIsCalledForPaypalPayment() {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ) {
+			$factory->setNullMessenger();
```
JeroenDeDauw: Right... we really should get to removing the need for this :)

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
JeroenDeDauw: Looked at the code now.

There are at least two things to test here:

* The if statement in `AddDonationHandler `that bails on non-paypal
* That `AddDonationHandler` invokes things (including the `PageViewTracker`) correctly

For the later, doing a single invocation should suffice, and is this done by the first test here. For the former we just need another test that verifies the `PageViewTracker` does not get called when non-paypal. (If it got more complicated than just that if statement, it'd be unfortunate to have it in the route handler.) That's almost what you did with the second test (it's making an assertion about `ServerSideTracker`, which I think it should not know about.)

So it seems like the spy is on the wrong level, and there is no advantage to having a `PageViewTracker` containing a `ServerSideTracker`, unless I'm missing something of course. But it looks like these can be merged.

src/Factories/FunFunFactory.php
gbirke: Technically, assigning a service to pimple, it should be a creation function that is accessed when the service is requested. Directly assigning the finished object also works, but I think assigning a new creation function might be more "in the spirit" of this DI container. I'll ask on StackOverflow about this.

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
gbirke: In the old application they were separated and I think it makes sense from a SRP perspective: The `ServerSideTracker` implementors abstract the tracking framework, the `PageViewTracker` builds the tracking data/url, allowing for separate testing, wich is a bit more high level. But feel free to merge the two in a followup after I've revised the code.

src/Factories/FunFunFactory.php
gbirke: Question is here: http://stackoverflow.com/questions/40590533/most-semantic-way-to-replace-services-in-pimple-container

JeroenDeDauw: Thing is there is nothing to create here. The variable already exists anyway.

PR #750, opened 2016-11-14
Remove NullMessenger in Tests

gbirke:
Followup to #749


tests/TestEnvironment.php
```diff
@@ -27,7 +27,6 @@ public static function newInstance( array $config = [] ) {
 
 		$installer->install();
 
-		$instance->factory->setNullMessenger();
```
JeroenDeDauw: :D

PR #754, opened 2016-11-15
Allow setting address type for new donations

gbirke:
addressType can now be passed in the request from the banner and will be
selected on the 2nd form page.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -406,7 +406,8 @@ $app->get( 'donation/new', function ( Request $request ) use ( $ffFactory ) {
 			$request->get( 'zahlweise', '' ),
 			intval( $request->get( 'periode', 0 ) ),
 			$validationResult->isSuccessful(),
-			$trackingInfo
+			$trackingInfo,
+			$request->get( 'addressType', 'person' )
 		)
 	);
 } )->method( 'POST|GET' );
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I'm not happy with using the name "address type" for this. I was confoosed due to that name for quite some time, until we got rid of it in favor of better modeling.

app/routes.php
```diff
@@ -406,7 +406,8 @@ function( Application $app, Request $request ) use ( $ffFactory ) {
 			$request->get( 'zahlweise', '' ),
 			intval( $request->get( 'periode', 0 ) ),
 			$validationResult->isSuccessful(),
-			$trackingInfo
+			$trackingInfo,
+			$request->get( 'addressType', DonorName::PERSON_PRIVATE )
```
JeroenDeDauw: Not sure it's appropriate to use this constant here. I'd need to double check which I can't do ATM, but in general domain stuff should not be used outside the domain and the UCs around it, ie in routes or presenters. Changing the value of the constant should not require matching changes in the js (or in the databse).

There are a few other such cases scattered around... not a huge issue, yet still subtle corruptions of the architecture

JeroenDeDauw:
This looks reasonable to me. Two things I don't like and would be nice if they could be changed. If the name change is easy (ie does not require also changing names in a pile of other places), it's probably good to do that right away.

gbirke:
The problem as I see it is that the `addressType` parameter from the form is used for two things in our model:

1. The type of `DonorName`
2. If the `donor` property of `Donation` is set to null or not.

Changing the name of the form field/parameter would be very tedious and should not be done during the campaign IMHO, if it should be done at all. But I agree that the constant, that's specifically for the name, should not be used for the default value. Will change that.

PR #762, opened 2016-11-23
Extend max bank name length for membership applications

gbirke:
Followup to #760


gbirke:
Looked at wrong master, change is already in.

PR #765, opened 2016-11-25
Use confirmation code directly.

gbirke:
Use confirmation code directly, (instead of the removed Hex confirmation code), fix success page templates.

### DIFF
--- a/app/templates/ConfirmSubscription.html.twig
+++ b/app/templates/ConfirmSubscription.html.twig
@@ -4,8 +4,8 @@
     {% if error_message %}
         {# TODO ask Kai if we can make the " Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut." string on
             the error page optional #}
-        {% include 'ErrorPage.twig' with { 'message': error_message|trans({}, 'validations') } %}
+        {$ include( 'ErrorPage.twig', { 'message': error_message|trans( {}, 'validations' ) } ) $}
     {% else %}
-        {% include 'Subscription_Confirmed.twig' %}
+        {$ include( 'Subscription_Confirmed.twig' ) $}
     {% endif %}
 {% endblock %}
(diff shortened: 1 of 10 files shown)


JeroenDeDauw:
+1 so far

gbirke:
No longer a WIP, ping @KaiNissen

app/templates/ConfirmSubscription.html.twig
```diff
@@ -4,8 +4,8 @@
     {% if error_message %}
         {# TODO ask Kai if we can make the " Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut." string on
             the error page optional #}
-        {% include 'ErrorPage.twig' with { 'message': error_message|trans({}, 'validations') } %}
+        {$ include( 'ErrorPage.twig', { 'message': error_message|trans( {}, 'validations' ) } ) $}
```
KaiNissen: Actually, both the tag and the function should behave the same.

app/templates/ConfirmSubscription.html.twig
```diff
@@ -4,8 +4,8 @@
     {% if error_message %}
         {# TODO ask Kai if we can make the " Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut." string on
```
KaiNissen: I found an easter-egg!

app/templates/ErrorPage.twig
```diff
@@ -0,0 +1,20 @@
+<section class="container clearfix">
```
KaiNissen: We have a similar file (that goes by the same name) in the content management wiki. Did you introduce this one to not display the "try again later" notice?

gbirke: See #769

app/templates/ConfirmSubscription.html.twig
gbirke: See #769

PR #766, opened 2016-11-28
allow postal codes between 1 and 6 digits if not defined otherwise

KaiNissen:
The pattern we've been using is actually invalid.
(see wmde/fundraising#1170)


JeroenDeDauw:
No tests for this functionality #grump

PR #767, opened 2016-11-28
use silex' twig instance

KaiNissen:
The way we created a `Twig_Environment` instance blocks us from using the twig bridge services provided by `TwigServiceProvider`. This change injects Silex' twig instance (and thus the twig bridge) into pimple.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -24,6 +24,7 @@ $app = new Application();
 $app->register( new SessionServiceProvider() );
 $app->register( new RoutingServiceProvider() );
 $app->register( new TwigServiceProvider() );
+$ffFactory->setTwigEnvironment( $app['twig'] );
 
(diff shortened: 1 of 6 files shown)


app/bootstrap.php
```diff
@@ -24,6 +24,7 @@
 $app->register( new SessionServiceProvider() );
 $app->register( new RoutingServiceProvider() );
 $app->register( new TwigServiceProvider() );
+$ffFactory->setTwigEnvironment( $app['twig'] );
```
JeroenDeDauw: This is a slightly surprising location to me. Before this file just set up $app, now it also has a side effect on the factory. 

We already have such side effects based on $app in index.dev.php, so the same could be done for this. However you'd need to put it in both entry points, and possibly in the test environment thing that includes this file as well.

PR #769, opened 2016-11-28
Improve error handling

gbirke:
Delete ErrorPage, a similar one already existed in the CMS.
Chnage the page style.

Use different template in ConfirmSubscription because the default error
page has an additional message that does not apply.


JeroenDeDauw:
+1

PR #772, opened 2016-11-29
Remove name validation for subscriptions

gbirke:
Only email address is a required field.


gbirke:
This is for https://github.com/wmde/fundraising/issues/1110

PR #773, opened 2016-11-30
Allow subscription via JSONP

gbirke:
This is for https://github.com/wmde/fundraising/issues/1110

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -25,9 +25,21 @@ $app->register( new SessionServiceProvider() );
 $app->register( new RoutingServiceProvider() );
 $app->register( new TwigServiceProvider() );
 
+$app['request_is_jsonp'] = $app->protect( function ( Request $request ) {
+	$jsonpContentTypes = [
+		'application/json',
+		'application/json; charset=utf-8',
+		'application/javascript',
+	];
+	return $request->get( 'jsonp_callback' ) !== null &&
+		$request->getMethod() === 'GET' &&
+		count( array_intersect( $jsonpContentTypes, $request->getAcceptableContentTypes() ) ) > 0;
+} );
+
 $app->before(
 	function ( Request $request, Application $app ) {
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
Is all this bootstrap stuff really needed? `list-comments.json` as JSONP support with

```php
if ( $request->query->get( 'f' ) ) {
	$response->setCallback( $request->query->get( 'f' ) );
}
```

gbirke:
Yeah, it was a bit premature to put this functionality in bootstrap instead of implementing it in the handler. I just thought "it would be neat if all our JSON response would also support JSONP". Feel free to rip it out of bootstrap and move it into the handler.

app/bootstrap.php
```diff
@@ -25,9 +25,21 @@
...
+	];
+	return $request->get( 'jsonp_callback' ) !== null &&
+		$request->getMethod() === 'GET' &&
+		count( array_intersect( $jsonpContentTypes, $request->getAcceptableContentTypes() ) ) > 0;
+} );
+
 $app->before(
 	function ( Request $request, Application $app ) {
 		$app['request_stack.is_json'] = in_array( 'application/json', $request->getAcceptableContentTypes() );
+		$app['request_stack.is_jsonp'] = $app['request_is_jsonp']( $request );
```
JeroenDeDauw: Is `'application/json; charset=utf-8'` really only for JSONP and not for JSON?

JeroenDeDauw:
Sounds reasonable to get JSONP for all our JSON stuffs.

This means the comment list thing has double JSONP support across the sky. OMG. What does it mean?

gbirke:
It only means that you can call it with two different callback parameters: When using `f` (the param we used so far), the code path in the handler will be used, if using `callback` the code path in bootstrap should be used (could not get that to work yet with curl, will try again). If you combine `f` and `callback`, the `callback` param should "win", but that needs to be tested once I've got to base `callback` case to work.

PR #775, opened 2016-12-01
Create and use Doctrine proxy classes

gbirke:
Create them as part of the release process.
Use them in the entity manager configuration.

This is for https://github.com/wmde/fundraising/issues/1187


gbirke:
I've not thoroughly tested if this really solves https://github.com/wmde/fundraising/issues/1187 and does not break anything, so this should be considered a WIP.

JeroenDeDauw:
I don't understand how this is a production specific thing

gbirke:
It also affects the local environment. Strangely enough only the subscription entity.
But for production, or rather deployment, we need to ensure the proxies are generated.

PR #776, opened 2016-12-05
Revert "Create and use Doctrine proxy classes"

KaiNissen:
Sorry, I merged this prematurely. Does not seem to work.

Reverts wmde/FundraisingFrontend#775


JeroenDeDauw:
Then submit a revert of this revert that can be fixed?

PR #777, opened 2016-12-05
pass proxy dir to StoreFactory

KaiNissen:
This patch requires wmde/FundraisingStore#120 to be merged first. It fixes wmde/fundraising#1187.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 		"twig/twig": "~1.23",
 		"pimple/pimple": "~3.0",
 
-		"wmde/fundraising-store": "~4.0",
+		"wmde/fundraising-store": "~4.1",
 		"wmde/euro": "~1.0",
 		"wmde/page-retriever": "~1.0",
 
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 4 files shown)


JeroenDeDauw:
+1

gbirke:
@JeroenDeDauw The underlying issue is that autogenerated proxy classes are generated in `/tmp` where they are deleted regularly by the OS. The solution is twofold - move path for proxy classes to the `var` dir and autogenerate the classes on deployment.

Only the first part of the solution (the path) can be unit tested, but I don't know if objects created be the `FunFunFactory` should be tested and where such a test would live.

The second part would be a test for the Ansible deployment playbook and we don't do that at the moment.

JeroenDeDauw:
Hmm sure

gbirke:
A thought - does the directory need to exist or will it be created automatically? Maybe the unit test can force a proxy object to be created while the directory is absent.

gbirke:
This is very hard to test in an automated manner - when manually testing the subscription confirmation manually, the test fails if the classes don't exist.`ConfirmSubscriptionRouteTest` runs without a hitch, regardless of the existence of the proxy classes. 

I've tested the Ansible patch manually to make sure the proxy classes are generated and the confirmation route works. It does, so this can be merged.

gbirke:
One remark about the deployment: After merging this, the checkout of this repo on the deployment server must be updated. I've set it to this branch temporarily to test this PR, but it should be reverted to master.

KaiNissen:
checked out master on the deployment server

PR #778, opened 2016-12-06
Fix storage of personal title

gbirke:
Personal titles were not sent when the donation/membership forms were submitted.

This fixes https://github.com/wmde/fundraising/issues/1198

### DIFF
--- a/web/res/js/donationForm.js
+++ b/web/res/js/donationForm.js
@@ -25,7 +25,7 @@ $( function () {
 			} ),
 			WMDE.Components.createRadioComponent( store, $( '.address-type-select' ), 'addressType' ),
 			WMDE.Components.createRadioComponent( store, $( '.salutation' ), 'salutation' ),
-			WMDE.Components.createRadioComponent( store, $( '.personal-title' ), 'title' ),
+			WMDE.Components.createSelectMenuComponent( store, $( '.personal-title' ), 'title' ),
 			WMDE.Components.createValidatingTextComponent( store, $( '#first-name' ), 'firstName' ),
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Where is teh test!

gbirke:
This could only be reproduced with a browser test.

web/res/js/donationForm.js
```diff
@@ -25,7 +25,7 @@ $( function () {
 			} ),
 			WMDE.Components.createRadioComponent( store, $( '.address-type-select' ), 'addressType' ),
 			WMDE.Components.createRadioComponent( store, $( '.salutation' ), 'salutation' ),
-			WMDE.Components.createRadioComponent( store, $( '.personal-title' ), 'title' ),
+			WMDE.Components.createSelectMenuComponent( store, $( '.personal-title' ), 'title' ),
```
JeroenDeDauw: What was the actual issue? I suppose `createRadioComponent` is not broken since it is still used above. Did this mismatch with code elsewhere?

JeroenDeDauw:
ok

ok

web/res/js/donationForm.js
gbirke: The issue was that we use the "selectmenu" jQuery UI component for all `select` fields. While the `RadioComponent` works for normal select fields, for the jquery ui selectmenu it doesn't, because the `change` event is not triggered and this the store (that stores all client side state) is not updated. This lead to a reset of the title field whenever a different field was changed (because all field values are updated with the values from the store whenever a value changes). The `SelectMenuComponent` listens to the `selectmenuchange` event and updates the store accordingly.

JeroenDeDauw:
+1

PR #780, opened 2016-12-08
Revert global JSONP handling

gbirke:
Browsers request JavaScript (and thus JSONP content) with "Accept: */*".
This means it's not possible to globally detect JSONP requests. As a
consequence, the logic for handling JSONP must reside with individual
route handlers that know better if a JSONP response is desired or not.

This partially reverts #773.


JeroenDeDauw:
+1

PR #781, opened 2016-12-08
Remove country-specific postcode validation

gbirke:
Since the country selection comes after the validation of the postcode,
valid 4-digit Austrian postcodes were getting marked as invalid. Quick
fix is to allow 4-6 digit postcodes in the templates and remove
country-specific validation.


gbirke:
This is for https://github.com/wmde/fundraising/issues/1200

JeroenDeDauw:
+1

KaiNissen:
I just deployed template changes as well.

PR #782, opened 2016-12-12
Check SOA record in DNS checks

gbirke:
Not all mail domains have a MX or A record, some are zones with only a
'SOA' record where the MX and A records are delegated to another
subdomain.

Example: web.de-mail.de

This fixes https://github.com/wmde/fundraising/issues/1207


JeroenDeDauw:
Not possible to test this?

gbirke:
A test would depend on a working network environment and a specific DNS entry configuration for the domains that will be queried. `InternetDomainNameValidator` specifically exists to decouple the tests from the network environment. While we could introduce an integration test for `InternetDomainNameValidator`, the only approach for this test to always work would be to run it in its own VM/container with a defined DNS setup.

JeroenDeDauw:
Did not know DNS config was needed. A VM does not seem worth it. If that was not needed then a test in the system folder would have been cool

PR #784, opened 2016-12-12
Use more specific locale in Paypal URL

gbirke:
On the official list of supported locales, our shortened locale "de"
does not show up and there is no mention if a shortened locale will be
processed by PayPal:
https://developer.paypal.com/docs/classic/api/locale_codes/

So now the default locale is one of the offical Paypal language codes,
'de_DE'.


gbirke:
This might avoid a behavior like described in https://github.com/wmde/fundraising/issues/1210

PR #786, opened 2016-12-13
switch to PageRetriever that fetches raw wikitext with expanded templates

KaiNissen:
This PR depends on wmde/PageRetriever#4.


JeroenDeDauw:
+1

PR #791, opened 2016-12-19
introduce alternative sender name

KaiNissen:
The application sends e-mails on behalf of two organizations. "Wikimedia Deutschland e. V." and "Wikimedia Fördergesellschaft mbH". The latter is to be used in almost every case apart from membership application related stuff.


app/config/config.test.json
```diff
@@ -32,6 +32,7 @@
 	"membership-application-timeframe-limit": "PT10M",
 	"operator-email": "email@operatorsownmailserver.com",
 	"operator-displayname": "Friendly Operator",
+	"operator-displayname-alternative": "Alternative Operator",
```
JeroenDeDauw: Well now I know why I got 50 emails or so from a "	Alternative Operator" in my inbox...

JeroenDeDauw:
Where is the name "alternative" coming from? I don't understand what this means. Is this for memberships and not for donations?

KaiNissen:
I updated the description to explain this.

app/config/config.test.json
KaiNissen: Wow, sorry. The method `FunFunFactory::setNullMessenger()` wasn't setting the alternative one. All the membership related tests were sending e-mails to your account, because it's in `ValidMembershipApplication`.

JeroenDeDauw:
Would it than make sense to rename the mailers from `mailer` and `alternativeMailer` to... well, whatever you call those two things.

KaiNissen:
@JeroenDeDauw I wanted it to be more general, but I agree, that this was not self-explanatory. I created wmde/fundraising#1230 to follow up on this.

JeroenDeDauw:
+1

PR #792, opened 2016-12-22
Reorganize email configuration

gbirke:
Membership-related emails (or rather, emails pertaining to the main
organization (WMDE) can now have a different sender/receiver than emails
that pertain to the sub-organization (Wikimedia Fördergesellschaft).

In the future, the `contact_info` configuration could be used for
different use cases (contact, subscription, etc), but at the moment
that's not needed and  would be too much duplication in the
configuration.

This is a follow-up to #791 and solves https://github.com/wmde/fundraising/issues/1230


KaiNissen:
+1, created a follow-up of minor stuff in #793

PR #794, opened 2016-12-28
output as expected by client side response handling

KaiNissen:
The output of the `validate-fee` route does not match what is expected by the client-side handler.

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -157,7 +157,7 @@ $app->post(
 			return $app->json( [ 'status' => 'OK' ] );
 		} else {
 			$errors = $result->getViolations();
-			return $app->json( [ 'status' => 'ERR', 'message' => implode( "\n", $errors ) ] );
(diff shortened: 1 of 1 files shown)


KaiNissen:
no reviewer any time soon, so self-merging this.

gbirke:
Looks good, +2

JeroenDeDauw:
such a no test

KaiNissen:
@JeroenDeDauw I think this is better covered in a browser test, since it's interaction between client-side and server-side code.

JeroenDeDauw:
It'd be nice to have such a test as well. However you still get the usual value from a more local test: you know that the output of the code is as you specified, you have such a specification, and when you refactor the code, you'll notice (unintentional) changes to the behavior.

no-testing-required-it-will-work

PS: found this for usage when the next AB test is requested from the management side:

asian-father-growth-giant

PR #796, opened 2017-01-03
Bank data validation

KaiNissen:
The bank data validation did not work well with IBANs that do not start with "DE".

- Bank data is not considered invalid when BIC is missing in the validation result.
- BIC is not overwritten with an empty value anymore.
- Bank name is not a mandatory field anymore.
- BIC is also validated by regex now.

This is a fix for wmde/fundraising#1247 and wmde/fundraising#1254

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -145,7 +145,7 @@ class AddDonationHandler {
 			->setBankCode( $request->get( 'blz', '' ) )
 			->setBankName( $request->get( 'bankname', '' ) );
 
-		if ( $bankData->hasIban() && !$bankData->hasCompleteLegacyBankData() ) {
+		if ( $bankData->hasIban() && !$bankData->hasBic() && !$bankData->hasCompleteLegacyBankData() ) {
 			$bankData = $this->newBankDataFromIban( $bankData->getIban() );
 		}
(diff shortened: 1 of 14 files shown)


app/RouteHandlers/AddDonationHandler.php
```diff
@@ -145,7 +145,7 @@ private function getBankDataFromRequest( Request $request ): BankData {
 			->setBankCode( $request->get( 'blz', '' ) )
 			->setBankName( $request->get( 'bankname', '' ) );
 
-		if ( $bankData->hasIban() && !$bankData->hasCompleteLegacyBankData() ) {
+		if ( $bankData->hasIban() && !$bankData->hasBic() && !$bankData->hasCompleteLegacyBankData() ) {
```
JeroenDeDauw: This logic is misplaced... just saying, don't think this should be fixed in this PR

contexts/PaymentContext/src/Domain/Model/BankData.php
```diff
@@ -74,6 +74,10 @@ public function hasIban(): bool {
 		return !empty( $this->getIban()->toString() );
 	}
 
+	public function hasBic(): bool {
+		return !empty( $this->getBic() );
```
JeroenDeDauw: `!== ''` is clearer. `empty` is also weird:

    php > var_dump(empty('0'));
    php shell code:1:
    bool(true)

PR #797, opened 2017-01-04
Email address blacklist

KaiNissen:
This PR depends on wmde/FundraisingStore#121 (and yet another release of the store).

It resolves wmde/fundraising#1106.

### DIFF
--- a/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
+++ b/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
@@ -64,6 +64,10 @@ class AddDonationUseCase {
 			$donation->markForModeration();
 		}
 
+		if ( $this->policyValidator->isAutoDeleted( $donationRequest ) ) {
+			$donation->markAsDeleted();
+		}
+
 		// TODO: handle exceptions
 		$this->donationRepository->storeDonation( $donation );
 
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -62,6 +62,7 @@
 			"badwords": ""
 		}
 	},
+	"email-address-blacklist": [],
 	"subscription-interval": "PT30M",
 	"locale": "de_DE",
 	"referrer-generalization": {
--- a/app/config/schema.json
+++ b/app/config/schema.json
@@ -305,6 +305,11 @@
         "comment"
       ]
     },
+    "email-address-blacklist": {
+      "type": "array",
+      "title": "List of blocked e-mail addresses",
+      "description": "List of e-mail addresses that automatically mark data sets as deleted"
+    },
     "locale": {
       "type": "string",
       "minLength": 2,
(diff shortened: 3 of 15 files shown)


gbirke:
Look fine to me in general, two questions: 

1. Should we really block individual mail addresses or should we rather use a regex to be able to block domains?
2. Shouldn't we communicate the fact the donation/membership was autodeleted to the user? This PR only changes the validators and use cases, not the presenters. Or is this done in a template?

KaiNissen:
Hm, I can't really think of a case in which we would want to block domains.

The second question might be answered by @TillWMDE. In my opinion, it's good to show a confirmation page to the user rather than challenge them to try again with a different e-mail address.

TillWMDE:
1. neither can I think of a domain which is blockable.
2. why don´t we give ´em an error message like "Bitte verwenden Sie nicht die e-Mail Adresse spenden@wikimedia.de"?

KaiNissen:
Because people using our e-mail address are most probably not intending to donate. If we display an error message, they will think of another (bogus) e-mail address that is not blocked. So we actually help circumventing our "spam filter".

I'm also remembering the case of a legal guardian asking us to block the e-mail address of their ward. In this case it wouldn't have been helpful to tell them to simply use a different address.

TillWMDE:
well, let´s keep it simple. Just put them on moderated and delete it afterwards. We might need a cron job for deletion of moderated emails anyway

contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -64,6 +64,10 @@ public function addDonation( AddDonationRequest $donationRequest ): AddDonationR
 			$donation->markForModeration();
 		}
 
+		if ( $this->policyValidator->isAutoDeleted( $donationRequest ) ) {
+			$donation->markAsDeleted();
```
gbirke: According to the last comment by @TillWMDE, the donation should not only be deleted but also marked for moderation, right?

TillWMDE:
I would propose to just put it on moderation at first. We might need to think about a cron job to delete all moderated donations with certain criteria. See also #880. Does such a cron job make sense?

contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
KaiNissen: For donations we do not use bits to reflect the status. Marking for moderation and subsequently marking as deleted would result in `status` being `D`.

Although @TillWMDE said to keep it simple, I'm slightly confused about what should actually happen here.

gbirke: You're totally right. What I said only makes sense for memberships, for donations we can't track moderation and deletion separately.

KaiNissen:
I don't strongly oppose setting the status to "moderated", but I actually don't see the point. Especially, if we automatically run a script that changes the status to "deleted" afterwards. I can't think of a case in which we would ever approve a donation data set that used a blacklisted e-mail address.

KaiNissen:
rebased

KaiNissen:
@gbirke I added regex support to block domains

JeroenDeDauw:
+1 to the latest commit

PR #798, opened 2017-01-05
Improve bank data conversion logic

gbirke:
To make the intention of the code in AddDonationHandler more explicit,
introduce the new `hasCompleteBankData` method in `BankData`.

Also, use string comparison instead of checking for `empty` because
empty('0') is true.

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -145,13 +145,15 @@ class AddDonationHandler {
 			->setBankCode( $request->get( 'blz', '' ) )
 			->setBankName( $request->get( 'bankname', '' ) );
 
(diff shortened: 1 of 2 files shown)


app/RouteHandlers/AddDonationHandler.php
```diff
@@ -145,13 +145,15 @@ private function getBankDataFromRequest( Request $request ): BankData {
 			->setBankCode( $request->get( 'blz', '' ) )
 			->setBankName( $request->get( 'bankname', '' ) );
 
-		if ( $bankData->hasIban() && !$bankData->hasBic() && !$bankData->hasCompleteLegacyBankData() ) {
-			$bankData = $this->newBankDataFromIban( $bankData->getIban() );
+		if ( $bankData->hasCompleteBankData() ) {
```
JeroenDeDauw: How about `isComplete`?

PR #799, opened 2017-01-06
Add special case for PayPal currency check

gbirke:
The PayPal notification
`recurring_payment_suspended_due_to_max_failed_payment` has no
`mc_currency` field which causes an unhandled exception instead of
being passed to the use case (where it is dropped). This can lead to
Paypal re-trying the notification and/or disabling notifications.

### DIFF
--- a/tests/Unit/Infrastructure/PayPalPaymentNotificationVerifierTest.php
+++ b/tests/Unit/Infrastructure/PayPalPaymentNotificationVerifierTest.php
@@ -154,7 +154,7 @@ class PayPalPaymentNotificationVerifierTest extends \PHPUnit_Framework_TestCase 
 		return $this->newClient( $body );
 	}
 
-	private function newClient( Stream $body, array $expectedParams=null ): Client {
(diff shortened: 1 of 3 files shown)


tests/Unit/Infrastructure/PayPalPaymentNotificationVerifierTest.php
```diff
@@ -113,7 +154,7 @@ private function newFailingClient(): Client {
 		return $this->newClient( $body );
 	}
 
-	private function newClient( Stream $body ): Client {
+	private function newClient( Stream $body, array $expectedParams=null ): Client {
```
JeroenDeDauw: ` = `

JeroenDeDauw:
Ping @KaiNissen

gbirke:
In don't know why this PR causes Scrutinizer to fail maybe you can merge it none the less, @JeroenDeDauw?

JeroenDeDauw:
My review and not merging here had nothing to do with Scrutinizer

JeroenDeDauw:
#LeeroyJenkins!

PR #804, opened 2017-01-12
Refactor validation dispatchers

gbirke:
Wrap each validator in a validator-specific `ValidationDispatcher` instance.


app/js/tests/validation_dispatchers/test_address_validation_dispatcher.js
```diff
@@ -0,0 +1,96 @@
...
+		testStore = { dispatch: sinon.stub() },
+		dispatcher = createAddressValidationDispatcher(
+			validator,
+			initialData
+		);
+
+	dispatcher.dispatchIfChanged( testData, testStore );
+
+	t.ok( validator.validate.calledOnce, 'validation function is called once' );
+	t.ok( validator.validate.calledWith( expectedData ), 'validation function is called with selected fields' );
```
JeroenDeDauw: I'm not a fan of having a var for everything and putting those all at the start of the scope. That results in more state, mutability and distance between related stuff. For instance for this `expectedData`, if you inline it, you have none of those downsides.

app/js/tests/validation_dispatchers/test_address_validation_dispatcher.js
```diff
@@ -0,0 +1,96 @@
...
+		validator = { validate: sinon.stub().returns( successResult ) },
+		testStore = { dispatch: sinon.spy() },
+		dispatcher = createAddressValidationDispatcher(
+			validator,
+			initialData
+		);
+
+	dispatcher.dispatchIfChanged( testData, testStore );
+
+	t.ok( testStore.dispatch.calledWith( Actions.newFinishAddressValidationAction( successResult ) ) );
```
JeroenDeDauw: I was expecting the whole value to be specified in the test here. Won't that also catch errors in `newFinishAddressValidationAction`? Based on our conversation I figured this is not tested elsewhere. (I'm not sure this should be done here or would actually be helpful if it was)

gbirke: I'd agree with you in other cases. In this case, actions are basically value objects and action creation functions like `newFinishAddressValidationAction` are constructors of value objects and those aren't tested, right?

app/js/tests/validation_dispatchers/test_address_validation_dispatcher.js
gbirke: I agree for `expectedData` and have put up PR #806

app/js/tests/validation_dispatchers/test_address_validation_dispatcher.js
JeroenDeDauw: I was thinking about catching mismatches between which fields those value objects have and what the code using them expects.

PR #805, opened 2017-01-13
Fix errors in asynchronous validation on form submit

gbirke:
Make donation and membership validations wait for asynchronous validators to finish.

This fixes https://phabricator.wikimedia.org/T155016

Based on #804

### DIFF
--- a/app/js/lib/reducers/async_requests.js
+++ b/app/js/lib/reducers/async_requests.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = function ( state, action ) {
-	if ( typeof state === "undefined" ) {
+	if ( typeof state === 'undefined' ) {
 		state = { isValidating: false, runningValidations: 0 };
 	}
 	switch ( action.type ) {
--- a/app/js/lib/reducers/async_requests.js
+++ b/app/js/lib/reducers/async_requests.js
@@ -1,0 +1,26 @@
+'use strict';
+
+module.exports = function ( state, action ) {
+	if ( typeof state === "undefined" ) {
+		state = { isValidating: false, runningValidations: 0 };
+	}
+	switch ( action.type ) {
+		case 'BEGIN_BANK_DATA_VALIDATION':
+		case 'BEGIN_EMAIL_ADDRESS_VALIDATION':
+		case 'BEGIN_ADDRESS_VALIDATION':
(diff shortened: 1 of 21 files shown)


app/js/lib/reducers/async_requests.js
```diff
@@ -0,0 +1,26 @@
...
+		case 'BEGIN_ADDRESS_VALIDATION':
+		case 'BEGIN_PAYMENT_DATA_VALIDATION':
+			return { isValidating: true, runningValidations: state.runningValidations + 1 };
+		case 'FINISH_BANK_DATA_VALIDATION':
+		case 'FINISH_EMAIL_ADDRESS_VALIDATION':
+		case 'FINISH_ADDRESS_VALIDATION':
+		case 'FINISH_PAYMENT_DATA_VALIDATION':
+			if ( state.runningValidations === 0 ) {
+				return state;
+			}
```
JeroenDeDauw: I do not understand what this if block is needed for

app/js/lib/store_update_handling.js
```diff
@@ -50,5 +50,27 @@ module.exports = {
...
+	makeEventHandlerWaitForAsyncFinish: function ( handler, store ) {
+		return function () {
+			var unsubscribe = null;
+
+			if ( !store.getState().asynchronousRequests.isValidating ) {
+				handler();
+				return;
+			}
+
+			if ( unsubscribe !== null ) {
```
JeroenDeDauw: So the execution of `handler` is what can change the `unsubscribe` var?

gbirke: No, it's the anonymous function in lines 66-73. That function checks if all asynchronous requests have finished and if they did, the handler and `unsubscribe` get executed and `unsubscribe` is unset.

app/js/lib/reducers/async_requests.js
gbirke: It's a guard clause against a case that "should never happen" - an asynchronous action being finished without being started. You could argue that this masks an error and throw an exception here. OTOH, there is no harm in just returning the state because all we care about is tracking running asynchronous requests.

JeroenDeDauw: Right. I actually somehow got the impression that state had a different format than what is normally returned. Looking again this seems not to be the case.

PR #807, opened 2017-01-16
Create a standalone Vagrant distribution

gbirke:
The Fundraising Frontend can new be run with the provided Vagrantfile. Only the Template wiki is needed.

### DIFF
--- a/build/vagrant/configureForMysql.sh
+++ b/build/vagrant/configureForMysql.sh
@@ -1,9 +1,17 @@
 #!/bin/bash -ex
 
+if [[ -z "$DB_PASSWD" ]]; then
+    DB_PASSWD="INSECURE PASSWORD"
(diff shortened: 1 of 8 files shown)


build/vagrant/configureForMysql.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -ex
+
+if [[ -z "$DB_PASSWD" ]]; then
+    DB_PASSWD="INSECURE PASSWORD"
+fi
+
+mysql -pPASSWORD_HERE -u root -e 'CREATE DATABASE fundraising;'
```
JeroenDeDauw: should there not be a space here?

JeroenDeDauw: I don't even

PR #813, opened 2017-01-19
Replaced constant visibility docs with PHP 7.1 feature

JeroenDeDauw:
Switches TravisCI to PHP 7.1. Should not be merged before we are fine with having 7.1 on master clearly :)


JeroenDeDauw:
Such a merge? @KaiNissen @gbirke

PR #814, opened 2017-01-19
Use nullable parameters and return types

JeroenDeDauw:
Requires PHP 7.1

Updated CS rules. Had to disable two due to
https://github.com/squizlabs/PHP_CodeSniffer/issues/1292

Temporarily disabled MW CS sniffs since it still hardcodes its
PHPCS version...
https://github.com/wikimedia/mediawiki-tools-codesniffer/pull/1
https://gerrit.wikimedia.org/r/#/c/333123/


gbirke:
I'd rather wait with using new language features until the tools have caught up with the language instead of disabling checks, but that's just my opinion. Let's talk about this in the team.

JeroenDeDauw:
Only had to disable some checks. Basically the CS checks go from checking 65% of stuff to 64% of CS stuff (made up numbers, but I figure it is something like that) (not counting the temporary disable of the MW checks). That ought to be made up for easily by the value we get from the new features. With just this change I found a dead constant. And you verify stuff that is supposed to be private is indeed not used elsewhere (CS tool does not do that). And that is just partial application of the constant visibility stuff...

JeroenDeDauw:
Also note how we disabled PHPMD at the start of the project since this this is broken with PHP7.

JeroenDeDauw:
Oh, also note how this update to the CS stuff caught some issues that did not get caught before. Search `??` in the diff for an example.

PR #821, opened 2017-02-03
Membership application confirmation

KaiNissen:
In order to reflect the additional payment type on the confirmation page, further payment related parameters need to be passed to the template.

Once this is merged, the templates can be changed.

### DIFF
--- a/src/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenter.php
+++ b/src/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenter.php
@@ -5,7 +5,6 @@ namespace WMDE\Fundraising\Frontend\Presentation\Presenters;
 use WMDE\Fundraising\Frontend\MembershipContext\Domain\Model\Applicant;
 use WMDE\Fundraising\Frontend\MembershipContext\Domain\Model\Application;
 use WMDE\Fundraising\Frontend\MembershipContext\UseCases\ShowMembershipApplicationConfirmation\ShowMembershipAppConfirmationResponse;
-use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\BankData;
 use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\DirectDebitPayment;
 use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\PaymentMethod;
 use WMDE\Fundraising\Frontend\Presentation\TwigTemplate;
@@ -44,6 +43,8 @@ class MembershipApplicationConfirmationHtmlPresenter {
 	private function getApplicationArguments( Application $membershipApplication, string $updateToken ): array {
 		return [
 			'id' => $membershipApplication->getId(),
+			'paymentType' => $membershipApplication->getPayment()->getPaymentMethod()->getType(),
+			'status' => $this->mapStatus( $membershipApplication->isConfirmed() ),
 			'membershipFee' => $membershipApplication->getPayment()->getAmount()->getEuroString(),
 			'paymentIntervalInMonths' => $membershipApplication->getPayment()->getIntervalInMonths(),
 			'updateToken' => $updateToken
@@ -72,6 +73,16 @@ class MembershipApplicationConfirmationHtmlPresenter {
 		}
 
 		return [];
+	}
+
+	/**
+	 * Maps the membership application's status to a translatable message key
+	 *
+	 * @param bool $isConfirmed
+	 * @return string
+	 */
+	private function mapStatus( bool $isConfirmed ): string {
+		return $isConfirmed ? 'status-booked' : 'status-unconfirmed';
 	}
 
 }
--- a/tests/Integration/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenterTest.php
+++ b/tests/Integration/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenterTest.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Tests\Integration\Presentation\Presenters;
+
+use WMDE\Fundraising\Frontend\MembershipContext\Tests\Data\ValidMembershipApplication;
+use WMDE\Fundraising\Frontend\MembershipContext\UseCases\ShowMembershipApplicationConfirmation\ShowMembershipAppConfirmationResponse;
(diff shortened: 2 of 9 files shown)


src/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenter.php
```diff
@@ -74,4 +76,8 @@ private function getBankDataArguments( PaymentMethod $payment ): array {
 		return [];
 	}
 
+	private function mapApplicationStatusToTranslatablePaymentStatus( bool $isConfirmed ): string {
+		return $isConfirmed ? 'B' : 'X';
```
gbirke: I don't like using short "magic strings". If you proceed with this PR, please put them in constants with meaningful names.

JeroenDeDauw:
At DDDEU one person had this nice slide with the line

> DRY is COUPLING

That was in a session about Bounded Contexts. And in the closing keynote by Eric Evans there was

img_20170203_180057

--

So unless the cost of not doing so is really high, we should not share code between Donation and Membership context. Right now we have some stuff in the Payment Context as a shared core. That's fine... though we better be careful with that. What really is not good here (even ignoring the Bounded Context part) is that this PR introduces a cyclic dependency. Donation depends on Payment and Membership Depends on Payment. This PR also reverses the arrows, making everything depend on everything.

Let me know if you want me to comment more here. Will try to look again tomorrow

gbirke:
I'd also recommend breaking up `PaymentStatus` into two classes for Donation and Membership.

KaiNissen:
But wouldn't a cyclic dependency be, err, cyclic? I see the point of bounded contexts, though.

So I split up `PaymentStatus` into `MembershipStatusMapper` and `DonationStatusMapper`. I moved away from static functions in these classes and injected the mappers into the respective presenters.

JeroenDeDauw:
> But wouldn't a cyclic dependency be, err, cyclic?

I am confoose. I'm getting the impression we should talk about the relation of the high level parts of our codebase and make sure we're on the same page, perhaps doing some simple whiteboard stuff. Also we could fully split the contexts (also in the top level factory) and run the tests for each individually, which gains us speed (no point running membership tests if you only touched donation code) and detection of corruptions of the abstraction.

contexts/DonationContext/src/Domain/Model/DonationStatusMapper.php
```diff
@@ -0,0 +1,39 @@
...
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\DonationContext\Domain\Model;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class DonationStatusMapper {
```
JeroenDeDauw: Here it's good to have a short comment that says `Maps x to y`

contexts/DonationContext/src/Domain/Model/DonationStatusMapper.php
```diff
@@ -0,0 +1,39 @@
...
+
+	const DONATION_STATUS_PENDING = 'P';
+	const DONATION_STATUS_NEW = 'N';
+	const DONATION_STATUS_UNCONFIRMED = 'X';
+	const DONATION_STATUS_PLEDGE = 'Z';
+	const DONATION_STATUS_BOOKED = 'B';
+	const DONATION_STATUS_BOOKING = 'E';
+
+	public function mapStatus( $status ): string {
```
JeroenDeDauw: can type hint or not?

contexts/DonationContext/src/Domain/Model/DonationStatusMapper.php
JeroenDeDauw: If I understand correctly this thing maps the Donation status we use in our domain model to a presentation specific status.

Two issues:

1. The Domain model really should not contain anything presentation specific. What made you not simply put this in the presenter? (Not saying it should be there, but it's the place I'd start with considering)

2. It is weird this mapper thing defines the constants for values in the domain model. Don't we already have those defined as constants somewhere? If not, having them in `Donation` or some `DonationStatus` thing seems much more appropriate.

src/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenter.php
```diff
@@ -74,4 +75,14 @@ private function getBankDataArguments( PaymentMethod $payment ): array {
 		return [];
 	}
 
+	/**
+	 * Maps the membership application's status to a translatable message key
+	 *
+	 * @param bool $isConfirmed
+	 * @return string
+	 */
+	private function mapStatus( bool $isConfirmed ): string {
```
JeroenDeDauw: Small thing: I'd probably have passed the whole application here. Now the caller kinda knows that only the isConfirmed part is being used to determine the status

tests/Integration/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenterTest.php
```diff
@@ -0,0 +1,72 @@
...
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class MembershipApplicationConfirmationHtmlPresenterTest extends \PHPUnit_Framework_TestCase {
+
+	private const STATUS_BOOKED = 'status-booked';
+	private const STATUS_UNCONFIRMED = 'status-unconfirmed';
+
+	/** @dataProvider applicationStatusProvider */
+	public function testWhenPresenterPresents_itPassesParametersToTemplate( $isConfirmed, $expectedMappedStatus ) {
```
JeroenDeDauw: Forgetti'd type hints

tests/Integration/Presentation/Presenters/MembershipApplicationConfirmationHtmlPresenterTest.php
```diff
@@ -0,0 +1,72 @@
...
+		$twig = $this->getMockBuilder( TwigTemplate::class )->disableOriginalConstructor()->getMock();
+		$twig->expects( $this->once() )
+			->method( 'render' )
+			->with( $this->getExpectedRenderParams( $expectedMappedStatus ) );
+
+		$membershipApplication = ValidMembershipApplication::newDomainEntityUsingPayPal();
+		if ( $isConfirmed === true ) {
+			$membershipApplication->confirm();
+		}
```
JeroenDeDauw: This seems like abusing data providers a bit. Having two tests with factored out Arrange Act and Assert parts as appropriate is probably nicer.

PR #822, opened 2017-02-06
Switch to PHPUnit 6.x

JeroenDeDauw:


composer.json
```diff
@@ -51,16 +51,19 @@
 					]
 				}
 			}
+		},
+		{
+			"type": "vcs",
+			"url": "https://github.com/JeroenDeDauw/covers-validator.git"
```
JeroenDeDauw: We can remove this once https://github.com/oradwell/covers-validator/pull/12 is merged, but this works for now

PR #824, opened 2017-02-07
apply donation confirmation page styles to membership confirmation page

KaiNissen:
While fixing the [browser tests](https://github.com/wmde/fundraising-browsertests/pull/17), a change to the "membership application bank data confirmation page" broke the layout a bit. This PR reapplies the styles for that page.

### DIFF
--- a/web/skin/10h16/_styles/styles.css
+++ b/web/skin/10h16/_styles/styles.css
@@ -528,17 +528,17 @@ main { padding-top: 40px; padding-bottom: 100px; background: #f6f6f6; display: b
 .donation-submit button[type="submit"] { vertical-align: bottom; }
 .donation-submit .label { font-size: 0.8em; line-height: 1.5em; margin-top: 2px; }
 
-#donation-sepa-confirmation .box-header { background: #64aadb; }
-#donation-sepa-confirmation .box-section { font-size: 0.8em; color: #000; }
-#donation-sepa-confirmation .title { font-size: 1.2em; }
-#donation-sepa-confirmation .address { width: 250px; line-height: 1.5em; }
-#donation-sepa-confirmation .payment { line-height: 1.5em; }
-#donation-sepa-confirmation .payment .label { width: 175px; }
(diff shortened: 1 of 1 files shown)


web/skin/10h16/_styles/styles.css
```diff
@@ -528,17 +528,17 @@ main { padding-top: 40px; padding-bottom: 100px; background: #f6f6f6; display: b
...
-#donation-sepa-confirmation .title { font-size: 1.2em; }
-#donation-sepa-confirmation .address { width: 250px; line-height: 1.5em; }
-#donation-sepa-confirmation .payment { line-height: 1.5em; }
-#donation-sepa-confirmation .payment .label { width: 175px; }
-#donation-sepa-confirmation .address + .payment .label { width: 82px; }
-#donation-sepa-confirmation .step-list { font-size: 0.8em; }
-#donation-sepa-confirmation .step-list li { width: auto; padding-right: 25px; background: #80cd71 url(../_assets/img/sheet_step_list_item_2.png) right center no-repeat; color: #b6e6ad; }
-#donation-sepa-confirmation .step-list li.first { background: #85d576 url(../_assets/img/sheet_step_list_item_1.png) right center no-repeat; }
-#donation-sepa-confirmation .step-list li.last { background: #7bc46d; }
+section[id$="-sepa-confirmation"] .box-header { background: #64aadb; }
```
JeroenDeDauw: So now this selects sections with an id that is equal to `-sepa-confirmation`? Not sure if the code is wrong or I simply can't read this stuff :)

gbirke: It selects sections with IDs that end in "-sepa-confirmation". The id on the membership page was also "donation-sepa-confirmation". 

A better way to solve this would be a CSS class that'll be used for all the headers. Using classes is faster than using attribute selectors and looks more familiar. But since the HTML and CSS will be thrown away anyway later this year, I'll just merge.

PR #831, opened 2017-02-15
Fixed risky test in PiwikVariableCollectorTest

JeroenDeDauw:


JeroenDeDauw:
Btw, I wanted to use `iterable` type hint here since there is no reason to assume the stuff is `array` in the method, but PHPStorm is derp: https://youtrack.jetbrains.com/issue/WI-32378

PR #834, opened 2017-02-15
automatically set salutation for companies

KaiNissen:
Fix for wmde/fundraising#1115

This adds a fixed salutation for company donors and membership applicants. I also added some constant scopes to the `Application` domain and refactored the way `MembershipApplicationBuilder` creates the `ApplicantName`.

### DIFF
--- a/contexts/DonationContext/src/Domain/Model/DonorName.php
+++ b/contexts/DonationContext/src/Domain/Model/DonorName.php
@@ -18,6 +18,8 @@ class DonorName {
 	public const PERSON_COMPANY = 'firma';
 	public const PERSON_ANONYMOUS = 'anonym';
 
+	public const COMPANY_SALUTATION = 'Firma';
+
 	private $personType = '';
 
 	private $companyName = '';
@@ -53,7 +55,7 @@ class DonorName {
 	}
 
 	public function getSalutation(): string {
-		return $this->salutation;
+		return $this->personType === self::PERSON_COMPANY ? self::COMPANY_SALUTATION : $this->salutation;
 	}
 
 	public function setSalutation( string $salutation ) {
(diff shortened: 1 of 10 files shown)


JeroenDeDauw:
There are some issues here.

First off, the `Donation` and `Application` objects do not directly know about salutation, so having the constant there seems rather odd. `DonorName` and `ApplicantName` seem better.

Secondly, I'm not sure about this defaulting happening when the entities are first created. I suspect the salutation for a company should always be `"Firma"`. (Is that the case?) If so, then doing this defaulting in the UC is not ideal. It can be modified later on, donations getting into the system via another route don't have this applied to them, and ones already persisted are also not affected.

Perhaps we could have the name object always return `"Firma"` in the salutation getter for companies. (And perhaps we should have a company name object rather than the somewhat awkward stuff we have now, not sure). This leads me to the question of what this Firma thing is for though. What I wrote only makes sense if this really is part of the domain. The linked issue does not really answer that for me.

contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -116,7 +116,7 @@ private function getPhysicalAddressFromRequest( AddDonationRequest $request ): D
 	private function getNameFromRequest( AddDonationRequest $request ): DonorName {
 		$name = $request->donorIsCompany() ? DonorName::newCompanyName() : DonorName::newPrivatePersonName();
 
-		$name->setSalutation( $request->getDonorSalutation() );
+		$name->setSalutation( $request->donorIsCompany() ? Donation::COMPANY_SALUTATION : $request->getDonorSalutation() );
```
JeroenDeDauw: So even if a salutation is provided for a company, it will be changed? Is that intentional?

KaiNissen:
@JeroenDeDauw, what you commented definitely makes sense. I changed my PR accordingly.

Your assumption regarding the purpose of this was correct. The salutation for companies is a default value. Currently, there is no way for users to set this (and probably never will be).

PR #835, opened 2017-03-01
never moderate donations using external payment providers

KaiNissen:
PR for wmde/fundraising#1224

### DIFF
--- a/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
+++ b/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
@@ -60,7 +60,7 @@ class AddDonationUseCase {
 
 		$donation = $this->newDonationFromRequest( $donationRequest );
 
-		if ( $this->policyValidator->needsModeration( $donationRequest ) ) {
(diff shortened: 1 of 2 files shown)


contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -60,7 +60,7 @@ public function addDonation( AddDonationRequest $donationRequest ): AddDonationR
 
 		$donation = $this->newDonationFromRequest( $donationRequest );
 
-		if ( $this->policyValidator->needsModeration( $donationRequest ) ) {
+		if ( !$donation->hasExternalPayment() && $this->policyValidator->needsModeration( $donationRequest ) ) {
```
JeroenDeDauw: At this point I'd have factored out the predicate into a donationNeedsModeration method

KaiNissen: Yeah, I was thinking about that, too. But then it felt awkward to have a `needsModeration` that itself calls `needsModeration`. Let me fix that quickly.

PR #836, opened 2017-03-01
extract condition regarding moderation

KaiNissen:
see https://github.com/wmde/FundraisingFrontend/pull/835#discussion_r103739153

### DIFF
--- a/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
+++ b/contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
@@ -60,7 +60,7 @@ class AddDonationUseCase {
 
 		$donation = $this->newDonationFromRequest( $donationRequest );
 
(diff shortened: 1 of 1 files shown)


contexts/DonationContext/src/UseCases/AddDonation/AddDonationUseCase.php
```diff
@@ -185,4 +185,9 @@ private function sendDonationConfirmationEmail( Donation $donation ) {
 			$this->mailer->sendConfirmationMailFor( $donation );
 		}
 	}
+
+	private function donationNeedsModeration( AddDonationRequest $donationRequest, Donation $donation ): bool {
+		return !$donation->hasExternalPayment() && $this->policyValidator->needsModeration( $donationRequest );
```
JeroenDeDauw: You beat me to it! :) But meh... I had not noticed one used the Entity and the other used the request model >_>

PR #837, opened 2017-03-03
omit validation dispatch on unset required fields

KaiNissen:
A validation should not be dispatched when one of the validated fields have not been touched by the user, since it will always return an error.

see wmde/fundraising#1277, but this also happens when switching the payment method before actually choosing a donation amount on page 1 of the donation form.

### DIFF
--- a/app/js/lib/validation_dispatchers/amount.js
+++ b/app/js/lib/validation_dispatchers/amount.js
@@ -17,13 +17,15 @@ var objectAssign = require( 'object-assign' ),
 	 * @return {ValidationDispatcher}
 	 */
 	createAmountValidationDispatcher = function ( validator,  initialValues ) {
-		var fieldNames = [ 'amount', 'paymentType' ];
+		var fieldNames = [ 'amount', 'paymentType' ],
+			fieldValuesIfNotSet = { amount: '0,00' };
 
 		return objectAssign( Object.create( ValidationDispatcher ), {
 			validationFunction: validator.validate.bind( validator ),
 			finishActionCreationFunction: Actions.newFinishPaymentDataValidationAction,
 			beginActionCreationFunction: Actions.newBeginPaymentDataValidationAction,
 			fields: fieldNames,
+			fieldValuesIfNotSet: fieldValuesIfNotSet,
 			previousFieldValues: _.pick( initialValues || {}, fieldNames )
 		} );
 	};
--- a/app/js/lib/validation_dispatchers/base.js
+++ b/app/js/lib/validation_dispatchers/base.js
@@ -20,6 +20,7 @@ var _ = require( 'underscore' ),
 		finishActionCreationFunction: null,
 		beginActionCreationFunction: null,
 		fields: null,
+		fieldValuesIfNotSet: null,
 		previousFieldValues: {},
 
 		/**
@@ -36,6 +37,10 @@ var _ = require( 'underscore' ),
 				return;
 			}
 
+			if ( !this.fieldsRequiredForDispatchAreSet( selectedValues ) ) {
+				return;
+			}
+
 			this.previousFieldValues = selectedValues;
 
 			if ( this.beginActionCreationFunction ) {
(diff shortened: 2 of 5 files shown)


app/js/lib/validation_dispatchers/base.js
```diff
@@ -36,13 +37,31 @@ var _ = require( 'underscore' ),
...
+
+			if ( _.isObject( formValues ) && this.fieldValuesIfNotSet !== null ) {
+				_.each( this.fieldValuesIfNotSet, function ( initalValue, fieldName ) {
+					if ( initalValue === formValues[ fieldName ] ) {
+						allFieldsSet = false;
+					}
+				} );
+			}
+
+			return allFieldsSet;
```
JeroenDeDauw: Is this a political statement that you support the state? :)

More seriously though, you can just `return false` inside the if and `return true` at the end of the function, no need for state.

app/js/lib/validation_dispatchers/base.js
```diff
@@ -20,6 +20,7 @@ var _ = require( 'underscore' ),
 		finishActionCreationFunction: null,
 		beginActionCreationFunction: null,
 		fields: null,
+		fieldValuesIfNotSet: null,
```
JeroenDeDauw: I was a bit confused by this, since it's never written to, until I realized this is a base class. (Such code sharing via inheritance!) I think it'd be nice to have the expected type indicated here, so you don't need to go look at derivative classes to see what's expected.

JeroenDeDauw: I'm not sure about the name of this. It could be accurate, but I'm unable to figure out what exactly this is. Is it the initial values? Looking at the loop below that does not seem to be far off, but then the derivatives are not specifying this for all their fields...

Perhaps a description that says this should be a map with the fields required for dispatch and their initial values?

app/js/lib/validation_dispatchers/base.js
KaiNissen: This is true for `jQuery.each()`, not for `underscore.each()`. The latter does not break out of the loop, when the passed method returns something.

app/js/lib/validation_dispatchers/amount.js
```diff
@@ -17,13 +17,15 @@ var objectAssign = require( 'object-assign' ),
 	 * @return {ValidationDispatcher}
 	 */
 	createAmountValidationDispatcher = function ( validator,  initialValues ) {
-		var fieldNames = [ 'amount', 'paymentType' ];
+		var fieldNames = [ 'amount', 'paymentType' ],
+			fieldValuesIfNotSet = { amount: '0,00' };
```
gbirke: I'd rather name this `requiredFields` and also put `paymentType` in it as well, as we are not selecting a default payment type.

app/js/lib/validation_dispatchers/amount.js
```diff
@@ -17,13 +17,15 @@ var objectAssign = require( 'object-assign' ),
...
 
 		return objectAssign( Object.create( ValidationDispatcher ), {
 			validationFunction: validator.validate.bind( validator ),
 			finishActionCreationFunction: Actions.newFinishPaymentDataValidationAction,
 			beginActionCreationFunction: Actions.newBeginPaymentDataValidationAction,
 			fields: fieldNames,
+			fieldValuesIfNotSet: fieldValuesIfNotSet,
```
gbirke: Both arrays are short enough that they could be given inline.

app/js/lib/validation_dispatchers/base.js
gbirke: Yes, it should be called `requiredFields`

app/js/lib/validation_dispatchers/base.js
```diff
@@ -36,13 +37,31 @@ var _ = require( 'underscore' ),
...
 			validationResult = this.validationFunction( selectedValues );
 			return store.dispatch( this.finishActionCreationFunction( validationResult ) );
+		},
+
+		fieldsRequiredForDispatchAreSet: function ( formValues ) {
+			var allFieldsSet = true;
+
+			if ( _.isObject( formValues ) && this.fieldValuesIfNotSet !== null ) {
+				_.each( this.fieldValuesIfNotSet, function ( initalValue, fieldName ) {
+					if ( initalValue === formValues[ fieldName ] ) {
```
gbirke: Won't this break when initial values are set by a banner and are not in need of change?

app/js/lib/validation_dispatchers/base.js
JeroenDeDauw: Oh oops... I had not seen the `function` :) Still, the imperativeness of this code seriously irks me. I don't know this library (or JS well for that matter), so dunno if there is a better way. @gbirke any idea?

PR #839, opened 2017-03-06
Delay asyncnchronous validators

gbirke:
Modify the asynchronous form validators to avoid sending incomplete data
to the server which might result in validation error being displayed too
early.
Change the validation result reducers to handle the "INCOMPLETE"
validation result.
Add test cases.

### DIFF
--- a/app/js/lib/form_validation.js
+++ b/app/js/lib/form_validation.js
@@ -72,7 +72,11 @@ var jQuery = require( 'jquery' ),
 		validationUrl: '',
 		sendFunction: null,
 		validate: function ( formValues ) {
-			var postData = {
+			var postData;
+			if ( !formValues.email ) {
+				return { status: ValidationStates.INCOMPLETE };
+			}
+			postData = {
 				email: formValues.email
 			};
 			return jQueryDeferredToPromise( this.sendFunction( this.validationUrl, postData, null, 'json' ) );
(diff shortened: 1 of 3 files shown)


gbirke:
This PR is an alternative to #837, which implements a check at the dispatcher level. "Dispatcher" is the class that triggers the validation when form values change. 

I think we should use this code, as it has three advantages:
1. **Consistency**: So far, incomplete data (for bank data and address data) was checked at the validation level and returned the `INCOMPLETE` status. This PR follows the same pattern. 
2. **Encapsulation**: The knowledge what "incomplete data" means is better placed in the validator and not in the dispatcher. A good example is the amount validator where `'0,00'` is a non-empty JavaScript value, but still must be considered an incomplete value.

But I'm also open for suggestions, maybe #837 is the better approach and we can extend it to encompass the bank and address data checking and ditch the `INCOMPLETE` status.

app/js/lib/form_validation.js
```diff
@@ -83,7 +87,12 @@ var jQuery = require( 'jquery' ),
 		validationUrl: '',
 		sendFunction: null,
 		validate: function ( formValues ) {
-			var postData = {
+			var amountAsFloat = parseFloat( formValues.amount ),
+				postData;
+			if ( amountAsFloat === 0 || isNaN( amountAsFloat ) || !formValues.paymentType ) {
```
KaiNissen: condition could be extracted

app/js/lib/form_validation.js
```diff
@@ -95,7 +104,12 @@ var jQuery = require( 'jquery' ),
 		validationUrl: '',
 		sendFunction: null,
 		validate: function ( formValues ) {
-			var postData = {
+			var amountAsFloat = parseFloat( formValues.amount ),
+				postData;
+			if ( amountAsFloat === 0 || isNaN( amountAsFloat ) || !formValues.addressType || !formValues.paymentIntervalInMonths ) {
```
KaiNissen: condition could be extracted

app/js/lib/form_validation.js
KaiNissen: Since we use the German locale for the amount, this only works, because the minimum donation value is 1 Euro.

gbirke: You mean this would fail for `0,50`? Right, that could bite us. The correct way would be to handle the amount in a localized fashion, but that's out of the scope of this PR. Will put a comment in the extracted condition.

PR #846, opened 2017-03-20
Content source switch

KaiNissen:
This PR is about changing the source of the content from a wiki to local files. (see [phab:T157561](https://phabricator.wikimedia.org/T157561))


gbirke:
Looks fine to me, but due to browser testing SNAFUs I could not run our browser tests against it.
Added some small improvements in #848 

Unit tests will fail when the content repository exists, as the rendered output will contain the translated strings instead of the translation keys that are tested for. But the introduction of the `i18n-base-path` configuration setting and leaving it empty in the test environment (and maybe adding an `if` check in the Translator initialization) will take care of that.

PR #851, opened 2017-03-24
Define undefined constant

JeroenDeDauw:
I'm not sure what the idea where was, but supposdely this should be defined


gbirke:
I'd rather remove the method where the ID is used.

PR #862, opened 2017-04-06
T162281: integrating content as (dev) dependency

wiese:
For easy dev env setup, as discussed.

Unfortunately I am not sure if master is the right target branch as the restructuring is an in-progress dev itself. @gbirke Please advise

Compatible config entries:

    "i18n-base-path": "vendor/wmde/fundraising-frontend-content/i18n",
    "twig": {
        "loaders": {
          "filesystem": {
              "template-dir": [ "vendor/wmde/fundraising-frontend-content/i18n/%_locale_%/pages", "vendor/wmde/fundraising-frontend-content/templates" ]

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,10 @@
 					]
 				}
 			}
+		},
+		{
+			"type": "vcs",
+			"url": "https://github.com/wmde/fundraising-frontend-content"
 		}
 	],
 	"require-dev": {
@@ -67,7 +71,8 @@
 
 		"symfony/browser-kit": "~2.7",
 		"silex/web-profiler": "~2.0",
-		"sorien/silex-dbal-profiler": "~2.0"
+		"sorien/silex-dbal-profiler": "~2.0",
+		"wmde/fundraising-frontend-content": "dev-master as 1.0.x-dev"
 	},
 	"autoload": {
 		"psr-4": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -67,7 +71,8 @@
 
 		"symfony/browser-kit": "~2.7",
 		"silex/web-profiler": "~2.0",
-		"sorien/silex-dbal-profiler": "~2.0"
+		"sorien/silex-dbal-profiler": "~2.0",
+		"wmde/fundraising-frontend-content": "dev-master as 1.0.x-dev"
```
JeroenDeDauw: Cool, did not know this stuff. https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

What is the advantage of having it here?

JeroenDeDauw:
+0.8

composer.json
wiese: Mainly because (for the moment = dev/prototyping) we are tracking the master instead of a tag, which is not the recommended way of using package management.

JeroenDeDauw: You are still stacking the master branch, even if you have the alias. I can see how an alias would make a difference if it is defined in the repository, but not what the inline thing here solves.

wiese: AFAIK this is _the_ way to track master, otherwise we get

     Could not find package wmde/fundraising-frontend-content at any version for your minimum-stability (stable). Check the package spelling or your minimum-stability

or we would have to reduce the minimum stability of the consuming project (which we don't want, I assume)

gbirke:
For the time being, I think master is the right target branch.

PR #864, opened 2017-04-06
Fix Vagrant environment and documentation

gbirke:
Fix inconsistencies in Vagrant config template and installation scripts
introduced by the move to the Git-based content repository.

Fix systemd service description to match user names, bug introduced by #860

Add dummy paypal and credit card configuration to config template to
make it more obvious what needs to be configured.

Update README to reflect the changes, make dependcies more obvious and
document the changes introduced by #862

Avoid overriding existing configuration files, if one exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,14 @@ you will need to do additional configuration, as per the configuration section.
 
 System dependencies:
 
-* PHP >= 7
-* php7.0-intl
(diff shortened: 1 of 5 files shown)


KaiNissen:
Generally looks good to me. Will merge when #865 is merged, too.

PR #870, opened 2017-04-12
make twig reusable

wiese:
Separated twig factory methods to make it reusable (for content template sandbox)

### DIFF
--- a/src/Factories/TwigFactory.php
+++ b/src/Factories/TwigFactory.php
@@ -1,0 +1,94 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Factories;
+
+use WMDE\Fundraising\Frontend\Presentation\FilePrefixer;
+use RuntimeException;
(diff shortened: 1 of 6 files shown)


src/Factories/TwigFactory.php
```diff
@@ -0,0 +1,94 @@
...
+
+class TwigFactory {
+
+	private const LOCALE_PLACEHOLDER = '%_locale_%';
+
+	private $config;
+	private $cachePath;
+	private $locale;
+
+	public function __construct(array $config, string $cachePath, string $locale) {
```
JeroenDeDauw: Ugh, we should see if we can re-enable those CS checks

JeroenDeDauw:
Beware that there now is a WIP commit in the history: https://github.com/wmde/FundraisingFrontend/commit/a76f62408aa59a2d7bff346d5335a42a79326a5a

wiese:
@JeroenDeDauw I am sorry for that - still trying to encorporate the different requirements concerning log keeping into one workflow that works for me.

PR #878, opened 2017-05-05
T163880 Move word blacklist/whitelist into Content repo

gbirke:
* Move the word lists to content repo
* Removed of regex feature in word list
* Removed PageRetriever dependency
* Added `ErrorLoggingFileReader` that logs exceptions and always returns a string.
* Improved absolute and relative path handling in FunFunFactory: i18n path and word list paths no longer need to be relative to app directory (but still can be).

Testing this feature is kind of dependent on https://github.com/wmde/fundraising-frontend-content/pull/17 being merged first and the content dependency updated.


JeroenDeDauw:
Personally I prefer to have a link to the issue tracker in the description rather than a number prefix in the title. You can click the link, and people that don't know how to construct the URL from the number are not lost.

src/Infrastructure/ErrorLoggingFileFetcher.php
```diff
@@ -0,0 +1,40 @@
...
+		} catch ( FileFetchingException $e ) {
+			$this->logger->error( $e->getMessage(), [
+				'code' => $e->getCode(),
+				'trace' => $e->getTraceAsString()
+			] );
+			return '';
+		}
+	}
+
+	public function setLogger( LoggerInterface $logger ) {
```
JeroenDeDauw: What's the benefit of implementing `LoggerAwareInterface` (and thus having this method)? As far as I can tell neither are used.

src/Infrastructure/ErrorLoggingFileFetcher.php
```diff
@@ -0,0 +1,40 @@
...
+		$this->logger = $logger;
+	}
+
+	public function fetchFile( $fileUrl ) {
+		try {
+			return $this->wrappedFileFetcher->fetchFile( $fileUrl );
+		} catch ( FileFetchingException $e ) {
+			$this->logger->error( $e->getMessage(), [
+				'code' => $e->getCode(),
+				'trace' => $e->getTraceAsString()
```
JeroenDeDauw: I think there is some standard monolog thing where you log `[ 'exception' => $ex ]`. Looking at the code now, this is what I did elsewhere

tests/Unit/Infrastructure/ErrorLoggingFileFetcherTest.php
```diff
@@ -0,0 +1,54 @@
...
+use WMDE\Fundraising\Frontend\Infrastructure\ErrorLoggingFileFetcher;
+use PHPUnit\Framework\TestCase;
+
+class ErrorLoggingFileFetcherTest extends TestCase {
+
+	public function testGivenSucceedingFileFetcher_itsContentIsReturned() {
+		$logger = $this->createMock( LoggerInterface::class );
+		$logger->method( 'error' )->with( $this->never() );
+		$errorLoggingFileFetcher = new ErrorLoggingFileFetcher(
+			new InMemoryFileFetcher( ['song.txt' => 'I\'m a little teapot'] ),
```
JeroenDeDauw: CS error: `[ 'stuff' ]`, not `['stuff']`

tests/Unit/Infrastructure/ErrorLoggingFileFetcherTest.php
```diff
@@ -0,0 +1,54 @@
...
+use FileFetcher\InMemoryFileFetcher;
+use Psr\Log\LoggerInterface;
+use WMDE\Fundraising\Frontend\Infrastructure\ErrorLoggingFileFetcher;
+use PHPUnit\Framework\TestCase;
+
+class ErrorLoggingFileFetcherTest extends TestCase {
+
+	public function testGivenSucceedingFileFetcher_itsContentIsReturned() {
+		$logger = $this->createMock( LoggerInterface::class );
+		$logger->method( 'error' )->with( $this->never() );
```
JeroenDeDauw: See the second point at https://github.com/wmde/PsrLogTestDoubles/#motivation

`LoggerSpy` is already included in this project

src/Infrastructure/ErrorLoggingFileFetcher.php
```diff
@@ -0,0 +1,40 @@
...
+			] );
+			return '';
```
JeroenDeDauw: I am not too sure about returning an empty string. Like this it's an ErrorLoggingAndSupressingFileFecther. Is there a reason to not rethrow the exception? Perhaps the user of the FileFetcher assumes an exception is thrown on error, as per the interface, and will then do the wrong thing when it gets an empty string? Or perhaps it will assume the error is getting suppressed and then fails when it is not. Like this it falls to the code constructing the object graph to make sure everything fits together, and to do so it needs to know about details of the implementations, "peeking behind" the interface, which I find very suspicious. Perhaps there is an issue with the interface itself?

JeroenDeDauw: Our Logging decorators in this codebase seem to all rethrow

app/templates/White_Words.txt.twig
```diff
@@ -1,13 +0,0 @@
...
-Tuntenhausen
-Marschner
-Möser
-Massing
-Ramsau
-Wixhausen
-Musterle
-Farschweiler
-Asse
-Fucker
```
JeroenDeDauw: This is advancing my German studies by leaps and bounds

src/Infrastructure/ErrorLoggingFileFetcher.php
gbirke: The reasoning behind this is as follows: We want to have some kind of notification when the files are missing, but this error condition is not not serious enough to bring down the whole application with an exception. And it's not an error we can test against because it's a server configuration thing. 

I could do a try...catch in the WordListFileReader instead, but then I'd have to pass it a logger, which in my view is not a proper separation of concerns.

The previous code did exactly the same, only it did log a notice, which was dropped, so were never aware that our files were missing.

gbirke: I could change the WordListFileReader class to expect an ErrrorLoggingAndSuppressingFileFetcher instead of a FileFetcher, would that be an acceptable solution for you?

tests/Unit/Infrastructure/ErrorLoggingFileFetcherTest.php
gbirke: You're right, and I'll change it.

But as a side note: Does our phpcs config reflect that already? When we started this project, we used the MediaWiki standard which got [changed in Sep 2016](https://www.mediawiki.org/w/index.php?title=Manual:Coding_conventions/PHP&oldid=2227991) to the more spacey style and I think our PHP code mostly adheres to the old style.

src/Infrastructure/ErrorLoggingFileFetcher.php
JeroenDeDauw: > I could do a try...catch in the WordListFileReader instead

What about doing one there "as well" instead of "instead"? You would not need to log there, just catch the error, and have that code decide that the error should not bring down the app. Isn't that a better place to have this decision made? That also leaves you with a decorator that does not have an And in its name.

tests/Unit/Infrastructure/ErrorLoggingFileFetcherTest.php
JeroenDeDauw: I'm not sure it was changed... might just have been unspecified in the doc before, since you could not use PHP 5.4 stuff in MediaWiki anyway. AFAIK we used the spacy style in Wikibase since we first started using `[]`. And it's what I've been using in the fun code since the start.

Not sure about the state of the CS. Might be one of those things that got dropped since it was being derp with (nullable) return types. @thiemowmde do you know if there is a sniff for this?

gbirke:
@wiese Can't add type hints to `fetchFile` since the interface definition has none. But will add `@inheritdoc`.

tests/Unit/Infrastructure/ErrorLoggingFileFetcherTest.php
thiemowmde: Yes, there is a PHPCS sniff that enforces spaces in `[ 'stuff' ]`. It`s MediaWiki.WhiteSpace.SpaceyParenthesis, and requires [mediawiki/mediawiki-codesniffer](https://packagist.org/packages/mediawiki/mediawiki-codesniffer).

JeroenDeDauw:
Bunch of CS fails now related to spacing. I guess your home editor is not configured correctly

JeroenDeDauw:
I made the FileFetcher thing ready for moving over the test double and adding scalar type hints to it with https://github.com/JeroenDeDauw/FileFetcher/pull/7

gbirke:
Ok, waiting for JeroenDeDauw/FileFetcher#7 and then I can make the changes requested by @wiese

JeroenDeDauw:
Merging this already so we can move ahead with this task and then do the scalar type hints as soon as the next version of this lib is released

wiese:
For the record: Thumbs up to @JeroenDeDauw for already updating the library, but we agreed on the fact there is no need to block this story because of a 3rd party interface.

PR #880, opened 2017-05-10
[WIP] Atomic deploy

wiese:
Atomic deploy in line w/ T157575 (https://github.com/wmde/fundraising-infrastructure/pull/129).


deployment/tasks/after_update_code.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Install PHP dependencies
+  composer: command=install working_dir={{ build_dir }} no_dev={{ composer_no_dev }} optimize_autoloader=yes
+
+- name: Install JavaScript dependencies
+  npm: path={{ build_dir }}
+
+- name: Compile JavaScript
+  command: npm run build-js chdir={{ build_dir }}
```
gbirke:
```
- name: Remove node modules
  file:
    path: {{ build_dir }}/node_modules
    state: absent
```

deployment/tasks/before_symlink.yml
```diff
@@ -0,0 +1,32 @@
...
+  file: path={{ app_dir }}/{{ release_name }}/var owner={{app_owner}} group={{app_group}} state=directory
+
+- name: Create var/cache dir
+  file: path={{ app_dir }}/{{ release_name }}/var/cache owner={{app_owner}} group={{app_group}} state=directory mode=775
+
+- name: Create web resources symlink
+  file: src={{ app_dir }}/{{ content_directory }}/resources dest={{ app_dir }}/{{ release_name }}/web/resources owner={{app_owner}} group={{app_group}} state=link
+
+- name: Create cache-busting prefix
+  copy: dest={{ app_dir }}/{{ release_name }}/var/file_prefix.txt content="{{ release_name|hash('md5') }}" owner={{app_owner}} group={{app_group}}
```
gbirke: Until we have a build pipeline, we could use the RELEASE  file instead, or duplicate it for this purpose.

deployment/tasks/before_symlink.yml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Copy config file
+  copy: src={{ config_dir }}/{{ inventory_hostname }}/config.prod.json  dest={{ app_dir }}/{{ release_name }}/app/config/config.prod.json
```
gbirke: We could take the opportunity and convert these entries from equal-sing statements to proper YAML, which would increase the readabiliy.

deployment/tasks/before_symlink.yml
wiese: That would be more of a change in the application than the build process - I'd argue however, that the functioning of the application should not rely on a tick of a specific deployment method (what if the REVSION file is not there?). I do agree though, that the REVSION file would be completely up to the job.

JeroenDeDauw:
Perhaps a bit of a noob question: where are those config files used?

gbirke:
@JeroenDeDauw They are stored (and versioned) on the deploy server in the `configurations` directory of the deploy user. Since they contain passwords, we don't want to put them anywhere else.

wiese:
@JeroenDeDauw As an alternative take to answer your question (b/c I understand it differently), these files are the project-specific deployment steps ("plays") that need to be taken to get FundraisingFrontend working on the machines described in the infrastructure repo. These will stay with the FundraisingFrontend project, while in wmde/fundraising-infrastructure#129 we develop a generic mechanism to accommodate all our projects and call these project-specific hooks at the [appropriate location](https://github.com/wmde/fundraising-infrastructure/pull/129/files#diff-0fc051018e7376f01ff412beef422dbdR12).

JeroenDeDauw:
Ok, thanks for answering. @wiese understood me better.

No updates in 5 days? I hope you where not waiting for me to reply

wiese:
Hi @JeroenDeDauw,
progress has been made, but in wmde/fundraising-infrastructure#129. This feature branch was useful in understanding the requirements (deploying more than one project) but is not really in the sprint, yet (that is T165458).

timEulitz:
This branch has not really seen any action in over a year, is this something we could / should pursue further? Would you have some time to advise on this during the next sprint, maybe @wiese?

deployment/README.md
```diff
@@ -2,38 +2,14 @@
...
- - At the start of the process, create a new directory with a timestamp in its name, e.g. `release-20160816120255`.
- - At the end of the deployment process, change the symbolic link to the new directory.
- - If everything went well, delete the oldest `release-XXX` directory, keeping only the last 5 deployments on the server. The number of kept deployments can be adjusted by changing the variable `keep_releases` in the file `deployment/deployment.yml`
-
-To roll back ("undo") a deployment, log in to the server, change the symlink to a previous deployment and clear the PHP-FPM opcache with the following command:
-
-    php cachetool.phar opcache:reset --fcgi=/var/run/php/php-fpm-current.sock
-
-## Preparing a new server for deployment
+See the documentation in the infrastructure repo.
```
wiese: Moved into infrastructure, see https://github.com/wmde/fundraising-infrastructure/pull/129/files#diff-93d48044e36c7144d51af8956e73fbe5

deployment/tasks/after_update_code.yml
```diff
@@ -0,0 +1,22 @@
+---
```
wiese: This has to be mended to match current (docker & make-base) project build, compare https://github.com/wmde/FundraisingFrontend/blob/master/Makefile

deployment/tasks/after_update_code.yml
```diff
@@ -0,0 +1,22 @@
...
+- name: Install JavaScript dependencies
+  npm:
+    path: "{{ project_dir }}"
+
+- name: Compile JavaScript
+  command: npm run build-js
+  args:
+    chdir: "{{ project_dir }}"
+
+- name: Remove node modules
```
wiese: This became more complicated since as each skin can have its own node_modules.
Rational was to minimize the size of the dist file, could be done likewise of in a more clever way (project-specific ignore file fed to the role [not supported yet!] maybe?).

deployment/tasks/before_symlink.yml
```diff
@@ -0,0 +1,76 @@
+---
```
wiese: There probably are docker/make equivalents for all of this too now.

deployment/tasks/before_symlink.yml
dannyzix: yes you are right my friend

wiese:
This may or may not be the "2 year old 'atomic deploy' patch" that was mentioned during b***&c*** recently.

I stand by [my comment from a year ago](https://github.com/wmde/FundraisingFrontend/pull/880#pullrequestreview-134301930) but want to add the following: Instead of putting more energy into this approach maybe the same time could be better invested taking the next step and deploy containers instead - it's 2019 after all.

As this wasn't a priority for quite some time now (ever since I left for greener pastures and arguably even before) I'll create a reminder for myself to close this in 4 weeks' time - I'm definitely not "owning" it any longer and having it sit here may induce assumptions that will not be met.

Again, I am sure that @gbirke is totally on top of this, but give me a shout if I can be of any help.

/cc @gbirke @timEulitz @tzhelyazkova

gbirke:
Thank you for your comment @wiese, that puts this patch into a new perspective! Now that we have a [docker image for the FundraisingFronted environment](https://hub.docker.com/r/wikimediade/fundraising-frontend), the effort to move the deployment to the atomic-deploy playbook (for which this patch would have to be rewritten) is only slightly smaller than a first start with a containerized version of the app (leaving nginx and db on the host for a minimum of effort). I've created a ticket on how I imagine this transition to go: https://phabricator.wikimedia.org/T225616

Please have a look, feel free to edit/comment.

If you have any ideas on how to keep auto-deploying the i18n repo separately from the manually-deployed Docker image, I would be very thankful.

PR #883, opened 2017-05-22
Rework sliding bank detail banner

KaiNissen:
This PR is a complete overhaul of the bank detail banner that is popping up after a certain idle time on the donation form.

The related bug is described in wmde/fundraising#1282, the task is tracked in [phab:162444](https://phabricator.wikimedia.org/T162444).


JeroenDeDauw:
Such CSS

+0.1

app/templates/Bank_Detail_Banner.html.twig
```diff
@@ -1,131 +1,79 @@
...
-$( document ).ready( function() {
-  timer = window.setTimeout( function() {
-    $( '#bank-detail-banner' ).fadeIn();
-  }, 5000 );
-
-  $( 'input' ).focus( function() {
-    $( '#bank-detail-banner' ).fadeOut();
-    clearTimeout( timer );
-  } );
+	var timer, bankDetailBannerClosed = false;
```
wiese: Mind if we move these inside the .ready callback to make absolutely obvious these are not global?

app/templates/Bank_Detail_Banner.html.twig
```diff
@@ -1,131 +1,79 @@
...
 
-  /* switch banner pages */
-  $( '.switch-button' ).click( function() {
-    $( '#bank-details' ).slideToggle();
-  } );
-} );
+		/* switch banner pages */
+		$( '.switch-button' ).click( function() {
+			var contentWrapper = $( '.content-wrapper' ),
+				newValue = contentWrapper.css( 'top' ) === '0px'
```
wiese: `jQuery.css()`
"When a number is passed as the value, jQuery will convert it to a string and add px to the end of that string. _If the property requires units other than px, convert the value to a string and add the appropriate units before calling the method._" (emphasis mine) TIL

app/templates/Bank_Detail_Banner.html.twig
```diff
@@ -1,131 +1,79 @@
...
+                    ? '-4.4em' : '0em';
+			contentWrapper.animate( { top: newValue } );
```
wiese: `jQuery.animate()`
"The units em and % can be specified where applicable."

app/templates/Base_Layout.html.twig
```diff
@@ -78,6 +78,10 @@
             {% endblock %}
         </div>
     </footer>
+
+    <!-- Banner -->
+    {% block banner %}
+    {% endblock %}
 </div>
```
wiese: Yes!

app/templates/Bank_Detail_Banner.html.twig
wiese: @KaiNissen Would love to hear your opinion before merging...

KaiNissen: No, I'm fine with that. It's unlikely that we need to access them from any other context.

PR #885, opened 2017-05-23
T166225 Fix dependencies of PayPal IPN handling

KaiNissen:
`guzzlehttp/guzzle` was removed in #878. PayPal's IPNs could not be processed since we deployed that, because the handler uses that library to verify the IPN. We have tests for that, but they did not catch any problems, because guzzle was still `require-dev`.


JeroenDeDauw:
Fail :D

PR #888, opened 2017-05-31
T166143 removed passing of non-existing second constructor parameter

wiese:
Fixes phpstan issue:
"AddSubscriptionUseCase: EmailAddress constructor invoked with 2 parameters, 1 required."

see story, https://github.com/wmde/FundraisingFrontend/tree/hey-stan

### DIFF
--- a/contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
+++ b/contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
@@ -4,7 +4,6 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\SubscriptionContext\Tests\Integration\UseCases\AddSubscription;
 
-use PHPUnit_Framework_MockObject_MockObject;
 use WMDE\Fundraising\Entities\Subscription;
 use WMDE\Fundraising\Frontend\Infrastructure\TemplateBasedMailer;
 use WMDE\Fundraising\Frontend\MembershipContext\Domain\Model\EmailAddress;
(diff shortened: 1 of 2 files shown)


contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
```diff
@@ -139,4 +141,19 @@ public function testGivenDataThatNeedsToBeModerated_requestNotBeMailed() {
 		$useCase->addSubscription( $request );
 	}
 
+	public function testNewMailAddressFromSubscription(): void {
+		$useCase = new AddSubscriptionUseCase( $this->repo, $this->validator, $this->mailer );
+		$class = new ReflectionClass( AddSubscriptionUseCase::class );
```
JeroenDeDauw: Taking advantage of me being on holiday to sneak in evil reflection code are we?!

JeroenDeDauw: I'd have gone for giving the UC a request model and seeing if the mailer gets invoked with the correct email object

JeroenDeDauw: I set out to refactor this that way and then noticed that the test is already there. So I removed this addition now in https://github.com/wmde/FundraisingFrontend/pull/896

wiese: Sneaking... The alternative being to insist on not using reflection, claiming that only public API needs testing, and never writing this test.
I removed a constructor parameter and no test twitched an eye, so I wanted to at least make sure that the flow of the relevant data (the mail address) is tested in preparation for the refactoring due in T166687 - fully aware that it's not beautiful, but working with what we got. After #896 we've got none of this security.

JeroenDeDauw: Just to avoid misunderstandings: I was kidding with the sneaking part. If the sentence contains "evil" I'm probably not being serious.

---

As far as I can tell there was no uncovered desired behavior and not even additional behavior of the non-harmful type that was not tested. Just dead code.

Perhaps you missed the part where I said that the used constructor parameter is already tested? Before removing your test I verified this by breaking the code and the failure being caught by both your test and an earlier one.

JeroenDeDauw: > not using reflection, claiming that only public API needs testing

I think that is a good rule by and large... there might be exceptions, esp in legacy codebases... though I can't think of this being called for in code that was TDD'd or well designed code in general. If you have another opinion I'd love to hear your rationale.

PR #890, opened 2017-06-01
T166143 Introducing phpstan + fixing smells

wiese:
This introduces an additional static code analysis tool, [phpstan](https://github.com/phpstan/phpstan/) that detects code smells.  Amongst others, it detects the usage of classes that are not defined. In order to address the initial problem T166143, phpstan is run against the project with composer installed in --no-dev mode. Consequently, phpstan can not be set-up as a dev-dependency itself, and is run in a dedicated docker container instead.

This PR also fixes many smaller smells detected by phpstan.

e8462c2, 009ab1f were reviewed already in #887 and #888 respectively.

### DIFF
--- a/docker/phpstan/Dockerfile
+++ b/docker/phpstan/Dockerfile
@@ -3,35 +3,35 @@
 
 FROM phpstan/phpstan:latest
 
-RUN echo "go" \
-    # for intl
-    && apk add --no-cache --virtual .persistent-deps icu-dev \
-    # for curl
-    && apk add --no-cache --virtual .persistent-deps curl-dev \
-    # for xml
-    && apk add --no-cache --virtual .persistent-deps libxml2-dev \
-    # compare vagrant/install_packages.sh
-    && docker-php-ext-configure intl --enable-intl \
-    && docker-php-ext-install intl curl xml pdo_mysql mbstring
+RUN \
+	# for intl
+	apk add --no-cache --virtual .persistent-deps icu-dev && \
+	# for curl
+	apk add --no-cache --virtual .persistent-deps curl-dev && \
+	# for xml
+	apk add --no-cache --virtual .persistent-deps libxml2-dev && \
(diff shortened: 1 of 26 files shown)


src/Infrastructure/TemplateMailerInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+use WMDE\Fundraising\Frontend\MembershipContext\Domain\Model\EmailAddress;
+
+interface TemplateMailerInterface {
```
JeroenDeDauw: Why interface needed?

KaiNissen: To make sure both the `TemplateBasedMailer` and the `LoggingMailer` expose the same method signature?

wiese: @KaiNissen offered a formidable explanation, thank you. @JeroenDeDauw Which solution do you propose?

docker/phpstan/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# phpstan in docker
+# used as sanity check of code when installed via "composer install --no-dev" (so on prod) w/o other dev tools to check
+
+FROM phpstan/phpstan:latest
+
+RUN echo "go" \
```
KaiNissen: You already explained that this was common practice to have the double ampersands in the lines below aligned. I'm not opposing this and there is a lot of output produced anyway. But what do you think about printing something meaningful rather than just "go"?

wiese:
Rebased

src/Infrastructure/TemplateMailerInterface.php
JeroenDeDauw: I was just asking the question, not objecting. Sounds good

PR #895, opened 2017-06-08
Add working dev default for i18n-base-path setting

JeroenDeDauw:
Without this you need to do extra config for the stuff to work
after dev installation.

Kinda part of https://phabricator.wikimedia.org/T163918

### DIFF
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -20,7 +20,7 @@
 			"array": {}
 		}
 	},
-	"i18n-base-path": "",
+	"i18n-base-path": "vendor/wmde/fundraising-frontend-content/i18n",
 	"default-layout-templates": {
 		"header": "Page_Header.html.twig",
 		"footer": "Page_Footer.html.twig",


gbirke:
Curious, I thought this would break some tests that assume untranslated content.

As fundraising-frontend-content is a dev dependency I'd rather not have this configuration in dist but in test.

wiese:
Bit `config.test.json` does have  `i18n-base-path`. @JeroenDeDauw Can you describe "need to do extra config for the stuff to work after dev installation". Which stuff - tests or browser?

JeroenDeDauw:
Neither: console

```
php console dump-mail-tpl -o /tmp/a
PHP Notice:  Uninitialized string offset: 0 in /vagrant/src/Factories/FunFunFactory.php on line 1350
PHP Notice:  Uninitialized string offset: 0 in /vagrant/src/Factories/FunFunFactory.php on line 1350

  [WMDE\Fundraising\ContentProvider\SetupException]      
  An exception occurred setting up the ContentProvider.  
                                                                
  [Twig_Error_Loader]                                                                                                  
  The "/vagrant/src/Factories/../..//de_DE/web" directory does not exist ("/vagrant/src/Factories/../..//de_DE/web").
```

> As fundraising-frontend-content is a dev dependency I'd rather not have this configuration in dist but in test.

That does not make this console command work out of the box though

JeroenDeDauw:
After reading @gbirke his comment on https://phabricator.wikimedia.org/T163918 I'm thinking this is not needed, since the command is temporary anyway and not actually needed for the task I was poking at.

PR #896, opened 2017-06-08
Remove not needed test

JeroenDeDauw:
Follow up to https://github.com/wmde/FundraisingFrontend/pull/888

This newly added test does not test anything not already tested before.
Introduced a new constant in the test to try make it a bit clearer.

### DIFF
--- a/contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
+++ b/contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
@@ -25,6 +25,8 @@ use ReflectionClass;
  */
 class AddSubscriptionUseCaseTest extends TestCase {
 
+	private const A_SPECIFIC_EMAIL_ADDRESS = 'curious@nyancat.com';
+
 	/**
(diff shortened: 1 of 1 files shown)


contexts/SubscriptionContext/tests/Integration/UseCases/AddSubscription/AddSubscriptionUseCaseTest.php
```diff
@@ -141,19 +143,4 @@ public function testGivenDataThatNeedsToBeModerated_requestNotBeMailed() {
...
 	}
 
-	public function testNewMailAddressFromSubscription(): void {
-		$useCase = new AddSubscriptionUseCase( $this->repo, $this->validator, $this->mailer );
-		$class = new ReflectionClass( AddSubscriptionUseCase::class );
-
-		$method = $class->getMethod( 'newMailAddressFromSubscription' );
-		$method->setAccessible( true );
-
-		$subscription = $this->createMock( Subscription::class );
```
JeroenDeDauw: It's weird to mock a non-service. Though since this is a Doctrine ORM thing and not something more simple I can see why you did it.

I'm writing this comment now since I realized that this should not be the case: UCs should not bind to data access details. And indeed there already is a FIXME for this in SubscriptionRepository

PR #897, opened 2017-06-14
Use tool friendly function refferences

JeroenDeDauw:
This caused me some minutes of WTF at not finding any usages of getWeb and getMail,
both via IDE search and string searcg on ->getWeb(


JeroenDeDauw:
If you want to read my rant on the subject https://www.entropywins.wtf/blog/2016/01/30/missing-in-php7-function-references/

PR #900, opened 2017-06-15
T162439: fix missing link tracking

KaiNissen:
While the tracking is working in the non-test template (`app/templates/Membership_Application_Form.html.twig`), it is missing in the test template.


gbirke:
Deployed to test and production

PR #902, opened 2017-06-18
Test mail templates against previously rendered content

JeroenDeDauw:
https://phabricator.wikimedia.org/T163918

~~This includes https://github.com/wmde/iterable-functions and https://github.com/wmde/traversable-iterator
which should also be reviewed. (Don't worry, each of them only has about nyanthousand LOC.)~~

### DIFF
--- a/cli/RenderMailTemplatesCommand.php
+++ b/cli/RenderMailTemplatesCommand.php
@@ -59,7 +59,11 @@ class RenderMailTemplatesCommand extends Command {
 
 		$testData = require __DIR__ . '/../tests/Data/mail_templates.php';
 
-		$this->validateTemplateFixtures( $testData, $config['twig']['loaders']['filesystem']['template-dir'], $output );
+		$this->validateTemplateFixtures(
+			$testData,
+			iterator_to_array( $ffFactory->newMailTemplateFilenameTraversable() ),
+			$output
+		);
 
 		$outputPath = $input->getOption( 'output-path' ) ?? '';
 		if ( $outputPath && substr( $outputPath, -1 ) !== '/' ) {
@@ -85,14 +89,13 @@ class RenderMailTemplatesCommand extends Command {
 	 * Check that there are templates for all fixtures and (even more important) vice-versa
 	 *
 	 * @param array $testData Template names and fixture information to render these templates
-	 * @param array $mailTemplatePaths Directories containing templates
+	 * @param array $mailTemplatePaths
 	 * @param OutputInterface $output Command output
 	 */
 	private function validateTemplateFixtures( array $testData, array $mailTemplatePaths, OutputInterface $output ): void {
-		$mailTemplatesOnDisk = $this->getMailTemplatesOnDisk( $mailTemplatePaths );
 		$testTemplateNames = array_keys( $testData );
 
-		$untestedTemplates = array_diff( $mailTemplatesOnDisk, $testTemplateNames );
+		$untestedTemplates = array_diff( $mailTemplatePaths, $testTemplateNames );
 
 		if ( !empty( $untestedTemplates ) ) {
 			$output->writeln(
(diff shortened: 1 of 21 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -1372,4 +1373,11 @@ public function setContentProvider( ContentProvider $contentProvider ): void {
 	private function getContentProvider(): ContentProvider {
 		return $this->pimple['content_provider'];
 	}
+
+	public function newMailTemplateFilenameTraversable(): \Traversable {
```
wiese: - I'd prefer being specific in return types (and generous in parameter type hints) unless there is a compelling reason against it
- I reckon we agreed on using `use` statements instead of inline absolute namespaces

src/Infrastructure/MailTemplateFilenameTraversable.php
```diff
@@ -0,0 +1,27 @@
...
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class MailTemplateFilenameTraversable implements \IteratorAggregate {
```
wiese: I reckon we agreed on using `use` statements instead of inline absolute namespaces

src/Infrastructure/MailTemplateFilenameTraversable.php
```diff
@@ -0,0 +1,27 @@
...
+
+	private $mailTemplatePaths;
+
+	public function __construct( array $mailTemplatePaths ) {
+		$this->mailTemplatePaths = $mailTemplatePaths;
+	}
+
+	public function getIterator() {
```
wiese: yield implied that `Generator` is returned. Can we hint that?

cli/RenderMailTemplatesCommand.php
```diff
@@ -152,10 +141,16 @@ private function renderTemplates( array $testData, Twig_Environment $twig, strin
...
-						$templateSettings['context'],
-						$templateSettings['variants'][$variantName]
-					) ) );
+					file_put_contents(
+						$outputName,
+						$twig->render(
+							$templateFileName,
+							array_merge_recursive(
+								$templateSettings['context'],
+								$additionalContext
```
wiese: Nice catch (the previously unused `additionalContext`)!

src/Factories/FunFunFactory.php
JeroenDeDauw: Can you elaborate on why you want it to be more specific? My reasoning: being more specific here would mean using `IteratorAggregate`. Then if you change the implementation, suddenly you have a breaking change here. And why would the consumer care it's an `IteratgorAggregate`? Suppose this was a `Generator`, then I think it would be sensible to type hint as `Iterator`, since the consumer benefits, and most refactorings would result in an `Iterator`.

For use statements I'm using the PHPStorm default: inline when it's a ref to a top level NS class, with use statement if it is not. This default makes sense to me and I'm not sure why you'd want to deviate.

src/Infrastructure/MailTemplateFilenameTraversable.php
JeroenDeDauw: Sure. IIRC there are some other type hints missing or in need of minor tweaking.

This is one of those cases where the benefit of the return type hint is very small tho. I can't think of any valid reason to call this method on the type it is defined in. Still, it won't hurt having it either

src/Factories/FunFunFactory.php
wiese: What is returned is a very concrete `MailTemplateFilenameTraversable` which sure does implement `IteratgorAggregate` but may implement many more interfaces or add methods not described by any interface. We _can_ (as in having the privilege) be more specific about our **return** type w/o any disadvantage, while - sure - anyone hinting **parameter** types should hint the interface (s)he actually wants to use - Robustness principle.

For the use statements I thought there was a rule/agreement. If this is not the case... never mind. I like an inventory of used classes in the head of the file, also it's a small step from `\Foo` to `\Foo\Bar`.

JeroenDeDauw: There is a cost to being more specific here. If you hint it as Traversable, then the consumer knows it should just be used as a Traversable. If you hint as the specific class, which IMO is an implementation detail, then the consumer needs to first figure out that it is just a Traversable and nothing more is going on. In this case it does not matter much either way... if it was a factory in a library or similar situation it'd be much more important.

The rule I've been following is use for \Foo\Bar but not for \Foo, though with doing it anyway for \Foo being fine. Example: our tests used to have `extends \PHPUnit_Framework_TestCase`

JeroenDeDauw: I've updated it now as I don't think this is worth spending much time on

PR #904, opened 2017-06-23
T162438: Membership application confirmation template

KaiNissen:
This PR adds the date of the first payment to the membership application data set to be later retrieved when displaying the confirmation page.

It also adds changes to the membership application confirmation page as described in wmde/fundraising#1083.

### DIFF
--- a/contexts/MembershipContext/src/Domain/Model/Application.php
+++ b/contexts/MembershipContext/src/Domain/Model/Application.php
@@ -166,7 +166,7 @@ class Application {
 		return $this->isDeleted;
 	}
 
-	public function setFirstPaymentDate( string $firstPaymentDate ) {
+	public function notifyOfFirstPaymentDate( string $firstPaymentDate ) {
 		$paymentMethod = $this->getPayment()->getPaymentMethod();
 		if ( $paymentMethod instanceof PayPalPayment ) {
 			$payPalData = $paymentMethod->getPayPalData() ?: new PayPalData();
--- a/contexts/MembershipContext/src/Domain/Model/Application.php
+++ b/contexts/MembershipContext/src/Domain/Model/Application.php
@@ -167,15 +167,12 @@ class Application {
 	}
 
 	public function setFirstPaymentDate( string $firstPaymentDate ) {
-		/** @var PayPalPayment $payPalPayment */
-		$payPalPayment = $this->getPayment()->getPaymentMethod();
-		$payPalData = $payPalPayment->getPayPalData();
-		if ( $payPalData === null ) {
-			$payPalData = new PayPalData();
+		$paymentMethod = $this->getPayment()->getPaymentMethod();
+		if ( $paymentMethod instanceof PayPalPayment ) {
+			$payPalData = $paymentMethod->getPayPalData() ?: new PayPalData();
+			$payPalData->setFirstPaymentDate( $firstPaymentDate );
+			$this->addPayPalData( $payPalData );
 		}
-
-		$payPalData->setFirstPaymentDate( $firstPaymentDate );
-		$this->addPayPalData( $payPalData );
 	}
 
 }
--- a/contexts/MembershipContext/src/Domain/Model/Application.php
+++ b/contexts/MembershipContext/src/Domain/Model/Application.php
@@ -166,4 +166,16 @@ class Application {
 		return $this->isDeleted;
 	}
 
+	public function setFirstPaymentDate( string $firstPaymentDate ) {
+		/** @var PayPalPayment $payPalPayment */
+		$payPalPayment = $this->getPayment()->getPaymentMethod();
+		$payPalData = $payPalPayment->getPayPalData();
+		if ( $payPalData === null ) {
+			$payPalData = new PayPalData();
+		}
+
+		$payPalData->setFirstPaymentDate( $firstPaymentDate );
+		$this->addPayPalData( $payPalData );
+	}
+
 }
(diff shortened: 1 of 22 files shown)


contexts/MembershipContext/src/Domain/Model/Application.php
```diff
@@ -166,4 +166,16 @@ public function isDeleted(): bool {
...
+	public function setFirstPaymentDate( string $firstPaymentDate ) {
+		/** @var PayPalPayment $payPalPayment */
+		$payPalPayment = $this->getPayment()->getPaymentMethod();
+		$payPalData = $payPalPayment->getPayPalData();
+		if ( $payPalData === null ) {
+			$payPalData = new PayPalData();
+		}
+
+		$payPalData->setFirstPaymentDate( $firstPaymentDate );
+		$this->addPayPalData( $payPalData );
```
JeroenDeDauw: Some questions related to this that probably should not be acted upon in this PR:

Why can PayPalPayment have a null PayPalData? I'm seeing some code that is giving it an empty PayPalData, so can't a PayPalData be required? If not for some reason, could the getPayPalData method not always return a PayPalData anyway (by creating an empty instance if the field is null)?

Why do we have a PayPalData class in the first place? PayPalPayment looks like a rather empty container like this. Could the stuff that is in PayPalData not be in PayPalPayment?

contexts/PaymentContext/src/Domain/CalculatingPaymentDelayCalculator.php
```diff
@@ -0,0 +1,28 @@
...
+namespace WMDE\Fundraising\Frontend\PaymentContext\Domain;
+
+use DateInterval;
+use DateTime;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class CalculatingPaymentDelayCalculator implements PaymentDelayCalculator {
```
JeroenDeDauw: This name invoked actual LOL on my part :)

JeroenDeDauw: More seriously though, `Default` or `Normal` are probably better prefixes, and perhaps the domain has some better terms to offer. Question to ask: how would you identify this type of payment calculation when talking to the fun people?

If this is the only implementation we have, it is general, and you don't expect more, then finding a good name is kinda weird since the most obvious one is already stolen by the interface. And in such a case it is not clear to me you need the interface. (Not saying you should not have it)

JeroenDeDauw:
I added a commit that fixes a failing test as it was using the real dates in the production code and a fixed one in the assertion.

contexts/MembershipContext/src/Domain/Model/Application.php
```diff
@@ -166,4 +166,16 @@ public function isDeleted(): bool {
 		return $this->isDeleted;
 	}
 
+	public function setFirstPaymentDate( string $firstPaymentDate ) {
```
JeroenDeDauw: This method blows up when the payment is not of type paypal.

contexts/MembershipContext/src/UseCases/ApplyForMembership/ApplyForMembershipUseCase.php
```diff
@@ -56,6 +60,11 @@ public function applyForMembership( ApplyForMembershipRequest $request ): ApplyF
 		if ( $this->policyValidator->isAutoDeleted( $application ) ) {
 			$application->markAsDeleted();
 		}
+
+		if ( $application->getPayment()->getPaymentMethod()->getType() === PaymentType::PAYPAL ) {
+			$application->setFirstPaymentDate( $this->paymentDelayCalculator->calculateFirstPaymentDate()->format( 'Y-m-d' ) );
+		}
```
JeroenDeDauw: This strikes me as a tell-dont-ask violation. Also not good that this makes the UC bind to a specific payment type. My intuitive reaction here is that this should go into the domain, much like we did when we refactored some stuff in AddDonationUseCase into Donation::notifyOfPolicyValidationFailure. I need to look at this more to figure out what this would look like though.

contexts/MembershipContext/src/Domain/Model/Application.php
JeroenDeDauw: I tried out making the field non-nullable and it worked: https://github.com/wmde/FundraisingFrontend/pull/905

The other question still stands. Furthermore the expectorations around PayPalData are not clear to me. Looks like if you try to get one of its values when it is not initialized yet, you will get an error. And finally the whole PaymentMethod thing comes over as weirdly structured. Not sure how it could be better. Perhaps @wiese with his fresh pair of eyes has an idea?

Not liking the addPayPalData in Donation and Application. Think this should be in the payment method itself. Looking at where this method is used, that should be easy enough to do. (In another PR, since that really is not part of the task here.)

contexts/PaymentContext/src/Domain/CalculatingPaymentDelayCalculator.php
KaiNissen: Calculating payment delay calculator is calculating...

Yeah, I was kind of struggling with finding a name for it.

contexts/MembershipContext/src/Domain/Model/Application.php
KaiNissen: If I remember correctly, we wanted to keep the data blob field clear of unset fields. `PayPalData` is actually being set during the payment notification sent by PayPal. Apart from that I don't see any reason not to use `PayPalPayment` for it.

wiese: Hi,

this is a very complex dependency tree, there is definitely more to discuss, but given I only took a brief look and the actual scope of the ticket...
+1 for not allowing the empty PayPalData in PayPalPayment (#905). Making it immutable is the logical next step.

JeroenDeDauw: > If I remember correctly, we wanted to keep the data blob field clear of unset fields. 

Hmm, if that is the case, then we can change it, because the responsibility of no unset fields in the data blob lies in the data access layer, not the domain model.

PR #906, opened 2017-06-27
Removed addPayPalData from Donation and Application

JeroenDeDauw:
Following up to my comments on https://github.com/wmde/FundraisingFrontend/pull/904

---

I'd like to clean this up a bit further though do not understand the current situation well enough (which indicates cleanup would be good :p). This PR itself is finished though.

Do I understand correctly that the (non-empty) PayPalData thing gets created only when we get a specific notification from PayPal (one for Donation, one for Membership applications), and that afterwards it does not get modified? Except for the new `notifyOfFirstPaymentDate` in Application, which sets just one field.

And is `notifyOfFirstPaymentDate` supposed to always be called before the callback from PayPal that sets the whole object gets invoked? Do we need to retain the old value that was already set then or should we just store the info from the last paypall call?

Disregarding `notifyOfFirstPaymentDate`, does modification ever make sense? `PayPalPayment` has `addPayPalData`, the name of which is a misleading, since it overrides what was set before. Perhaps this should (before introduction of `notifyOfFirstPaymentDate` have thrown an exception when trying to set stuff when it is already set?)

### DIFF
--- a/contexts/DonationContext/src/Domain/Model/Donation.php
+++ b/contexts/DonationContext/src/Domain/Model/Donation.php
@@ -229,20 +229,6 @@ class Donation {
 	}
 
 	/**
-	 * @param PayPalData $payPalData
-	 * @throws RuntimeException
-	 */
-	public function addPayPalData( PayPalData $payPalData ) {
-		$paymentMethod = $this->payment->getPaymentMethod();
-
-		if ( !( $paymentMethod instanceof PayPalPayment ) ) {
-			throw new RuntimeException( 'Cannot set PayPal data on a non PayPal payment' );
-		}
-
-		$paymentMethod->addPayPalData( $payPalData );
-	}
-
-	/**
 	 * @param CreditCardTransactionData $creditCardData
 	 *
 	 * @throws RuntimeException
--- a/contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
+++ b/contexts/DonationContext/src/UseCases/HandlePayPalPaymentNotification/HandlePayPalPaymentNotificationUseCase.php
@@ -80,7 +80,9 @@ class HandlePayPalPaymentNotificationUseCase {
 	}
 
 	private function handleRequestForDonation( PayPalPaymentNotificationRequest $request, Donation $donation ): PaypalNotificationResponse {
-		if ( !( $donation->getPayment()->getPaymentMethod() instanceof PayPalPayment ) ) {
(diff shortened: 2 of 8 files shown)


wiese:
I also tripped on `addPayPalData` - if it's there to stay it should be called `setPayPalData`. However, I would strive to lose it instead. Having it as both, a compulsory constructor param, and set-able via method is not right - hence, my proposal for immutability. 

For the other questions I am also interested in the answers to strengthen my domain knowledge before making ill-advised suggestions.

JeroenDeDauw:
Ping @gbirke @KaiNissen 

Such a mawrg?

JeroenDeDauw:
delivar

KaiNissen:
> Do I understand correctly that the (non-empty) PayPalData thing gets created only when we get a specific notification from PayPal (one for Donation, one for Membership applications), and that afterwards it does not get modified? Except for the new `notifyOfFirstPaymentDate` in Application, which sets just one field.

Correct.

> And is `notifyOfFirstPaymentDate` supposed to always be called before the callback from PayPal that sets the whole object gets invoked? Do we need to retain the old value that was already set then or should we just store the info from the last paypall call?

This field is something that we define before redirecting donors to the payment page. I assume, this is also provided by the payment confirmation sent by PayPal (or can at least be calculated by the data in it) and can be overwritten while processing their "IPN". This whole thing only applies to membership applications, though. For donations we don't provide such a delayed first payment.

> Disregarding `notifyOfFirstPaymentDate`, does modification ever make sense? `PayPalPayment` has `addPayPalData`, the name of which is a misleading, since it overrides what was set before. Perhaps this should (before introduction of `notifyOfFirstPaymentDate` have thrown an exception when trying to set stuff when it is already set?)

Hm, actually `PayPalData` was introduced to hold the data PayPal sent to us when confirming the payment. The data never changes (apart from the fact that we set the date of the first payment before the payment was actually made).

JeroenDeDauw:
@KaiNissen then how about having firstPaymentDate in PayPalPayment rather than PayPalData, making PayPal data immutable, and making addPayPalData throw an exception when you call it when there already is a value?

KaiNissen:
Sounds like a reasonable approach. But what about dropping the whole `PayPalData` thing and stuffing it all into `PayPalPayment`?

JeroenDeDauw:
I was asking about that approach when I did not know this data gets set at different places and times. So now I'm thinking it's nice to keep the mutable (PayPalPayment) and immutable (PayPalData and firstPaymentDate) separate. Did not spend a lot of time thinking about this, so I dunno

PR #901, opened 2017-06-16
[DNM] Enable PayPal as a payment option in membership application form

KaiNissen:
related to [phab:T147400](https://phabricator.wikimedia.org/T147400)

PR #909, opened 2017-06-29
T167875 + T167879 Sofortüberweisung

wiese:
* feature toggle for donation payment types
* introduced SUB to Views + Mail

### DIFF
--- a/app/ab_test/a/templates/Donation_Confirmation_Phone.html.twig
+++ b/app/ab_test/a/templates/Donation_Confirmation_Phone.html.twig
@@ -116,7 +116,7 @@
 						</div>
 
 						<div class="payment f-left">
-							{% if donation.paymentType == 'PPL' or donation.paymentType == 'MCP' %}
+							{% if donation.paymentType == 'PPL' or donation.paymentType == 'MCP' or donation.paymentType == 'SUB' %}
 							<p>
 								Ihre Spende per {$ donation.paymentType|trans({}, 'paymentTypes') $} {$ donation.status|trans({}, 'paymentStatus') $}.
 							</p>
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -160,6 +160,9 @@
 		},
 		"PPL": {
 			"enabled": true
+		},
+		"SUB": {
+			"enabled": true
 		}
 	}
 }
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -147,5 +147,19 @@
 			"pageTitles": "pageTitles.json",
 			"membershipTypes": "membershipTypes.json"
 		}
+	},
+	"payment-types": {
+		"BEZ": {
+			"enabled": true
+		},
+		"UEB": {
+			"enabled": true
+		},
+		"MCP": {
+			"enabled": true
+		},
+		"PPL": {
+			"enabled": true
+		}
 	}
 }
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -150,19 +150,19 @@
 	},
 	"payment-types": {
 		"BEZ": {
-			"enabled": true
+			"donation-enabled": true
 		},
 		"UEB": {
-			"enabled": true
+			"donation-enabled": true
 		},
 		"MCP": {
-			"enabled": true
+			"donation-enabled": true
 		},
 		"PPL": {
-			"enabled": true
+			"donation-enabled": true
 		},
 		"SUB": {
-			"enabled": true
+			"donation-enabled": true
 		}
 	}
 }
--- a/app/config/config.dist.json
+++ b/app/config/config.dist.json
@@ -119,6 +119,11 @@
 		"return-url": "https://spenden.wikimedia.de/show-donation-confirmation",
 		"testmode": true
 	},
+	"sofort": {
+		"config-key": "",
+		"return-url": "",
+		"cancel-url": ""
+	},
 	"confirmation-pages": {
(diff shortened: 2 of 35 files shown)


app/templates/payment_type_list.twig
```diff
@@ -0,0 +1,33 @@
...
+			class="last"
+		{% endif %}>
+			<input
+				id="{$ formId $}-payment-type-{$ loop.index $}"
+				type="radio"
+				class="payment-type-select"
+				name="zahlweise"
+				value="{$ paymentType $}"
+				title="{$ paymentType|trans({}, 'paymentTypes') $}"
+				{% if paymentTypesRequired %}required="required"{% endif %}
```
gbirke: You can drop this line and parameter in the templates. `required` is a vestige from the times when the form was validated with both HTML5 and JavaScript. Since HTML5 is such a mess (cross-browser incompatibilities, bad stylability) we dropped that validation method and its attributes.

src/Presentation/TwigEnvironmentConfigurator.php
```diff
@@ -61,6 +61,7 @@ public function getEnvironment( Twig_Environment $twig, array $loaders, array $e
 
 	private function setDefaultTwigVariables( Twig_Environment $twig ): void {
 		$twig->addGlobal( 'basepath', $this->config['web-basepath'] );
+		$twig->addGlobal( 'appconfig', $this->config );
```
gbirke: I have a bad feeling about passing the whole application configuration (with all the database and payment provider passwords) into the template layer. While it should be secure enough (we're the only ones that touch the templates), I created `setDefaultTwigVariables` especially for this purpose of passing only needed variables.

web/res/js/membershipForm.js
```diff
@@ -162,7 +162,8 @@ $( function () {
 						'BEZ': 'Lastschrift',
 						'UEB': 'Überweisung',
 						'MCP': 'Kreditkarte',
-						'PPL': 'PayPal'
+						'PPL': 'PayPal',
+						'SUB': 'Sofortüberweisung'
```
gbirke: Hmm, this option object seems to be a copy-paste relict from `donationForm`. The only payment option for memberships is `BEZ`, with `PPL` in planning. As long as Sofortüberweisung does not support standing orders (dauerauftrag), it will never be an option for memberships, same thing with MCP. I'm torn between removing unused options for clarity and leaving them in for "future-proofing" and consistency. You decide.

app/templates/payment_type_list.twig
wiese: I'll take your word for it.

src/Presentation/TwigEnvironmentConfigurator.php
wiese: You missed [one level of inflection](https://github.com/wmde/FundraisingFrontend/pull/909/files#diff-b27bd68e2f2f9595828e438d3c9d2e24R300) - but sure, this is clearly up for a change, threw it in here for prototypical purposes.
I think
* we may have to look a the (payment) config anyways as there are differences of applicability of payment types for the different checkouts (donation, memberships) which are currently not reflected therein. Wanted to discuss that first before elaborating (potentially in the wrong direction)
* list of applicable payment types should definitely be figured out / injected into the view in/depending on the use case, not globally

web/res/js/membershipForm.js
wiese: Would love to lose what is without value, but need your knowledge and gut feeling to make the call. I'd have to understand the whole module (unrelated to this story) before going about and nuking this list. Chances are you are right, because the membership form does only show direct debit, irrespective of this list.

KaiNissen: +1 to what @gbirke says

app/ab_test/a/templates/Donation_Confirmation_Phone.html.twig
```diff
@@ -116,7 +116,7 @@
 						</div>
 
 						<div class="payment f-left">
-							{% if donation.paymentType == 'PPL' or donation.paymentType == 'MCP' %}
+							{% if donation.paymentType == 'PPL' or donation.paymentType == 'MCP' or donation.paymentType == 'SUB' %}
```
KaiNissen: This depends on whether we want the status to be reflected here. It may be `status-pledge`, `status-unconfirmed` or `status-booked`.

wiese: @KaiNissen You mean this is, given the lack of detailed status, more a case for the text simply saying "Sie zahlen per SUB" (cp. BEZ)?

app/config/config.dist.json
```diff
@@ -147,5 +147,22 @@
...
+		},
+		"MCP": {
+			"enabled": true
+		},
+		"PPL": {
+			"enabled": true
+		},
+		"SUB": {
+			"enabled": true
+		}
```
gbirke: Looking at the whole "paypal for memberships" issue where we do a "feature toggle" by delaying template changes (#901,#911), maybe we should do the feature toggle by domain (membership and donation).

app/ab_test/a/templates/Donation_Confirmation_Phone.html.twig
KaiNissen: Yes, exactly.

wiese: After some discussion w/ @KaiNissen and @gbirke decided to leave it like that until further notice (stakeholder).

wiese:
Completly contained in #910

PR #910, opened 2017-06-29
T167879 Sofort Add donation

wiese:
* sofort client facade
* refactored payment type validator to be not coupled to existing payment types (injection rather)
* (ajax) validate payment type only successfully when feature toggle says true
* support SUB

Contains all of the view/js implementation PR #909 (T167875), always rebased.
Depends on https://github.com/wmde/fundraising-frontend-content/pull/22

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -96,6 +96,7 @@ class AddDonationHandler {
 			case PaymentType::SOFORT:
 				$httpResponse = $this->app->redirect(
 					$this->ffFactory->newSofortUrlGeneratorForDonations()->generateUrl(
+						// @todo Is id the right criteria to pass? Will be put in the reason field. Cp. WQ-Code debate
 						$responseModel->getDonation()->getId(),
 						$responseModel->getDonation()->getAmount(),
 						$responseModel->getAccessToken()
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -93,6 +93,15 @@ class AddDonationHandler {
 					)
 				);
 				break;
+			case PaymentType::SOFORT:
+				$httpResponse = $this->app->redirect(
+					$this->ffFactory->newSofortUrlGeneratorForDonations()->generateUrl(
+						$responseModel->getDonation()->getId(),
+						$responseModel->getDonation()->getAmount(),
+						$responseModel->getAccessToken()
+					)
+				);
+				break;
 			case PaymentType::CREDIT_CARD:
 				$httpResponse = new Response(
 					$this->ffFactory->newCreditCardPaymentHtmlPresenter()->present( $responseModel )
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -80,7 +80,6 @@ class AddDonationHandler {
 					),
 					Response::HTTP_SEE_OTHER
 				);
-
 				break;
 			case PaymentType::PAYPAL:
 				$httpResponse = $this->app->redirect(
@@ -96,8 +95,8 @@ class AddDonationHandler {
 			case PaymentType::SOFORT:
 				$httpResponse = $this->app->redirect(
 					$this->ffFactory->newSofortUrlGeneratorForDonations()->generateUrl(
-						// @todo Is the id the right criteria to pass? Will be put in the reason field. Cp. WQ-Code debate
 						$responseModel->getDonation()->getId(),
+						$responseModel->getDonation()->getPayment()->getPaymentMethod()->getUuid(),
 						$responseModel->getDonation()->getAmount(),
 						$responseModel->getAccessToken()
 					)
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -96,7 +96,7 @@ class AddDonationHandler {
 			case PaymentType::SOFORT:
 				$httpResponse = $this->app->redirect(
 					$this->ffFactory->newSofortUrlGeneratorForDonations()->generateUrl(
-						// @todo Is id the right criteria to pass? Will be put in the reason field. Cp. WQ-Code debate
+						// @todo Is the id the right criteria to pass? Will be put in the reason field. Cp. WQ-Code debate
 						$responseModel->getDonation()->getId(),
 						$responseModel->getDonation()->getAmount(),
 						$responseModel->getAccessToken()
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
 		"symfony/browser-kit": "^3.2",
 		"silex/web-profiler": "~2.0",
 		"sorien/silex-dbal-profiler": "~2.0",
-		"wmde/fundraising-frontend-content": "dev-sofort",
+		"wmde/fundraising-frontend-content": "dev-test",
 		"symfony/css-selector": "^3.2",
 		"phpstan/phpstan": "^0.7.0"
 	},
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
 		"symfony/console": "^3.2",
 		"piwik/piwik-php-tracker": "^1.0",
 		"guzzlehttp/guzzle": "^6.0",
-		"symfony/stopwatch": "^3.3"
+		"symfony/stopwatch": "^3.3",
+		"sofort/sofortlib-php": "^3.2"
 	},
 	"repositories": [
 		{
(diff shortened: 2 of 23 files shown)


app/RouteHandlers/AddDonationHandler.php
```diff
@@ -93,6 +93,15 @@ private function newHttpResponse( AddDonationResponse $responseModel ): Response
 					)
 				);
 				break;
+			case PaymentType::SOFORT:
+				$httpResponse = $this->app->redirect(
+					$this->ffFactory->newSofortUrlGeneratorForDonations()->generateUrl(
```
gbirke: this needs error handling, since sofort API might return an error, which throws a `RuntimeError`

app/RouteHandlers/AddDonationHandler.php
```diff
@@ -93,6 +93,15 @@ private function newHttpResponse( AddDonationResponse $responseModel ): Response
...
+						$responseModel->getDonation()->getId(),
```
gbirke: For Till it was more important to have a `WQ-Code` (transfer code) here, like in the bank transfer payment. That has to be generated when the donation is persisted and then read here from the payment data. Maybe add a TODO or FIXME here, since the transfer code is not generated yet?

src/Factories/FunFunFactory.php
```diff
@@ -916,6 +919,11 @@ private function getPayPalUrlConfigForMembershipApplications() {
 		return PayPalUrlConfig::newFromConfig( $this->config['paypal-membership'] );
 	}
 
+	public function newSofortUrlGeneratorForDonations(): SofortUrlGenerator {
+		$config = SofortUrlConfig::newFromConfig( $this->config['sofort'] );
```
gbirke: I think the item name might be localized, plz ask Till about that

tests/Unit/Presentation/SofortUrlGeneratorTest.php
```diff
@@ -0,0 +1,102 @@
...
+			->willReturn( 'https://awsomepaymentprovider.tld/784trhhrf4' );
+
+		$sut = new SofortUrlGenerator( $config, $api );
+		$this->assertSame(
+			'https://awsomepaymentprovider.tld/784trhhrf4',
+			$sut->generateUrl( 529836, Euro::newFromCents( 500 ), 'letmein' )
+		);
+	}
+
+	public function testGenerateUrlApiError(): void {
```
gbirke: `testWhenApiReturnsErrorAnExceptionWithApiErrorMessage`

composer.json
```diff
@@ -29,7 +29,8 @@
 		"symfony/console": "^3.2",
 		"piwik/piwik-php-tracker": "^1.0",
 		"guzzlehttp/guzzle": "^6.0",
-		"symfony/stopwatch": "^3.3"
+		"symfony/stopwatch": "^3.3",
+		"sofort/sofortlib-php": "^3.2"
```
JeroenDeDauw: What. No evil PHP extension we need to patch to work with PHP 7? Seems like those payment providers can't even agree on a standard.

src/Presentation/SofortUrlGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+use Sofort\SofortLib\Sofortueberweisung;
+use WMDE\Euro\Euro;
+use RuntimeException;
+
+class SofortUrlGenerator {
+
+	private const CURRENCY = 'EUR';
+
+	/**
+	 * @var \WMDE\Fundraising\Frontend\Presentation\SofortUrlConfig
```
JeroenDeDauw: FQN? I'm surprise now :)

tests/Unit/Presentation/SofortUrlGeneratorTest.php
```diff
@@ -0,0 +1,102 @@
...
+			'return-url' => 'https://us.org/yes',
+			'cancel-url' => 'https://us.org/no'
+		] );
+
+		$api = $this->createMock( Sofortueberweisung::class );
+		$api
+			->expects( $this->once() )
+			->method( 'setAmount' )
+			->with( '5.00' )
+			->willReturnSelf();
```
JeroenDeDauw: This prompted me to pull this code and have a look at `Sofortueberweisung`, because it's weird to mock out non-services. Looking at the code a bit it seems reasonable enough here though.

I'm wondering if doing this via subclassing would not be clearer.

I'm also wondering if the `->expects( $this->once() )` make sense. Do you need to verify these methods only get called a specific amount of times? (See point 3 in https://www.entropywins.wtf/blog/2016/05/13/5-ways-to-write-better-mocks/)

tests/Unit/Presentation/SofortUrlGeneratorTest.php
```diff
@@ -0,0 +1,102 @@
...
+		$api
+			->expects( $this->once() )
+			->method( 'isError' )
+			->willReturn( true );
+		$api
+			->expects( $this->once() )
+			->method( 'getError' )
+			->willReturn( 'boo boo' );
+
+		$sut = new SofortUrlGenerator( $config, $api );
```
JeroenDeDauw: I think it is clearer to call this `$urlGenerator`. Not sure what the others in the team think about this.

src/Presentation/SofortUrlGenerator.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @var \Sofort\SofortLib\Sofortueberweisung
+	 */
+	private $api;
+
+	public function __construct( SofortUrlConfig $config, Sofortueberweisung $api ) {
+		$this->config = $config;
+		$this->api = $api;
+	}
+
+	public function generateUrl( string $itemId, Euro $amount, string $accessToken ): string {
```
wiese: Chances are, we only need one id to id a donation. As this offers us an interface to pass an identifying criteria into the generator it should be fine. The using side may need a change, [as you correctly indicated there](https://github.com/wmde/FundraisingFrontend/pull/910#discussion_r124843683) (i.e. use of a different donation method to generate an _id_)

app/RouteHandlers/AddDonationHandler.php
wiese: We _do_ have global exception handling. What would be your suggestion how we handle said exception? We fail to produce an URL we wanted to use to redirect her/him, would need an alternative user flow. It's really a more a question of UX than technicality.

app/RouteHandlers/AddDonationHandler.php
wiese: Done.

wiese: Now using the bank transfer code / W-Q code as sofort reason text

src/Infrastructure/Sofort/Transfer/Request.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @return Euro
+	 */
+	public function getAmount(): Euro {
+		return $this->amount;
+	}
+
+	/**
+	 * @param Euro $amount
+	 */
+	public function setAmount( Euro $amount ): void {
```
JeroenDeDauw: These comments are not adding any value here

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -345,6 +347,29 @@ public function testGivenValidCreditCardData_showsIframeEmbeddingPage() {
 		$this->assertContains( 'thatother.paymentprovider.com', $response->getContent() );
 	}
 
+	public function testValidSofortInput_RedirectsTo3rdPartyPage() {
```
JeroenDeDauw: The naming scheme we follow has a lowercase letter after the underscore

JeroenDeDauw:
Don't see anything really bad... need to look at this more though

gbirke:
I'm now fine with the UrlGenerator being in the Presentation namespace:

* We aren't storing any additional sofort data in the AddDonation use case where we could retrieve it at the presentation layer
* Generating the URL in the use case would be misplaced, as it's not part of the Domain.
* The API is now behind a service facade that masks the fact that an HTTP request is happening.

We *could* move all the URL generators to the `Infrastructure\UrlGenerator` namespace, but I have no strong feelings about this.

gbirke:
+2

src/Infrastructure/Sofort/Transfer/Request.php
wiese: They are not really, but they are part of my workflow for value objects - specifying the properties and using PHPStorm to auto-generate the accessors.
You could argue that their presence increases the chance for inconsistency whenever either the method signature or the phpdoc change independently of the other.. do you feel strongly about this?
Given - I believe - there is no real rule not to have them, what is the consensus on this?
@gbirke @KaiNissen

gbirke: I always remove the autogenerated comments in PHPStorm after generating the accessors.

KaiNissen: Me, too. In my opinion there should be either type hints or phpdoc comments (to avoid inconsistency and/or redundance). I would also prefer type hints over phpdoc except there are reasons not to use them.

wiese:
Updated where "reason texts" come from.

PR #913, opened 2017-07-07
Remove old wikimedia logo

gbirke:
This for https://phabricator.wikimedia.org/T168725


gbirke:
There might be another logo that can be removed - class `logo-wikimedia-grey` and corresponding image file. Waiting for comments ...

PR #917, opened 2017-07-12
T167886 sofort bank transfer code

wiese:
* assigning bank transfer code on donation for sofort, too
* setting sofort donation initial status to PROMISE

~~Composer needs to be pointed to a stable version of Store before merge.~~


wiese:
Using store's master HEAD, still needs switch to pin.

contexts/DonationContext/tests/Unit/DataAccess/DoctrineDonationRepositoryTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+		$method = new ReflectionMethod( $repo, 'getBankTransferCode' );
+		$method->setAccessible( true );
+
+		$paymentMethod = new SofortPayment( 'fff' );
+		$this->assertSame( 'fff', $method->invoke( $repo, $paymentMethod ) );
+		$paymentMethod = new BankTransferPayment( 'ggg' );
+		$this->assertSame( 'ggg', $method->invoke( $repo, $paymentMethod ) );
+		$paymentMethod = new CreditCardPayment( null );
+		$this->assertSame( '', $method->invoke( $repo, $paymentMethod ) );
```
JeroenDeDauw: I would much rather have a test that checks if a SofortPayment gets stored correctly, which will catch the not setting of the bank transfer code as well.

wiese:
Updated.

JeroenDeDauw:
I think this is linking the wrong issue, since T167886 is just about the store. Not finding a fitting one though

JeroenDeDauw:
ugggh need to also update th lock file... can't easily do that on this box

wiese:
Use stable version of the store (composer.json & composer.lock)

PR #918, opened 2017-07-12
T167879 sofort donation state

wiese:
New sofort donations are supposed do be in promised state. Cleared up confusion in stories (said X in some places).


contexts/DonationContext/tests/Unit/UseCases/AddDonation/AddDonationUseCaseTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\DonationContext\UseCases\AddDonation\AddDonationUseCase
+ */
+class AddDonationUseCaseTest extends TestCase {
+
+	public function testGetInitialDonationStatus(): void {
+		$useCase = $this->createMock( AddDonationUseCase::class );
+		$method = new ReflectionMethod( $useCase, 'getInitialDonationStatus' );
+		$method->setAccessible( true );
```
JeroenDeDauw: Reflection like this undermines encapsulation and breaks automated refactoring. Do you think this is somehow still better than what we've done so far for testing UC code?

contexts/DonationContext/tests/Unit/UseCases/AddDonation/AddDonationUseCaseTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+		$this->assertEquals( 'N', $method->invoke( $useCase, 'BEZ' ) );
+		$this->assertEquals( 'Z', $method->invoke( $useCase, 'UEB' ) );
+		$this->assertEquals( 'Z', $method->invoke( $useCase, 'SUB' ) );
+
+		$this->assertEquals( 'X', $method->invoke( $useCase, 'MCP' ) );
+		$this->assertEquals( 'X', $method->invoke( $useCase, 'foo' ) );
```
JeroenDeDauw: assertSame possible

contexts/DonationContext/tests/Unit/UseCases/AddDonation/AddDonationUseCaseTest.php
wiese: I do not think it is _better_, but another layer of security (the first actual unit test for this class). All this while the code is great overall and the existing tests are sweet.
The disadvantage of testing private methods this way is evident, but the flipside of having a class of 200 lines with only 2 public methods (which I do not oppose, just claim there is a price). I'd much rather have a unit test know the internals, the "how", of a class and have complete coverage of a method than having to craft n permutations of dummy object (we still do that _additionally_ for a few samples in integration, ...) to get anywhere close to the same level of confidence.
Concerning the "breaks automated refactoring": I'd actually want this test to break then.

Could that work for you?

JeroenDeDauw: It is not an approach I am happy with. If I rename a private method or somehow change the internal structure of a class, I don't want tests to break.

In case where dealing with the whole class is too cumbersome I prefer refactoring out the logic into a new class that can then be tested easier. And if the code is too trivial for that, then having a public static method (for functions that do not use state or have side effects) seems like a better trade off than using reflection to me.

wiese:
Updated.

PR #919, opened 2017-07-13
use @requires annotation to check for extension

wiese:
Use @requires annotation instead of code to optionally skip tests that could not run w/o.


JeroenDeDauw:
Cool

PR #920, opened 2017-07-14
phpstan all around

wiese:
* running phpstan as part of "normal" CI (`tests/` folders, too, as @JeroenDeDauw mentioned correctly upon first introduction) & against prod code
* fixed many namespace inconsistencies in tests
* rm some excess constructor parameters
* mended missing parent::__constructor calls (nothing to be proud of)
* mv mail template config into a dedicated class
* mv mail template config class into `app`. Can't be in `tests/` if used by non-test CLI.
* fixed konto_check madness (documented in [T170913](https://phabricator.wikimedia.org/T170913))


gbirke:
I'd place the `MailTemplates` in `app/config`, as it's purely configuration data.

wiese:
> I'd place the MailTemplates in app/config, as it's purely configuration data.

Thanks, @gbirke. Makes for a nice NS, `WMDE\Fundraising\Frontend\App\Config\MailTemplates`. @JeroenDeDauw Would that be ok with you?

JeroenDeDauw:
I dunno about the mail templates. Moving it as suggested sounds plausible to me. If you want I can look at it and think about it more.

I do suggest trying to merge these changes ASAP, because mawrg conflicts are maw. You could disable the new CS rules so the test pass and have those changes merged already and then do one or more follow ups that make them pass.

wiese:
Rebased and fixed newly introduced problems

PR #921, opened 2017-07-19
T167882 Sofort notification + mail

wiese:
* created endpoint for sofort payment notifications
* created sofort donation confirmation mail, triggered on sofort payment notification
* pimped FakeDonationRepository so that objects stored inside of them are identical before & after
  * asserting donation passed to mailer in case of credit card & paypal, too
* abstracted `assertEventLogContainsExpression` into trait (not used for sofort in the end, but kept)

Need to roll this to test (with www-reachable endpoint) to test integration with actual payment provider.

### DIFF
--- a/app/RouteHandlers/SofortNotificationHandler.php
+++ b/app/RouteHandlers/SofortNotificationHandler.php
@@ -37,15 +37,17 @@ class SofortNotificationHandler {
 	}
 
 	private function newUseCaseRequestFromRequest( Request $request ): SofortNotificationRequest {
-		$usecaseRequest = new SofortNotificationRequest();
-		$usecaseRequest->setDonationId( $request->query->getInt( 'id' ) );
-		$usecaseRequest->setTransactionId( $request->request->get( 'transaction', '' ) );
+		$useCaseRequest = new SofortNotificationRequest();
+		$useCaseRequest->setDonationId( $request->query->getInt( 'id' ) );
+		$useCaseRequest->setTransactionId( $request->request->get( 'transaction', '' ) );
+
 		$time = DateTime::createFromFormat( DateTime::ATOM, $request->request->get( 'time', '' ) );
+
 		if ( $time instanceof DateTime ) {
-			$usecaseRequest->setTime( $time );
+			$useCaseRequest->setTime( $time );
 		}
 
-		return $usecaseRequest;
+		return $useCaseRequest;
 	}
 
 	private function logResponseIfNeeded( SofortNotificationResponse $response, Request $request ) {
--- a/app/RouteHandlers/SofortNotificationHandler.php
+++ b/app/RouteHandlers/SofortNotificationHandler.php
@@ -1,0 +1,87 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\RouteHandlers;
+
+use Psr\Log\LogLevel;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Euro\Euro;
+use WMDE\Fundraising\Frontend\PaymentContext\ResponseModel\PaypalNotificationResponse;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+use WMDE\Fundraising\Frontend\Infrastructure\PayPalPaymentNotificationVerifierException;
+use WMDE\Fundraising\Frontend\PaymentContext\RequestModel\SofortPaymentNotificationRequest;
+use DateTime;
+
+class SofortNotificationHandler {
+
+	private $ffFactory;
+
+	public function __construct( FunFunFactory $ffFactory ) {
+		$this->ffFactory = $ffFactory;
+	}
+
+	public function handle( Request $request ): Response {
+		$post = $request->request;
+
+		try {
+			$this->ffFactory->getPayPalPaymentNotificationVerifier()->verify( $post->all() );
+		} catch ( PayPalPaymentNotificationVerifierException $e ) {
+			$this->ffFactory->getPaypalLogger()->log( LogLevel::ERROR, $e->getMessage(), [
+				'post_vars' => $request->request->all()
+			] );
+			return $this->createErrorResponse( $e );
+		}
+
+		$useCase = $this->ffFactory->newHandlePayPalPaymentNotificationUseCase( $this->getUpdateToken( $post ) );
+
+		$response = $useCase->handleNotification( $this->newUseCaseRequestFromPost( $post ) );
+		$this->logResponseIfNeeded( $response, $request );
+
+		return new Response( '', Response::HTTP_OK ); # PayPal expects an empty response
+	}
+
+	private function getUpdateToken( ParameterBag $postRequest ): string {
+		return $this->getValueFromCustomVars( $postRequest->get( 'custom', '' ), 'utoken' );
+	}
+
+	private function getValueFromCustomVars( string $customVars, string $key ): string {
+		$vars = json_decode( $customVars, true );
+		return !empty( $vars[$key] ) ? $vars[$key] : '';
+	}
+
+	private function newUseCaseRequestFromPost( ParameterBag $postRequest ): SofortPaymentNotificationRequest {
+		return ( new SofortPaymentNotificationRequest() )
+			->setTransactionId( $postRequest->get( 'transaction', '' ) )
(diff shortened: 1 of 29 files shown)


app/RouteHandlers/SofortNotificationHandler.php
```diff
@@ -0,0 +1,68 @@
...
+	public function handle( Request $request ): Response {
+		$useCase = $this->ffFactory->newHandleSofortPaymentNotificationUseCase( $request->query->get( 'updateToken' ) );
+
+		$response = $useCase->handleNotification( $this->newUseCaseRequestFromRequest( $request ) );
+
+		$this->logResponseIfNeeded( $response, $request );
+
+		if ( !$response->notificationWasHandled() ) {
+			return new Response( 'Bad request', Response::HTTP_BAD_REQUEST );
+		} elseif ( $response->hasErrors() ) {
```
gbirke: Please check for Errors first, otherwise this statement will never be reached, as the error case also is unhandled.

contexts/DonationContext/tests/Integration/UseCases/SofortPaymentNotification/SofortPaymentNotificationUseCaseTest.php
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\DonationContext\UseCases\SofortPaymentNotification\SofortPaymentNotificationUseCase
+ */
+class SofortPaymentNotificationUseCaseTest extends TestCase {
+
+	public function testWhenRepositoryThrowsException_errorResponseIsReturned(): void {
+		$useCase = new SofortPaymentNotificationUseCase(
+			new DoctrineDonationRepository( ThrowingEntityManager::newInstance( $this ) ),
+			new FailingDonationAuthorizer(),
```
gbirke: Maybe use a succeeding authorizer to avoid false positives?

contexts/DonationContext/tests/Integration/UseCases/SofortPaymentNotification/SofortPaymentNotificationUseCaseTest.php
```diff
@@ -0,0 +1,218 @@
...
+
+		$reponse = $useCase->handleNotification( $request );
+		$this->assertFalse( $reponse->notificationWasHandled() );
+		$this->assertTrue( $reponse->hasErrors() );
+	}
+
+	public function testWhenNotificationIsForNonExistingDonation_unhandledResponseIsReturned(): void {
+		$useCase = new SofortPaymentNotificationUseCase(
+			new FakeDonationRepository(),
+			new FailingDonationAuthorizer(),
```
gbirke: Again, authorizer should succeed theoretically.

app/RouteHandlers/SofortNotificationHandler.php
JeroenDeDauw: Done by https://github.com/wmde/FundraisingFrontend/pull/922

contexts/DonationContext/tests/Integration/UseCases/SofortPaymentNotification/SofortPaymentNotificationUseCaseTest.php
JeroenDeDauw: Nice catch. I've changed this on this branch itself

contexts/DonationContext/tests/Integration/UseCases/SofortPaymentNotification/SofortPaymentNotificationUseCaseTest.php
JeroenDeDauw: Nice catch. I've changed this on this branch itself

JeroenDeDauw:
I've taken the liberty to push some minor CS changes onto this branch and address the SucceedingDonationAuthorizer remark here as well. The error stuff change I did on a PR that merges into this branch: https://github.com/wmde/FundraisingFrontend/pull/922

+2 from my side after that is merged

app/RouteHandlers/SofortNotificationHandler.php
```diff
@@ -0,0 +1,70 @@
...
+	private function newUseCaseRequestFromRequest( Request $request ): SofortNotificationRequest {
+		$useCaseRequest = new SofortNotificationRequest();
+		$useCaseRequest->setDonationId( $request->query->getInt( 'id' ) );
+		$useCaseRequest->setTransactionId( $request->request->get( 'transaction', '' ) );
+
+		$time = DateTime::createFromFormat( DateTime::ATOM, $request->request->get( 'time', '' ) );
+
+		if ( $time instanceof DateTime ) {
+			$useCaseRequest->setTime( $time );
+		}
```
JeroenDeDauw: @gbirke told me about this and that the error should probably be handled here and not in the UC. I fully agree with this, as what we are catching here is a parsing error, which is presentation specific. However this PR is already quite big, has been sitting here for multiple days and it is quite annoying to have to deal with the whole thing if only small stuff gets changed. So I suggest we merge as is and do a follow up. I'm happy to make that follow up and will do so if no one beats me to it. This will include having a good look at the logging as well, which I think does not belong in this Handler, and could perhaps be done in a way that makes it easier to test (no e2e needed).

JeroenDeDauw:
maw.gif for no review having happened yet!

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
```diff
@@ -0,0 +1,211 @@
...
+
+	public function testGivenWrongPaymentType_applicationRefuses(): void {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ) {
+			$factory->setTokenGenerator( new FixedTokenGenerator(
+				self::TOKEN_VALID,
+				new DateTime( '2039-12-31 23:59:59Z' )
+			) );
+
+			$repo = $factory->getDonationRepository();
+			$repo->storeDonation( ValidDonation::newIncompletePayPalDonation() );	// creates donation w/ id=1
```
JeroenDeDauw: You don't need to do this and then have a magical 1 further down. You can just get the id from the donation.

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
```diff
@@ -0,0 +1,211 @@
...
+				'99999-53245-5483-4891',
+				$logger->getLogCalls()->getFirstCall()->getContext()['post_vars']['transaction']
+			);
+
+			$this->assertSame(
+				'1',
+				$logger->getLogCalls()->getFirstCall()->getContext()['query_vars']['id']
+			);
+
+			$this->assertSame( LogLevel::INFO, $logger->getLogCalls()->getFirstCall()->getLevel() );
```
JeroenDeDauw: This is a LOOOOONG method. Looks like some of the assertions can be extracted together

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
```diff
@@ -0,0 +1,211 @@
...
+					'time' => '2010-04-14T19:01:08+02:00'
+				]
+			);
+
+			$this->assertSame( 'Bad request', $client->getResponse()->getContent() );
+			$this->assertSame( Response::HTTP_BAD_REQUEST, $client->getResponse()->getStatusCode() );
+
+			$this->assertSame(
+				[ 'Duplicate notification' ],
+				$logger->getLogCalls()->getMessages()
```
JeroenDeDauw: This is very brittle as it assumes nothing else gets logged to the system wide logger.

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
JeroenDeDauw: Like this: https://github.com/wmde/FundraisingFrontend/pull/923/commits/060f8abcbaab0b610f8dcac8d3af127a0d53c12b

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
JeroenDeDauw: Like in https://github.com/wmde/FundraisingFrontend/pull/923

tests/EdgeToEdge/Routes/SofortPaymentNotificationRouteTest.php
JeroenDeDauw: I suppose the only way to avoid this is to write more complicated assertions. All the more reason to not have to do this in e2e tests

PR #923, opened 2017-07-23
Misc imporvements to SofortPaymentNotification code

JeroenDeDauw:
Follow up to https://github.com/wmde/FundraisingFrontend/pull/921

### DIFF
--- a/app/RouteHandlers/SofortNotificationHandler.php
+++ b/app/RouteHandlers/SofortNotificationHandler.php
@@ -8,6 +8,7 @@ use DateTime;
 use Psr\Log\LogLevel;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\Frontend\DonationContext\UseCases\SofortPaymentNotification\SofortPaymentNotificationUseCase;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\PaymentContext\RequestModel\SofortNotificationRequest;
 use WMDE\Fundraising\Frontend\PaymentContext\ResponseModel\SofortNotificationResponse;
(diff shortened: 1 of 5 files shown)


app/RouteHandlers/SofortNotificationHandler.php
```diff
@@ -16,16 +17,28 @@ class SofortNotificationHandler {
...
+	 */
+	private $request;
+
 	public function __construct( FunFunFactory $ffFactory ) {
 		$this->ffFactory = $ffFactory;
 	}
 
 	public function handle( Request $request ): Response {
-		$useCase = $this->ffFactory->newHandleSofortPaymentNotificationUseCase( $request->query->get( 'updateToken' ) );
+		$this->request = $request;
```
JeroenDeDauw: oh no! Wider scope and not needed mutation!

As this is (1) a class with only a single public method, (2) the only write happens at the start of that method and (3) the value is needed all over the place, I'm opting for this approach. Otherwise you end up with an extra parameter in all methods. (It is also possible to introduce a new class, but that seems premature.)

JeroenDeDauw:
Tests passing for all commits btw... Scrutinizer is just being derp

app/RouteHandlers/SofortNotificationHandler.php
gbirke: I think it's fine.

app/RouteHandlers/SofortNotificationHandler.php
```diff
@@ -16,16 +17,28 @@ class SofortNotificationHandler {
...
+
+		$useCaseRequest = $this->newUseCaseRequest();
+
+		if ( $useCaseRequest === null ) {
+			$this->logWebRequest( [ 'message' => 'Bad notification time' ], LogLevel::ERROR );
+			return new Response( 'Bad request', Response::HTTP_BAD_REQUEST );
```
gbirke: In the (admittedly unlikely) case that we have another validation in `newUseCaseRequest`, this will lead to a misleading error message. Could you use an exception and try...catch instead?

JeroenDeDauw: Yeah I was also wondering about this... always a pain when you factor out code that returns and has guard clauses like this... The exception case seems a bit dodgy but if you think it is better than what I did by all means refactor.

PR #925, opened 2017-07-25
add notice for membership applications using paypal

KaiNissen:
This PR is for displaying a notice in the membership application form when using PayPal. It depends on the message added by wmde/fundraising-frontend-content#24.

[phab:T168721](https://phabricator.wikimedia.org/T168721)


JeroenDeDauw:
Why did you put this payment delay days in the payment delay calculator? Looks like you can simply get at the config and not bother with this service at all.

PR #928, opened 2017-08-03
T167882 Sofort notification followup

wiese:
Discovered that the data received [is an XML](https://phabricator.wikimedia.org/T167882#3493855) clob that needs special treatment, so
* use 3rd party class to handle incoming notification requests
* dumping request body instead of POST to log in case of invalid requests

If you do not have strong objections, we should first try (can't do on dev) on test and then polish.

### DIFF
--- a/contexts/PaymentContext/src/RequestModel/SofortNotificationRequest.php
+++ b/contexts/PaymentContext/src/RequestModel/SofortNotificationRequest.php
@@ -4,6 +4,7 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\PaymentContext\RequestModel;
 
+use Sofort\SofortLib\Notification as VendorNotification;
 use DateTime;
 
 class SofortNotificationRequest {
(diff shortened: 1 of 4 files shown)


contexts/PaymentContext/src/RequestModel/SofortNotificationRequest.php
```diff
@@ -47,4 +48,25 @@ public function setTime( DateTime $time ): self {
 		$this->time = $time;
 		return $this;
 	}
+
+	public static function fromRawRequest( string $content ): ?self {
```
JeroenDeDauw: The request models are not supposed to contain anything presentation specific and be dumb value objects (no logic). You can argue that since this is a static method, it is not really part of the class of course... Even so I really rather not have this web request parsing code here. It adds web-bound stuff to files that define things that are delivery mechanism agnostic. If you somehow want to enforce the boundaries (or move the Domain bound stuff into its own git repo), this will create problems.

JeroenDeDauw:
Will change in follow up

PR #932, opened 2017-08-07
Make CC payment redirect based instead of iframe

JeroenDeDauw:
For https://phabricator.wikimedia.org/T170171

This should probably be manually tested by someone that has the
system level integration with the CC thing set up before merging.

Also perhaps this should not be deployed before more stuff is done as per Gabriels comment at https://phabricator.wikimedia.org/T170171#3506330


tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -357,8 +358,9 @@ public function testGivenValidCreditCardData_showsIframeEmbeddingPage(): void {
 		);
 
 		$response = $client->getResponse();
-		$this->assertSame( 200, $response->getStatusCode() );
-		$this->assertContains( 'paytext_cc 3 12,34 € per Kreditkarte.', $response->getContent() );
+		$this->assertSame( Response::HTTP_FOUND, $response->getStatusCode() );
+		$this->assertContains( 'amount=1234', $response->getContent() );
+		$this->assertContains( 'Redirecting to', $response->getContent() );
```
wiese: I think it would be worthwhile to assert what we actually want to happen - the location header on which the redirect will be based, not the fallback response body. Agree?

    $client->followRedirects( false );
    $client->request( ... );
    $response = $client->getResponse();
    $this->assertTrue( $response->isRedirect() );
    $this->assertContains( 'something.dot.somewhere', $response->headers->get( 'Location' ) );

src/Presentation/Presenters/CreditCardPaymentUrlGenerator.php
```diff
@@ -0,0 +1,41 @@
...
+
+	public function __construct( TranslatorInterface $translator, CreditCard $urlGenerator ) {
+		$this->translator = $translator;
+		$this->urlGenerator = $urlGenerator;
+	}
+
+	public function buildUrl( AddDonationResponse $response ): string {
+		$personalInfo = $response->getDonation()->getDonor();
+
+		return $this->urlGenerator->generateUrl(
```
wiese: Don't we have to pass the return-url to allow them to find their way back to our site?
Currently this is (somehow) addressed in `FunFunFactory::newCreditCardNotificationPresenter()` but probably will have to be adapted as well for the user journey to work.

KaiNissen: They hid it well... Their payment notifier expects the notification handler to provide the return URL (just as we do it now). There is no other way to define this.

(see [example script](https://www.micropayment.de/redirector/?to=download&what=example-cc-evt))

wiese: Then let's hope for the best and try it on test...

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
JeroenDeDauw: Yeah, this is much better

JeroenDeDauw:
@KaiNissen @gbirke Could one of you merge this and make sure stuff does not blow up (ie reverting or fixing if there is a problem)?

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
```diff
@@ -357,9 +358,10 @@ public function testGivenValidCreditCardData_showsIframeEmbeddingPage(): void {
 		);
 
 		$response = $client->getResponse();
-		$this->assertSame( 200, $response->getStatusCode() );
-		$this->assertContains( 'paytext_cc 3 12,34 € per Kreditkarte.', $response->getContent() );
-		$this->assertContains( 'thatother.paymentprovider.com', $response->getContent() );
+		$this->assertSame( Response::HTTP_FOUND, $response->getStatusCode() );
+		$this->assertTrue( $response->isRedirect() );
+		$this->assertContains( 'amount=1234', $response->getContent() );
```
gbirke: Shouldn't the content of a redirect response be empty?

wiese: I believe this is some fancy pants behaviour of symfony http response which adds a response body (which is not strictly forbidden as per RFC) with a link, maybe meta redirect, for clients not able/allowed to follow location headers... I anyway think we should properly assert the header, not the content.

JeroenDeDauw:
> I just deployed this branch to test and donated via credit card.

image

> we need to figure out first how

I am confoose now. Does this mean this PR should not be merged? It has two approvals now, which to me means I can be merged, but then the comments don't really match.

tests/EdgeToEdge/Routes/AddDonationRouteTest.php
JeroenDeDauw: Done in https://github.com/wmde/FundraisingFrontend/pull/935

gbirke:
We still need to get an approval from the FR people. This is tracked at https://phabricator.wikimedia.org/T172709

If you fear that takes too long, we could add a feature toggle to this branch and merge it.

JeroenDeDauw:
giphy

PR #938, opened 2017-08-11
Extract Donation Context

JeroenDeDauw:
This extracts the Donation Context and most of the Payment Context
into https://github.com/wmde/fundraising-donations

A bunch of infrastructure code and random stuff also had to be
extracted as it was dependend upon by the former code. Some of
this should be put in a nicer place if we continue with this
approach, and some of it should be fixed dependency wise regardless
if we continue or not (though depends on that decision for how to).

This code can be found at https://github.com/wmde/fundraising-donations/tree/master/todo

All files where just moved and not altered. This includes no changes
in their namespace. Namespace changes likely make sense though if
done at present would get in the way of easy revert.

There is one exception: IbanValidator was renamed to KontoCheckIbanValidator
and now implements a new IbanValidator intetface defined by the new component.
This was needed to avoid coupling to the evil konto check stuff in the new component.

---

I have checked all the PRs that are open and created since I copied over the files. The already merged ones are not touching the moved code (except one which I mirrored). Of the 3 open ones only https://github.com/wmde/FundraisingFrontend/pull/935/ touches the moved code. If it gets merged first this PR will need updating.

Out of the 6 PRs I looked at only one touched files on both sides of the new git repo "boundary" (as messy as it is atm...), which is encouraging.


Vagrantfile
```diff
@@ -13,7 +13,7 @@ Vagrant.configure(2) do |config|
   # the path on the host to the actual folder. The second argument is
   # the path on the guest to mount the folder. And the optional third
   # argument is a set of non-required options.
-  # config.vm.synced_folder "../data", "/vagrant_data"
+  config.vm.synced_folder "../FundraisingDonations", "/fundraising-donations"
```
wiese: Except for actual package management (composer file), I think we should leave it to the individual dev to sync the context into the main project - it's bound to break if it references locations on disk

JeroenDeDauw: Oh sure, I just put this in here before I created the repo on GH to try things out. Forgot to remove it.

JeroenDeDauw: Gone now

wiese:
Hi @JeroenDeDauw,
thanks a lot for the amount of work you put into this. Especially identifying the boundary violations (dependencies from to contexts, and from contexts to the main app) in the `todo` folder must have been a fun game and are valuable whatever we decide to do next.

I want to emphasize that this change is complex, has many aspects to it, and discussing all this in a nuanced fashion in written is probably not ideal. I'll try to set the stage for a discussion anyways.

The task was needed to get to the bottom of what still has to be done, but I wonder whether actually merging this into the main line just yet will bring us the best possible take away.
Some of the identified tasks are all but trivial...
* should we have a dedicated Validator repo or do we switch to a 3rd party library?
* where will the EmailAddress end up, what/how much can be reused? Should mailing in general be looked at when separating Donation & Membership?
* how do we share util code accross contexts/the app (e.g. FreezableValueObject)?

Going down the road of having these open questions inside fundraising-donations' (whilst letting it "write" into any namespace at will)
* puts a huge technical debt in the heart of our application just before we are faced with challenges like design-relaunch and campaign madness, **w/o a clear timeline or stories to resolve it**
* makes it **hard to navigate the projects**, know where one can expect a file
* closes the door to further separation of contexts (e.g. payment) w/o code duplication

As an alternative proposal I suggest we take the todos on a challenge to solve step-by-step _in the FF project_ and celebrate the day when we can move the context folders, maybe the pseudo-context payment first, into clean, isolated, dedicated repos.

JeroenDeDauw:
robmamk

JeroenDeDauw:
As per discussion yesterday, we will give this a try

PR #939, opened 2017-08-14
Introduce payment type toggle for membership applications

KaiNissen:


src/Factories/FunFunFactory.php
```diff
@@ -1011,7 +1011,13 @@ public function getTokenGenerator(): TokenGenerator {
 
 	public function newDonationConfirmationPresenter(): DonationConfirmationHtmlPresenter {
 		return new DonationConfirmationHtmlPresenter(
-			$this->getIncludeTemplate( 'Donation_Confirmation.html.twig', [ 'piwikGoals' => [ 3 ] ] )
+			$this->getIncludeTemplate(
```
gbirke: `getIncludeTemplate` is deprecated since we're no longer using content form the Wiki. The recommended way is to change the template that gets included to extend the base layout and use `getLayoutTemplate`.

app/templates/Membership_Application_Form.html.twig
```diff
@@ -245,6 +245,7 @@
 
 					<div class="container clearfix">
 
+{% if 'PPL' in paymentTypes %}
```
wiese: Does this exploit that `BEZ` is still the default value if no `payment_type` is received via the form? Sure is a quick way to get this back to state it was before the introduction of `PPL`, but looks confusing (disabling BEZ if PPL is not allowed)...

KaiNissen: Right. I added another commit that allows any combination of payment types and cares for not displaying the list of options if there is only one.

PR #940, opened 2017-08-15
Added transfer code generator with checksum

JeroenDeDauw:
First step towards https://phabricator.wikimedia.org/T170266 (task is https://phabricator.wikimedia.org/T170416)
    
    Has checkstum and 8 characeters instead of 7 to add some extra entropy.
    Like this we have 150 billion combinations, which should give us enough
    room for adding a bad word filter.
    
    Added a dash in the center of the thing for readability.
    
    Did not include the prefix in this class as it is not clear to me if
    it should go in there or not. If it depends on something else, such
    as the type of donation, then probably not.


gbirke:
This goes a bit against the goal of the codes being "as short as possible" (I've made a comment in the issue). Also, using two digits of a md5 checksum makes it too long and might introduce errors since we're truncating the MD5 and the characters might be ambigous. I have a different recommendation for the algorithm:

* Generate a random 7-character code from our character pool (Using your code, we'd need an iterator for `$characterSource` that generates random characters instead of just repeating what was put in).
* Generate a 4-bit checksum from the code, using a checksum algorithm that protects against wrong characters and wrong order of characters. 4 bit because that's the maximum number that can be displayed with our 22 chars. 
* Encode the checksum with our character set (like `base_convert`, but with our character set instead of the ambiguous characters).

What do you think, @KaiNissen @wiese ?

JeroenDeDauw:
Ok, so need `ACDEFKLMNPRSTWXYZ349` as character pool. 4-bit checksum? do you mean 4-byte?

gbirke:
Nope, 4 Bit. The checksum should be displayed as *one* character from our pool (now that we've lost `S` and `5` its no longer base22 but base20). That means we can display base10 numbers between 0 and 20 and [we need log2(20) bits ](https://stackoverflow.com/questions/12269096/how-many-bits-do-you-need-to-store-a-number) to display them.  The formula from the StackOverflow answer would give us 5 bits, but then we could not display checksums between 23 and 32 in one character, so the correct length is 4 bits.

gbirke:
I've just checked the behavior of `strtr`, it transforms characters only once:

    assert(strtr( "XXABXX", "AB", "BC") === "XXBCXX"); // NOT XXCCXX

Which means we can easily convert our checksum (base10 number) like this:

    $checksum = 5; // integer < 16, coming from some algorithm 
    $checksumChar = strtr( 
         base_convert( $checksum, 10, 20 ), 
         '0123456789ABCDEFGHIJ', 
         'ACDEFKLMNPRSTWXYZ349'
    );

JeroenDeDauw:
That code took me some time to understand, including looking up what those functions do. As far as I can tell it is equivalent to

```php
$checksum = 5; // integer < 16, coming from some algorithm 
$checksumChar = str_split('ACDEFKLMNPRSTWXYZ349')[$checksum];
```

For the algorithm, how about we md5 the thing, then add up all character ASCII values and get the remainder from division by 16? That's the simplest way I can think of to get the needed value and catch things such as position changes on chars. Then again this stuff is really not what I'm used to.

---

As to the "as short as possible to avoid errors": codes such as these often have dashes in them so they are easier to read and it thus being less likely you make errors. KJLHNGFE is harder to read correctly than KJLH-NGFE to most people I figure.

JeroenDeDauw:
I've now updated the code to follow the approach I outlined in my previous comment.

contexts/PaymentContext/tests/Unit/Domain/ChecksumGeneratorTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+		for ( $i = 0; $i < 1000; $i++ ) {
+			yield implode(
+				'',
+				array_map(
+					function() use ( $characters, $characterCount ) {
+						return $characters[mt_rand( 0, $characterCount - 1 )];
+					},
+					array_fill( 0, 10, null )
+				)
```
JeroenDeDauw: https://stackoverflow.com/questions/46308939/creating-a-php-array-with-values-from-a-function

contexts/PaymentContext/tests/Unit/Domain/ChecksumGeneratorTest.php
```diff
@@ -0,0 +1,67 @@
...
+	public function getRandomStrings(): iterable {
+		$characters = str_split( 'ACDEFKLMNPRSTWXYZ349-' );
+		$characterCount = count( $characters );
+
+		for ( $i = 0; $i < 1000; $i++ ) {
+			yield implode(
+				'',
+				array_map(
+					function() use ( $characters, $characterCount ) {
+						return $characters[mt_rand( 0, $characterCount - 1 )];
```
JeroenDeDauw: I looked into removing the randomness from the tests by providing a seed. Did not find anything that would not mutate global PHP state, so opted to stick with the randomness here, which in this case does not seem too evil (certainly less than the global mutation).

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -0,0 +1,43 @@
...
+	 */
+	public function createChecksum( string $string ): string {
+		$checksum = array_sum(
+			array_map(
+				'ord', // Returns the ASCII value of the character
+				str_split( md5( $string ) )
+			)
+		);
+
+		return $this->checksumCharacters[$checksum % count( $this->checksumCharacters )];
```
gbirke: This is brilliant because it makes the character selection independent from the bit-length of the checksum.

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @param string $string The string to create a checksum for
+	 *
+	 * @return string The checksum as a single character present in the constructors array argument
+	 */
+	public function createChecksum( string $string ): string {
+		$checksum = array_sum(
+			array_map(
+				'ord', // Returns the ASCII value of the character
+				str_split( md5( $string ) )
+			)
```
gbirke: This will produce a very limited range of low numbers between 1563 ( `ord('0') * 32` ) and 2496 ( `ord('F') * 32` ), equating a range between 0-960, [that's about 10 bits](https://stackoverflow.com/a/23400205/130121). The more bits, the more accurate the checksum. How about 

    base_convert( substr( md5( $string ), 0, 16 ), 16, 10 );

the `substr` will "convert" the 32-bit hash into a 16-bit hash, `base_convert` converts the hex string into an integer between 0 and 18446744073709552046

JeroenDeDauw: Does that matter if we `% 16` it later anyway? I checked the distribution created for the checksum and it was pretty even.

JeroenDeDauw:
So we filter out the dashes so that `AAAAAAAAA`, `AAAAAAAA-A` and `A--A--A--A-A-----AAA-A` are all valid forms of `AAAA-AAAA-A`?

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
JeroenDeDauw: Do I get an achievement card?

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
gbirke: It probably does not matter. However, as this algorithm has to be implemented elsewhere in some other language and documented, I'd like it to be as "mathematical" as possible and not to rely on string operations and ASCII conversion. `base_convert(substr( md5($string), 0, 16))` could also be expressed as "shift the bits of the MD5 hash integer value to the right by 64 places".

BTW, I confused bits and characters in my comment above. The truncation of 16 characters converts the 128-bit hash (expressed as 32 hex characters) into a 64-bit hash. Which worked on my machine because there PHP integers have 64 bits, but on some machines they only have 32, which is why I suppose we bit-shift the md5 hash to 32 bits:

    base_convert( substr( md5( $string ), 0, 8 ), 16, 10 );

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
gbirke: I don't have pretty cards, but here you go ...

img_4563

gbirke:
We had another look at https://phabricator.wikimedia.org/T170266 in the story time yesterday, where we also talked about the requirements again:
* as short as possible (please reduce the number of chars to 6).
* robust checksum when people are "creative" when writing the code (with and without dashes, with spaces or underscore instead of dash, etc).

contexts/PaymentContext/tests/Unit/Domain/LessSimpleTransferCodeGeneratorTest.php
```diff
@@ -0,0 +1,81 @@
...
+		);
+	}
+
+	public function invalidCodeProvider(): iterable {
+		yield 'Empty code' => [ '' ];
+		yield 'Without checksum' => [ 'ACD-EFK-' ];
+		yield 'Missing dash' => [ 'ACDEFK-X' ];
+		yield 'Missing checksum dash' => [ 'ACD-EFK' ];
+		yield 'Missing both dashes' => [ 'ACDEFK' ];
+		yield 'Extra dash' => [ 'ACD-EFK-X-' ];
```
JeroenDeDauw: These should now be valid no? @gbirke

gbirke: I thought about it when I read them and was unsure. Because we're not the not the ones validating the codes (that's done by the Gruen software), it's more of a courtesy or "reference implementation" of what we think valid codes should be. But then we'd also need to incorporate the 2-char prefixes in the tests. Yesterday in the story time we defined that they should be `XW` (for personalized donations) and `XR` (for anonymous donations). And at the beginning of this issue you wrote  

> Did not include the prefix in this class as it is not clear to me if
it should go in there or not. If it depends on something else, such
as the type of donation, then probably not.

As it turns out, it depends on the type of data in the donation. I would have included the fixed prefix in the `TransferCode` class constructor (simulating an enum type with constants of `TransferCode` and checks inside the constructor), but I don't feel strongly about it.

Questions to @KaiNissen : 

1) Should we implement `validateTransferCode` at all in the Frontend or just use the code as an example in the documentation for Gruen?
2) If we leave the code in, does it make sense to include the prefix in the `LessSimpleTransferCodeGenerator`? At least it must be part of the checks in  `validateTransferCode` because it must be passed to the checksum calculation.

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -30,14 +33,13 @@ public function __construct( array $checksumCharacters ) {
 	 * @return string The checksum as a single character present in the constructors array argument
 	 */
 	public function createChecksum( string $string ): string {
-		$checksum = array_sum(
-			array_map(
-				'ord', // Returns the ASCII value of the character
-				str_split( md5( $string ) )
-			)
-		);
+		$checksum = hexdec( substr( md5( $this->normalizeString( $string ) ), 0, 8 ) );
```
wiese: Why 8, not - say - 16?

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -30,14 +33,13 @@ public function __construct( array $checksumCharacters ) {
...
 
 		return $this->checksumCharacters[$checksum % count( $this->checksumCharacters )];
 	}
 
+	private function normalizeString( string $string ): string {
+		return strtoupper( preg_replace( $this->cleanupPattern, '', $string ) );
```
wiese: Intuitively, I think that

    $generator->createChecksum( 'AAAA-AAAA-' )

should give the same checksum as

    $generator->createChecksum( 'aaaa-aaaa-' )


But it does not. Is that a feature?

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
gbirke: 8 chars equals 32 bit. 16 chars would be 64 bits, which would break the checksum mechanism on 32 bit PHP platforms where `PHP_INT_SIZE` is 4 (bytes). On those platforms the resulting number from `hexdec` would be a float and modulos of floats are always 0. 

I don't have the necessary math to prove it, but my gut feeling is that using a smaller portion of the hash is not a problem for possible hash collisions since our character set only allows for 20 different outcomes, which could be expressed in 5 bits and we could even use CRC (which returns 8 an bit checksum) instead of MD5 for that.

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
gbirke: It's a bug. Thank you for catching it!

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
wiese: Not saying we should introduce more incompatibilities, but if we really strive to support x64 we [should strengthen our efforts](https://travis-ci.org/wmde/FundraisingFrontend/jobs/278578315#L3700).

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
wiese: #964

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
JeroenDeDauw: So this change broke our compat with x64? If so I suggest reverting the relevant part

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
JeroenDeDauw: Ugh, should have been more careful with merge. The regex stuff is also broken.

JeroenDeDauw: Fixed in https://github.com/wmde/FundraisingFrontend/pull/966

contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -22,6 +24,7 @@ public function __construct( array $checksumCharacters ) {
 		}
 
 		$this->checksumCharacters = $checksumCharacters;
+		$this->cleanupPattern = '/[^' . implode( '', $checksumCharacters ) . ']/';
```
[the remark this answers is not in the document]
JeroenDeDauw: "Fixed" in https://github.com/wmde/FundraisingFrontend/pull/966

PR #942, opened 2017-08-17
Payment delay in embedded membership form

KaiNissen:
This PR fixes that the payment delay for membership fees is not known to the membership form that is embedded in the donation confirmation page.


KaiNissen:
obsolete

JeroenDeDauw:
Do I still need to think about this if it is obsolete?

PR #946, opened 2017-08-21
 	Move mail templates to dedicated directory 

JeroenDeDauw:
https://phabricator.wikimedia.org/T172885


JeroenDeDauw:
Only need to review the second commit, first has been merged already

JeroenDeDauw:
still needed: url function in main templates. Thoughts so far: create own URL generation interface with adapter for the framework one so that we can avoid binding to the framework in the factory and tests.

PR #948, opened 2017-08-23
Improve moving of mail templates

gbirke:
Use correct templating environment in RenderMailTemplatesCommand.
Remove unneccessary filter from mail templating environment (mail
templates don't link to asset files that need to be prefixed).
Replace anonymous UrlGenerator class with standalone one since it's used
in 3 places now.
Add new config settings to vagrant environment.

### DIFF
--- a/app/UrlGenerator.php
+++ b/app/UrlGenerator.php
@@ -1,0 +1,19 @@
+<?php
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App;
+
+use Symfony\Bridge\Twig\Extension\RoutingExtension;
+use WMDE\Fundraising\Frontend\Infrastructure\UrlGenerator as UrlGeneratorInterface;
+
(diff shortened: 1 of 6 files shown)


app/UrlGenerator.php
```diff
@@ -0,0 +1,19 @@
+<?php
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App;
+
+use Symfony\Bridge\Twig\Extension\RoutingExtension;
+use WMDE\Fundraising\Frontend\Infrastructure\UrlGenerator as UrlGeneratorInterface;
+
+class UrlGenerator implements UrlGeneratorInterface {
```
JeroenDeDauw: I spend 2 min being confoosed at UrlGeneratorInterface, not finding which commit changed that name. Class could be TwigBasedUrlGenerator (or SilexBasedUrlGenerator?) so the unexpected aliasing is not needed.

PR #949, opened 2017-09-11
Moved templates to skin folder

JeroenDeDauw:


build/vagrant/install_app.sh
```diff
@@ -9,6 +9,7 @@ ln -fs ../vendor/wmde/fundraising-frontend-content/resources web/resources
 
 npm install
 npm run build-js
+npm run build-assets
```
gbirke: You need to switch the order because `build-assets` creates the `_js` directory that `build-js` uses.

deployment/tasks/build.yml
```diff
@@ -19,5 +19,8 @@
 - name: Compile JavaScript
   command: npm run build-js chdir={{ build_dir }}
 
+- name: Build assets
```
gbirke: You need to switch the order because `build-assets` creates the `_js` directory that `build-js` uses.

skins/10h16/templates/Base_Layout.html.twig
```diff
@@ -16,13 +16,13 @@
...
     <meta property="og:description" content="Mir liegt Wikipedia am Herzen, deshalb habe ich gespendet. Denn was wäre das Netz ohne Wikipedia?" />
 
-    <link rel="shortcut icon" href="{$ basepath $}/skin/10h16/_assets/img/favicon.ico" />
-    <link rel="apple-touch-icon" href="{$ basepath $}/skin/10h16/_assets/img/apple-touch-icon.png" />
+    <link rel="shortcut icon" href="{$ basepath $}/skins/10h16/_assets/img/favicon.ico" />
+    <link rel="apple-touch-icon" href="{$ basepath $}/skins/10h16/_assets/img/apple-touch-icon.png" />
 
     {% block styles %}
     <link href="{$ basepath $}/res/css/fontcustom.css" rel="stylesheet" type="text/css" />
     <link href="{$ basepath $}/res/css/lib/jquery.ui/jquery-ui.css" rel="stylesheet" type="text/css" />
```
KaiNissen: Shouldn't we move those files, too? They also seem quite related to the skin.

KaiNissen: Oh, not part of the task.

JeroenDeDauw: Well it is, but we don't need to do everything in one PR. Already did this stuff in https://github.com/wmde/FundraisingFrontend/pull/950 but that still needs a tiny bit of work

PR #952, opened 2017-09-14
Cookie configuration

KaiNissen:
This PR takes care of wmde/fundraising#1298, which is part of [phab:T174683](https://phabricator.wikimedia.org/T174683).

### DIFF
--- a/app/RouteHandlers/ApplyForMembershipHandler.php
+++ b/app/RouteHandlers/ApplyForMembershipHandler.php
@@ -9,6 +9,7 @@ use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+use WMDE\Fundraising\Frontend\Infrastructure\ConfigurableCookie;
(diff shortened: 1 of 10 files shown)


src/Infrastructure/ConfiguredCookie.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
+use Symfony\Component\HttpFoundation\Cookie;
+
+/**
+ * @licence GNU GPL v2+
+ * @author Kai Nissen < kai.nissen@wikimedia.de >
+ */
+class ConfiguredCookie {
```
JeroenDeDauw: CookieBuilder is probably a better name, since this itself is not a cookie

KaiNissen: Naming things... Thank you, just amended the commit.

gbirke:
I'm not sure if we really need to make the cookie settings configurable - what would be a scenario where we'd want to change the settings? And with `cookieBuilder` we already have a central entry point for creating cookies with the right settings. If you don't have objections I'd like to do a followup where I "hard-code" the cookie settings and remove the config.

KaiNissen:
No objections from my side. I can't think of a case in which we'd want to change the settings ever again.

JeroenDeDauw:
Sounds reasonable to me too

PR #955, opened 2017-09-19
T172892 Determine skin to use based on request

wiese:
Works best with additional `config.prod.json`

	"skin": {
		"options": [
			"10h16", "cat17"
		]
	},
	"cookie": {
		"secure": false
	}

Resources:
* http://srcmvn.com/blog/2013/02/22/writing-silex-service-providers
* https://silex.symfony.com/doc/2.0/providers.html#creating-a-provider

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -142,7 +142,7 @@ use WMDE\Fundraising\Frontend\Presentation\Presenters\InternalErrorHtmlPresenter
 use WMDE\Fundraising\Frontend\Presentation\Presenters\MembershipApplicationConfirmationHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\MembershipFormViolationPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\PageNotFoundPresenter;
-use WMDE\Fundraising\Frontend\Presentation\SkinManager;
+use WMDE\Fundraising\Frontend\Presentation\SkinSettings;
 use WMDE\Fundraising\Frontend\Presentation\TwigTemplate;
 use WMDE\Fundraising\Frontend\SubscriptionContext\DataAccess\DoctrineSubscriptionRepository;
 use WMDE\Fundraising\Frontend\SubscriptionContext\Domain\Repositories\SubscriptionRepository;
@@ -305,7 +305,7 @@ class FunFunFactory {
 
 		$pimple['twig'] = function() {
 			$config = $this->config['twig'];
-			$config['loaders']['filesystem']['template-dir'] = 'skins/' . $this->getSkinManager()->getSkin() . '/templates';
+			$config['loaders']['filesystem']['template-dir'] = 'skins/' . $this->getSkinSettings()->getSkin() . '/templates';
 
 			$twigFactory = $this->newTwigFactory( $config );
 			$configurator = $twigFactory->newTwigEnvironmentConfigurator();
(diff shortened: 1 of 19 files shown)


JeroenDeDauw:
For https://phabricator.wikimedia.org/T172892

tests/EdgeToEdge/SkinTest.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	private function assertSkinResponseCookie( string $expected, Response $response ): void {
+		$cookies = $response->headers->getCookies();
+		foreach ( $cookies as $cookie ) {
+			if ( $cookie->getName() !== 'skin' ) {
+				continue;
+			}
+			$this->assertSame( $expected, $cookie->getValue() );
+		}
```
JeroenDeDauw: If it is not found, no assertion will be made. While this results in a warning (or error?) when using PHPUnit strict mode, it is a bit dodgy. You could return after the assert and put a $this->fail() at the end of the method.

tests/EdgeToEdge/SkinTest.php
```diff
@@ -0,0 +1,90 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+class SkinTest extends WebRouteTestCase {
+
+	public function testDefaultSkinUsed(): void {
+		$client = $this->createClient( $this->getDummyConfig(), null, self::DISABLE_DEBUG );
+		$client->request( 'GET', '/' );
+
+		$this->assertContains( '10h16', $client->getResponse()->getContent() );
+		$this->assertNoSkinResponseCookie( $client->getResponse() );
```
JeroenDeDauw: If you put the assert methods (and all methods really) right below where they are first used, less vertical scrolling is needed to real the class.

JeroenDeDauw:
Looks good so far apart from minor CS-ish stuff

src/Factories/FunFunFactory.php
```diff
@@ -468,6 +471,10 @@ function( string $name, array $parameters = [] ): string {
 			return new SofortClient( $config['config-key'] );
 		};
 
+		$pimple['skin'] = function () {
```
gbirke: Since `skin` is [more of a parameter than a service](https://pimple.symfony.com/#defining-parameters) I think it's sufficient to assign it directly instead of using a function with a return value.

src/Presentation/SkinManager.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Presentation;
+
+use InvalidArgumentException;
+
+class SkinManager {
```
gbirke: I'm not sure if this class "manages" anything, it looks like a fancy mutable value object to me. Maybe call it `SkinSettings`, `SkinConfiguration` or just `Skins`?

tests/Unit/Presentation/SkinSettingsTest.php
```diff
@@ -0,0 +1,60 @@
...
+use InvalidArgumentException;
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\Presentation\SkinSettings
+ */
+class SkinSettingsTest extends TestCase {
+
+	public function testDefaultSkinIsSet(): void {
+		$manager = new SkinSettings( ['a', 'b'], 'a', 500 );
+		$this->assertSame( 'a', $manager->getDefaultSkin() );
```
JeroenDeDauw: I suppose SkinSettings was renamed, cause $manager is kinda weird like this

PR #959, opened 2017-09-21
Docker dev env

wiese:


build/app/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM php:7.1-fpm
+
+RUN mkdir -p /usr/share/nginx/www/spenden.wikimedia.de/current/var/cache
+RUN mkdir -p /usr/share/nginx/www/spenden.wikimedia.de/current/var/doctrine_proxies
+RUN mkdir -p /usr/share/nginx/www/spenden.wikimedia.de/current/var/log
+RUN chown -R www-data:www-data /usr/share/nginx/www/spenden.wikimedia.de/current/var
```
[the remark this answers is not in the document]
wiese: Developed step by step and not groomed thereafter - this way [clearly against the best practice](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers).

gbirke:
Rebased on current master. @timEulitz will try it out on his machine and when he's successful, we can  merge this.

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+      - mailhog
+
+  database:
+    image: mysql:5.6
+    environment:
+      - MYSQL_RANDOM_ROOT_PASSWORD=yes
+      - MYSQL_USER=fundraising
+      - MYSQL_PASSWORD=INSECURE PASSWORD
+      - MYSQL_DATABASE=fundraising
+    expose:
```
gbirke: @wiese Is there a reason why there is no volume defined for the database? The current setup will delete all data and tables as soon as docker-compose is stopped.

wiese: No explicit reason but I did not consider it needed so far. The container (and its volume) will be reused unless you explicitly remove it.
Mind "Where to Store Data" in https://hub.docker.com/_/mysql/

JeroenDeDauw:
image

JeroenDeDauw:
Please beware that the README has a table of contents at its top which is manually maintained. So when changing the section headers or removing/adding/moving sections, this thing needs updating. This PR broke the TOC without anyone noticing, hence me mentioning it here, since perhaps no one was aware of the need to keep this in mind.

gbirke:
Thank for pointing this out, I've added an npm task to autogenerate the TOC in #1221

PR #960, opened 2017-09-21
T176227 Fix layout of payment type options

gbirke:
Align payments horizontally in three or four columns, depending how many
payment types there are.

ATTENTION: "Sofortüberweisung" does not fit the column layout (neither 3
or 4). It only works if it comes as the last option.


JeroenDeDauw:
+

PR #964, opened 2017-09-22
T170416 add tests showing case-sensitity of createChecksum

wiese:
master is red with this!

PR #966, opened 2017-09-25
Made ChecksumGenerator insensitive just in case

JeroenDeDauw:
1) Made ChecksumGenerator insensitive just in case

2) Replaced broken regex with less mawgial replace. Even if the regex was not broken, it would allow for https://github.com/wmde/FundraisingFrontend/pull/940#issuecomment-330871351. With this PR the code just does https://github.com/wmde/FundraisingFrontend/pull/940#issuecomment-331111153

3) Made tests more focused (ie tests about capitalization no longer include ignored characters as well)

Based on https://github.com/wmde/FundraisingFrontend/pull/964, for https://phabricator.wikimedia.org/T170266, following up to https://github.com/wmde/FundraisingFrontend/pull/940

### DIFF
--- a/contexts/PaymentContext/src/Domain/ChecksumGenerator.php
+++ b/contexts/PaymentContext/src/Domain/ChecksumGenerator.php
@@ -11,8 +11,6 @@ class ChecksumGenerator {
 
 	private $checksumCharacters;
 
-	private $cleanupPattern;
-
 	/**
 	 * @param array $checksumCharacters Characters that can be used for the checksum
 	 */
@@ -24,7 +22,6 @@ class ChecksumGenerator {
 		}
 
 		$this->checksumCharacters = $checksumCharacters;
-		$this->cleanupPattern = '/[^' . implode( '', $checksumCharacters ) . ']/';
 	}
 
 	/**
@@ -39,7 +36,7 @@ class ChecksumGenerator {
 	}
 
 	private function normalizeString( string $string ): string {
-		return strtoupper( preg_replace( $this->cleanupPattern, '', $string ) );
+		return strtoupper( str_replace( [ '-', '_', ' ' ], [ '', '', '' ], $string ) );
 	}
 
 }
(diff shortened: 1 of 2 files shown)


contexts/PaymentContext/src/Domain/ChecksumGenerator.php
```diff
@@ -39,7 +36,7 @@ public function createChecksum( string $string ): string {
 	}
 
 	private function normalizeString( string $string ): string {
-		return strtoupper( preg_replace( $this->cleanupPattern, '', $string ) );
+		return strtoupper( str_replace( [ '-', '_', ' ' ], [ '', '', '' ], $string ) );
```
wiese: I'm torn about this, do not believe in blacklists.
[Your comment](https://github.com/wmde/FundraisingFrontend/pull/940#issuecomment-330871351) is right in that the 0-pad example feels dodgy, but if we keep in mind that the resulting normalized string will "only" be used for checksum generation and never output, the benefits of ignoring all possible funky delimiters outweighs the risk of a unintuitive checksum collision (which is not even a thing because it does not come with the pretense of uniqueness).

JeroenDeDauw: I mainly went with this approach because it was easier and is simpler than the regex. Try seeing it as a whitelist of characters with special handling, no blacklisting needed :)

JeroenDeDauw:
With this cycle time we won't get this task done in this sprint :(

wiese:
@JeroenDeDauw wrote
> With this cycle time we won't get this task done in this sprint :(

Please make sure you can attend the retro. I'd like to talk about this as well.

JeroenDeDauw:
Ugggh need a change to the interface for the next step and due to laziness put this implementation in this repo for now. Fail. There was no real benefit in doing that and now it bites us in the ass. Will merge this PR here now, then submit another one that removes the class, and another one adding it to the correct component. Will link that here so you can follow up.

PR #967, opened 2017-09-25
Make checksum 32-bit safe

gbirke:
Previous implementation failed on 32 bit systems, now it works.


JeroenDeDauw:
To me what I did originally is a lot easier to understand. That was generating this number by summing the ASCII value of every character, something I think is easy to grasp for developers. I don't think the same is true for what is on master, and even less so for what is in this PR.

JeroenDeDauw:
PS: I am happy to move the code there

JeroenDeDauw:
Obsoleted by https://github.com/wmde/fundraising-donations/pull/8 if I'm not mistaken

PR #969, opened 2017-09-26
Remove transfer code code to Donation repo

JeroenDeDauw:
Added again with https://github.com/wmde/fundraising-donations/pull/6


JeroenDeDauw:
Oops, I thought this was the re-add one and merged it already >_> Don't think this is worthwhile reverting and resubmitting in this case.

PR #976, opened 2017-10-25
T177416 Opt out of donation receipt - membership

wiese:
Add membership donation receipt opt-out functionality.

Requires `vendor/bin/doctrine orm:schema-tool:update`

Upstream PR https://github.com/wmde/FundraisingStore/pull/143 resulted in v7.0.0
https://github.com/wmde/fundraising-donations/pull/15 was store bump only.

In parallel with https://github.com/wmde/FundraisingFrontend/pull/975

### DIFF
--- a/contexts/MembershipContext/src/Domain/Model/Application.php
+++ b/contexts/MembershipContext/src/Domain/Model/Application.php
@@ -6,7 +6,6 @@ namespace WMDE\Fundraising\Frontend\MembershipContext\Domain\Model;
 
 use RuntimeException;
 use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\PaymentType;
-use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\PayPalData;
 use WMDE\Fundraising\Frontend\PaymentContext\Domain\Model\PayPalPayment;
 
 /**
@@ -42,8 +41,9 @@ class Application {
 	private $isCancelled;
 	private $isConfirmed;
 	private $isDeleted;
+	private $donationReceipt;
 
-	public static function newApplication( string $type, Applicant $applicant, Payment $payment ): self {
+	public static function newApplication( string $type, Applicant $applicant, Payment $payment, ?bool $donationReceipt ): self {
 		return new self(
 			null,
 			$type,
@@ -52,12 +52,13 @@ class Application {
 			self::NO_MODERATION_NEEDED,
 			self::IS_CURRENT,
 			self::IS_PENDING,
-			self::IS_NOT_DELETED
+			self::IS_NOT_DELETED,
+			$donationReceipt
 		);
 	}
 
 	public function __construct( ?int $id, string $type, Applicant $applicant, Payment $payment,
-		bool $needsModeration, bool $isCancelled, bool $isConfirmed, bool $isDeleted ) {
+		bool $needsModeration, bool $isCancelled, bool $isConfirmed, bool $isDeleted, ?bool $donationReceipt ) {
 		$this->id = $id;
 		$this->type = $type;
 		$this->applicant = $applicant;
@@ -66,6 +67,7 @@ class Application {
 		$this->isCancelled = $isCancelled;
 		$this->isConfirmed = $isConfirmed;
 		$this->isDeleted = $isDeleted;
+		$this->donationReceipt = $donationReceipt;
 	}
 
 	public function getId(): ?int {
@@ -86,6 +88,10 @@ class Application {
 
 	public function getType(): string {
 		return $this->type;
+	}
+
+	public function getDonationReceipt(): ?bool {
+		return $this->donationReceipt;
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


contexts/MembershipContext/tests/Data/ValidMembershipApplication.php
```diff
@@ -63,12 +63,15 @@ class ValidMembershipApplication {
 	const TEMPLATE_NAME = 'Some_Membership_Form_Template.twig';
 	const FIRST_PAYMENT_DATE = '2021-02-01';
 
+	private const DONATION_RECEIPT = true;
```
JeroenDeDauw: OPTS_INTO_

JeroenDeDauw:
Slightly offtopic:

The changes for this task made me wonder if this really is info that should go into the Donation. After looking at the ticket, which says this stuff should be in the export, this indeed seems like the way to go.

What about the newsletter signup? This info is inside the Donation object... I do not know at which point it is actually used. Looks like this is not happening in the FF app, and in the FOC it's a bit difficult to tell what's going on. I figure it might be the case that this info just happens to be stored in the same database table as the donation info, and only for that reason is in the Donation object. Anyone know what the deal with this is? @gbirke @KaiNissen

contexts/MembershipContext/src/Domain/Model/Application.php
```diff
@@ -42,8 +41,9 @@ class Application {
 	private $isCancelled;
 	private $isConfirmed;
 	private $isDeleted;
+	private $donationReceipt;
 
-	public static function newApplication( string $type, Applicant $applicant, Payment $payment ): self {
+	public static function newApplication( string $type, Applicant $applicant, Payment $payment, ?bool $donationReceipt ): self {
```
JeroenDeDauw: Needing a new parameter here makes me a bit sad, though can't think of a better way of doing this.

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
```diff
@@ -446,4 +449,14 @@ public function testCookieFlagsSecureAndHttpOnlyAreSet(): void {
...
 	}
 
+	public function testGivenDonationReceiptOptOutRequest_applicationHoldsThisValue(): void {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ): void {
+			$parameters = $this->newValidHttpParameters();
+			$parameters['donationReceipt'] = '0';
+
+			$client->request( Request::METHOD_POST, self::APPLY_FOR_MEMBERSHIP_PATH, $parameters );
+
+			$this->assertFalse( $factory->getMembershipApplicationRepository()->getApplicationById( 1 )->getDonationReceipt() );
```
JeroenDeDauw: `getDonationReceipt` sounds like it returns a donation receipt

JeroenDeDauw:
PR looks reasonable so far

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
wiese: Went with consistent naming in [Donation](https://github.com/wmde/FundraisingStore/commit/c8308cb8df6af7045ff0ee69b73c013da0e9b79d#diff-cd62f2322362fb9135726ef3359173e8R268) and Membership. Thinking about this led to [your other question](https://github.com/wmde/FundraisingFrontend/pull/976#issuecomment-339801033) - the answer to which would have blown things out of proportion. I think this is a good task for when the store parts get moved into the contexts.

gbirke:
> The changes for this task made me wonder if this really is info that should go into the Donation. After looking at the ticket, which says this stuff should be in the export, this indeed seems like the way to go.

The receipt stuff is closely tied to the respective donation and membership. It's saying "I don't want a receipt for *this* donation/membership". If, in some distant future, we have user accounts where people can manage their receipt preferences and addresses, then it would make sense to move this flag out to a then-to-be-created `donor`/`member` table.

The "opts into newsletter" stuff is more broad and only tied to donation by the export format (and the data storage layer at the moment). It *could* be migrated into a new context, depending on how the whole "Subscription" topic is developing (at the moment it's not).

JeroenDeDauw:
> The "opts into newsletter" stuff is more broad and only tied to donation by the export format (and the data storage layer at the moment). It could be migrated into a new context, depending on how the whole "Subscription" topic is developing (at the moment it's not).

I'd start by moving it out of Donation. The AddDonationUC would then simply call a service *such* as NewsletterThinghyService that would do the persisting (which in the first step would remain the same as it is now). No need to create a new context right off the bat. Same goes for things such as the tracking info (which is inside of Donation with a TODO but not inside (Mebership)Application)

tests/EdgeToEdge/Routes/ApplyForMembershipRouteTest.php
JeroenDeDauw: Oh wait, forgot in my last comment the name we're talking about is defined in the store component. Still think I'd have been better to go with another name, but indeed not an issue in this PR. I just wanted to point out I'm not 100% happy with this.

PR #981, opened 2017-11-02
T177934 First steps layout & content pages

wiese:
Based on #980, incorporates [`frontend_redesign`](https://github.com/wmde/FundraisingFrontend/tree/frontend_redesign), adds to it.

This is still WIP but in a dedicated folder, deactivated by the default configuration.

295da77 diff shows delta with existing WMDE scripts.

Change your config to test drive

	// ...
	"skin": {
		"options": [
			"10h16", "cat17"
		],
		"default": "cat17"
	}


wiese:
Tried deploying this to test to find that most likely our node version on deploy is too old (`v4.2.6`) - fails silently (w/o `-vvv` on ansible) but leaves cat17's web folder with many files to be desired.

    npm run build-assets [...]
    "stderr": "\nevents.js:141\n      throw er; // Unhandled 'error' event\n      ^\nTypeError: this is not a typed array.\n    at Function.from (native)\n

PR #984, opened 2017-11-06
Show warning when house number is missing in donation form

JeroenDeDauw:
https://phabricator.wikimedia.org/T174669

Currently done with a message shown below the field

Like most of the form the message is not localized yet. Skipped
that since it requires first adding a message to the content repo
and because the translation function is confoosing me.

### DIFF
--- a/skins/10h16/js/lib/view_handler/warning_box.js
+++ b/skins/10h16/js/lib/view_handler/warning_box.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var objectAssign = require( 'object-assign' ),
-	_ = require( 'underscore' ),
 
 	WarningBox = {
 		issueBox: null,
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
image

skins/10h16/js/lib/view_handler/warning_box.js
```diff
@@ -0,0 +1,31 @@
...
+			if ( this.validationFunction( fieldValue ) ) {
+				this.issueBox.show();
+			}
+			else {
+				this.issueBox.hide();
+			}
+		}
+	},
+
+	createHandler = function ( issueBoxElement, shouldShowWarningFunction ) {
```
gbirke: Although the other view handles also don't have doc comments, I'd wish we had them here.

skins/10h16/templates/Donation_Form_Personal_Data.html.twig
```diff
@@ -365,6 +365,7 @@
                             <label for="street">Stra&szlig;e</label>
                             <input id="street" name="street" type="text" value="{$ street $}" data-pattern="^((?![M|m]uster).)+$" placeholder="z.B. Musterstra&szlig;e 123" class="required">
                             <span class="validation icon-placeholder"></span>
+                            <p id="street-warning-box" style="margin-left: 180px; margin-top: 10px; color: darkred; display: none">Ist die Adresse korrekt oder fehlt noch Ihre Hausnummer?</p>
```
gbirke: We try to avoid inline styles, as they are all possible blockers for introducing [Content Security Policy](https://en.wikipedia.org/wiki/Content_Security_Policy) in the future. Here, the styles can be moved to a class and a rule in the stylesheet.

PR #988, opened 2017-11-06
[WIP] T177934 Build skins

wiese:
* make sure npm postinstall scripts are invoked (#981 follow-up)

Todo: update deploy's npm (https://github.com/wmde/fundraising-infrastructure/pull/185), let's see if it fixes itself then without this change to the task.


wiese:
This is indeed obsolete with https://github.com/wmde/fundraising-infrastructure/pull/185 applied and `node_modules` cleaned.

PR #993, opened 2017-11-08
Update Vagrant node.js to 8

gbirke:
Otherwise npm build-assets will fail on Vagrant for cat17 skin


wiese:
Can not reproduce

    vagrant@vagrant:/vagrant$ node --version
    v6.11.4

works fine for me.

PR #997, opened 2017-11-08
T177934 Resolve many content pages issues

wiese:
For T179930, T179929, T179936, T179931

### DIFF
--- a/skins/cat17/src/sass/layouts/pages/privacy_protection.scss
+++ b/skins/cat17/src/sass/layouts/pages/privacy_protection.scss
@@ -1,0 +1,7 @@
+.page-privacy_protection {
+	.content {
+		ul {
+			margin-left: 1.5em;
+		}
+	}
+}
--- a/skins/cat17/templates/Display_Page_Layout.twig
+++ b/skins/cat17/templates/Display_Page_Layout.twig
@@ -1,6 +1,6 @@
 {% extends 'Base_Layout.html.twig' %}
 
-{% block page_identifier %}{$ page_id $}{% endblock %}
+{% block page_identifier %}page-{$ page_id $}{% endblock %}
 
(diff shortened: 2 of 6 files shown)


skins/cat17/templates/Display_Page_Layout.twig
```diff
@@ -1,6 +1,6 @@
 {% extends 'Base_Layout.html.twig' %}
 
-{% block page_identifier %}{$ page_id $}{% endblock %}
+{% block page_identifier %}page-{$ page_id $}{% endblock %}
```
wiese: I don't particularly like the prefix but without it in the CSS selectors I am afraid to involuntarily match styles. Will follow-up and consolidate once all pages work the same way (prepend it to only the identifier rendered onto the body, not the page_identifier variable).

skins/cat17/src/sass/layouts/pages/privacy_protection.scss
```diff
@@ -0,0 +1,7 @@
+.page-privacy_protection {
+	.content {
+		li {
+			list-style-position: inside;
```
gbirke: The list at the top of the page looks ok, but when you have longer lines, they are not indented correctly when you have a line break in the list (see section "Betroffene Personengruppen und Daten oder Datenkategorien")

wiese: Looks pretty much like in 10h16 on prod - please define "correctly" (T179936).

gbirke: padding-left: 1.5em;
    text-indent: -1.5em;

looks nice

wiese:
Rebased and amended (T179936) follwing @gbirke's comment.

PR #1001, opened 2017-11-09
T177929 Improve cat17 build process

wiese:
Preparation for dev watch


wiese:
What do you think about - in the spirit of _continuous_ improvement - we merge this and further improve it in separate PR(s)?

PR #1004, opened 2017-11-09
T177935 Donation comments in cat17

gbirke:
Expand template with example
Add & adapt JavaScript from 10h16
Add script block to base template

### DIFF
--- a/skins/cat17/src/sass/layouts/pages/comments_list.scss
+++ b/skins/cat17/src/sass/layouts/pages/comments_list.scss
@@ -1,0 +1,24 @@
+.page-comments-list {
+	.introduction {
+		padding-top: 80px;
+		@include bkp(md) {
+			padding-top: 45px;
+		}
+	}
+	article {
+		margin: 40px 0 0;
+		span {
+			display: block;
+			margin-bottom: 10px;
+			&:first-child {
+				font-weight: $font-main-semibold;
+			}
+			&.date-time {
+				font-weight: $font-main-light;
+			}
+		}
+		&:last-child {
+			margin-bottom: 60px;
+		}
+	}
+}
--- a/skins/cat17/src/sass/layouts/pages/comments_list.scss
+++ b/skins/cat17/src/sass/layouts/pages/comments_list.scss
@@ -1,7 +1,7 @@
 .page-comments-list {
 
-	.rss-link {
-		color: #00adee;
+	.subtitle a {
+		color: $brand-primary;
 	}
 
 	.introduction {
--- a/skins/cat17/src/sass/layouts/pages/comments_list.scss
+++ b/skins/cat17/src/sass/layouts/pages/comments_list.scss
@@ -1,4 +1,9 @@
 .page-comments-list {
+
+	.rss-link {
+		color: #00adee;
+	}
+
 	.introduction {
 		padding-top: 80px;
 		@include bkp(md) {
--- a/skins/cat17/src/scripts/donationComments.js
+++ b/skins/cat17/src/scripts/donationComments.js
@@ -8,18 +8,18 @@
  *
  * @constructor
  */
-var DonationComments = function ( container ) {
-	this.container = container;
+var DonationComments = function ( commentContainer, paginationContainer ) {
+	this.commentContainer = commentContainer;
+	this.paginationContainer = paginationContainer;
 };
 
 $.extend( DonationComments.prototype, {
 	init: function () {
-		var paginationContainer = this.container.parent().find( '.pagination' );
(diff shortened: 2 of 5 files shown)


skins/cat17/templates/Base_Layout.html.twig
```diff
@@ -37,12 +37,14 @@
 		{% include 'partials/' ~ footer_template %}
 	</div>
 
+	{% block scripts %}
```
wiese: I think it would be beneficial to have a _separate_ script block, independent of the scripts we always have to load (which should be as small as possible but something tells me we'll always have it), in order not to inadvertently overwrite them (forgotten `{$ parent() $}`). What do you think?

skins/cat17/templates/Comment_List.html.twig
```diff
@@ -1,7 +1,46 @@
...
+	<!-- nav content filled with paginator-sidebar via css / dont delete -->
+	<nav class="paginator paginator-top col-xs-12">
+	</nav>
+	<div class="container">
+		<div class="row">
+			<div class="introduction col-xs-12 col-sm-7 col-md-8 clearfix">
+				<h2>Spendenkommentare</h2>
+				<p class="subtitle">
+					Nachfolgend finden Sie eine Übersicht der letzten Spendenkommentare. Die neuesten Einträge sind auch
+					als <a href="{$ path( 'list-comments.rss', { piwik_campaign: 'Spendenseite', piwik_kwd: 'spendenliste' } ) $}" target="_top">RSS-Feed</a> verfügbar.</p>
```
wiese: `Comment_List.rss.twig` still needs to be copied over.

gbirke:
Thanks for the comments, I agree and have addressed them.

skins/cat17/src/scripts/donationComments.js
```diff
@@ -0,0 +1,156 @@
...
+			dataType: 'json',
+			success: function ( data ) {
+				self.numPages = Math.ceil( data.length / self.itemsPerPage );
+				self.commentContainer.html( self.renderHtml( data ) );
+				self.updatePagination();
+			}
+		} );
+	},
+
+	renderHtml: function ( data ) {
```
wiese: Strictly speaking this does not render, but return mixed type results, either a string, or object containing HTML elements - `.html()` just happens to get along with them.

skins/cat17/src/sass/layouts/pages/comments_list.scss
```diff
@@ -0,0 +1,29 @@
+.page-comments-list {
+
+	.rss-link {
+		color: #00adee;
```
wiese: `$brand-primary`

skins/cat17/src/scripts/donationComments.js
```diff
@@ -0,0 +1,156 @@
...
+ * @constructor
+ */
+var DonationComments = function ( commentContainer, paginationContainer ) {
+	this.commentContainer = commentContainer;
+	this.paginationContainer = paginationContainer;
+};
+
+$.extend( DonationComments.prototype, {
+	init: function () {
+		this.commentContainer.html( 'Spendenkommentare werden geladen ...' );
```
wiese: Being that guy, but can we please get copy from content via twig? For this page I don't see us getting a second chance to do that "while we are at it", but only as a task dedicated to i18n. A lost opportunity in my eyes.

skins/cat17/templates/Comment_List.html.twig
```diff
@@ -1,7 +1,45 @@
...
+				<p class="subtitle">
+					Nachfolgend finden Sie eine Übersicht der letzten Spendenkommentare. Die neuesten Einträge sind auch
+					als <a href="{$ path( 'list-comments.rss', { piwik_campaign: 'Spendenseite', piwik_kwd: 'spendenliste' } ) $}" class="rss-link" target="_top">RSS-Feed</a> verfügbar.</p>
+			</div>
+			<nav class="paginator paginator-sidebar col-md-4 clearfix comment-paginationContainer">
+				<div class="container">
+					<div class="row">
+						<div class="pages-wrap col-xs-12 col-sm-5 col-sm-offset-7">
+                            <span class="controls prev"><a href="#"><i class="icon-keyboard_arrow_left"></i><span
+											class="sr-only">Previous page of comments</span></a></span>
```
wiese: Even `sr-only` content should probably come from content, be German. And while we are at it we could use it and pass it to JS.

KaiNissen:
Right, the current skin used server requests for each page it was displaying. Since the way the application is fetching the comments is not performing good anyway, I'd say we keep the PR as it is and keep that in mind for [phab:T180110](https://phabricator.wikimedia.org/T180110).

PR #1007, opened 2017-11-10
T177930 Add Donation

wiese:
* add donation form html
* fix form field names to make them compatible with backend
* mv copy into content repo, use copy from DOM in JS
* fix _some_ of the escaping problems (T180215)
* dynamized many components based on backend config (e.g. available payment types)
* rm JS GET-parameter handling overriding backend initial-form-values
* rm some experimental features (T180432 = deac6b5, T180461 = 2fc28e9)
* started JS clean-up

Depends on https://github.com/wmde/fundraising-frontend-content/pull/47 (not in `test`, yet)

I'd like to hear what you think about `FormDataExtractor` before adding tests for it.

### DIFF
--- a/skins/cat17/src/app/lib/view_handler/display_payment_summary.js
+++ b/skins/cat17/src/app/lib/view_handler/display_payment_summary.js
@@ -104,9 +104,6 @@ var objectAssign = require( 'object-assign' ),
         this.memberShipTypeTextElement.text(this.memberShipTypeText[formContent.membershipType]);
       }
     },
-    capitalize: function (s) {
-      return s[0].toUpperCase() + s.slice(1);
-    },
     formatPaymentInterval: function ( paymentIntervalInMonths ) {
       return this.intervalTranslations[ paymentIntervalInMonths ];
     },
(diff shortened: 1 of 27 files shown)


skins/cat17/src/app/lib/view_handler/display_payment_summary.js
```diff
@@ -141,20 +143,25 @@ var objectAssign = require( 'object-assign' ),
...
-				elements.each( function () {
-					if( $( this ).data( 'display-error' ) === undefined ) {
-						$( this ).addClass( 'icon-error' );
-					}
-				} );
+
+			if( icon === undefined ) {
+				elements
+					// only configured icons are supposed to communicate validation problems
+					.filter( function () {
```
gbirke: I like this pattern, using `filter` instead of `each`. 👍

skins/cat17/src/app/lib/view_handler/display_payment_summary.js
```diff
@@ -1,202 +1,193 @@
...
-            $(this).addClass('icon-error');
-          }
-        });
-      }
-      else {
-        elements.addClass( iconsDictionary[value] );
-      }
-    }
-  };
+	DATA_EMPTY_TEXT = 'empty-text',
```
gbirke: When I first saw `DATA_EMPTY_TEXT` I was a bit confused and thought it to be a translation key. I think we should make it clearer that these constants have something to do with the DOM. Maybe with custom functions that use the constants, maybe with a private `Element` class or just by adding a comment.

PR #1008, opened 2017-11-10
extract membership pledge header message

KaiNissen:
related: wmde/fundraising-frontend-content#45


wiese:
Messages are not in content prod, yet!

PR #1010, opened 2017-11-10
T177930 Donation Confirmation

gbirke:
The donation confirmation page now works for all payment types.

Functionality that was integrated into the confirmation page in the old skin is not integrated yet, as they are tracked in different tasks:
* membership form with donor data already filled out
* donation comment (see https://phabricator.wikimedia.org/T180344 )
* SEPA Mandate (needs to be defined)

Translations are in the PR https://github.com/wmde/fundraising-frontend-content/pull/48


gbirke:
No referee needed, I like your improvements, thanks!

PR #1011, opened 2017-11-10
T177933 Become member

wiese:
This should be about there - maybe not perfect but good enough for integration and first UAT.

* integrate delivered HTML & JS
* mv logic to state aggregation
* add a validity state to default behavior of base section info, headline now part of default repertoire
* fix hack that always indicated (the only) payment type as selected, probably for demo-purposes (eaea339)
* T180941 (`FeeOptionSwitcher` question) is unresolved
* like donation, this lacks translation of error messages

Depends on (untracked) https://github.com/wmde/fundraising-frontend-content/pull/58


wiese:
Rebased

gbirke:
Screenshot of broken title and SEPA mandate:

issues

JeroenDeDauw:
Can we merge this in master and then do a bunch of follow ups? That way it is easier to parallelize and to distinguish code that is already reviewed from what still need attention. Just a thought.

wiese:
Thanks for catching that /validate-fee in particular. Addressed most issues. Also updated the newly created message keys (i.e. not introducing them the "wrong" way) - but am highly skeptical towards the code/mark-up change (now). Even the backend calls it `amount` and an amount it is (dosn't get any more generic than that)...

wiese:
Rebased

PR #1012, opened 2017-11-16
Show canceled status on confirmation page

gbirke:
When a user cancels and then reload the confirmation page, the donation
should be shown with as canceled, not with an "unknown" state.

See https://phabricator.wikimedia.org/T180720

Needs the content from
https://github.com/wmde/fundraising-frontend-content/pull/49 to be
merged.


wiese:
The way it is implemented now it should not need further changes for cat17, right?

gbirke:
Yes, it's already done for cat17.

PR #1017, opened 2017-11-17
T177930 remove ill-placed view hack vs anonymous debit

wiese:
This apparently is already covered by the push to state and SuboptionDisplayHandler.

Can not come up with a scenario in which the interface breaks without this code. Check, please.


JeroenDeDauw:
+0.42

PR #1019, opened 2017-11-17
T180437 show dedicated 'recurrent' message for payment types

wiese:
_payment_type\_*\_recurrent_info_ can now be configured in content repo for all payment types and will be shown according to the same rules as discussed for the PPL message.

Depends on https://github.com/wmde/fundraising-frontend-content/pull/51, dependency not tracked in composer to avoid lock conflicts.


JeroenDeDauw:
+1

PR #1020, opened 2017-11-17
T180395 Fix validation state

gbirke:
When the form was sent with invalid data (mostly postcode), the validation state was not restored, leading to weird validation display and starting on page 1 instead of page 2, even when the payment data was correct.


src/Presentation/Presenters/DonationFormViolationPresenter.php
```diff
@@ -114,4 +116,36 @@ private function getBankData( AddDonationRequest $request ): array {
...
+			Result::SOURCE_DONOR_LAST_NAME => 'address',
+			Result::SOURCE_DONOR_SALUTATION => 'address',
+			Result::SOURCE_DONOR_TITLE => 'address',
+			Result::SOURCE_DONOR_STREET_ADDRESS => 'address',
+			Result::SOURCE_DONOR_POSTAL_CODE => 'address',
+			Result::SOURCE_DONOR_CITY => 'address',
+			Result::SOURCE_DONOR_COUNTRY => 'address',
+
+		];
+		return array_reduce( $violations, function( $validationResult, ConstraintViolation $violation ) use ( $fieldGroups ) {
```
JeroenDeDauw: When code is too long for a single line like this I really appreciate having one argument per line, like this:

```php
array_reduce(
    $input,
    function( $arguments ) use ( $stuff ) {
        doThings();
    } 
);
```

JeroenDeDauw:
Had a look but did not really understand what is going on, so +0 from my side

skins/10h16/js/lib/reducers/input_validation.js
```diff
@@ -11,11 +11,44 @@ function inputIsValid( value, pattern ) {
...
+			// initialValidationResult is filled, in case of validation errors, violatedFields is filled.
+			// In all other cases, we can't decide and just pass the initial state from the reducer, but that should be ok
+			dataEnteredTransformer = newDataEnteredState(
+				!_.isEmpty( action.payload.violatedFields ) ||
+				!_.isEmpty( action.payload.initialValidationResult )
+			);
+			_.each( validationState, function ( value, key ) {
+				newValidationState[ key ] = {
+					dataEntered: dataEnteredTransformer( validationState[ key ].dataEntered ),
+					isValid: _.has( action.payload.violatedFields, key ) ? true : validationState[ key ].isValid
```
wiese: Should this not rather be

    isValid: _.has( action.payload.violatedFields, key ) ? false : validationState[ key ].isValid

(a field is not valid if it is contained in violatedFields)?

PR #1022, opened 2017-11-20
T177930 rm sepa confirmation

wiese:
SEPA confirmation is now always implicitly assumed


wiese:
I do concur that we probably have to display it nonetheless. https://phabricator.wikimedia.org/T180956

PR #1025, opened 2017-11-20
T177930 Bank Name

wiese:
* show iban and bic in payment summary by default
* show bank name near debit fields, send it to server
* test for SuboptionDisplayHandler (after errorous match of form elements inside hidden suboption)

### DIFF
--- a/skins/cat17/src/sass/layouts/_borders-system.scss
+++ b/skins/cat17/src/sass/layouts/_borders-system.scss
@@ -217,7 +217,7 @@ fieldset {
                             margin-top: 0;
                             padding: 11px 0 14px 4px;
                         }
-                        .oden {
+                        .iban-decision {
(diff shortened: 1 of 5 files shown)


skins/cat17/src/sass/layouts/_borders-system.scss
```diff
@@ -217,7 +217,7 @@ fieldset {
                             margin-top: 0;
                             padding: 11px 0 14px 4px;
                         }
-                        .oden {
```
JeroenDeDauw: what :)

wiese: Maybe he was [hungry](https://upload.wikimedia.org/wikipedia/commons/5/52/Oden_by_Mori_Chan.jpg).

PR #1027, opened 2017-11-20
Fix digit count regexes

JeroenDeDauw:
Should "fix" https://phabricator.wikimedia.org/T180936

I am not sure the change for the bank code regexes I made is correct.
Perhaps the bounds where omitted on purpouse to allow for more than 8 digits.
I doubt it, but please double check.

### DIFF
--- a/skins/10h16/templates/Donation_Form_Personal_Data.html.twig
+++ b/skins/10h16/templates/Donation_Form_Personal_Data.html.twig
@@ -210,13 +210,13 @@
                         <div id="non-sepa" class="slide-non-sepa">
                             <div class="input-container">
                                 <label for="account-number">Kontonummer</label>
-                                <input class="required bank-check" id="account-number" name="konto" type="tel" value="{$ konto $}" placeholder="Kontonummer" data-pattern="[0-9]{4,10}" title="Bis zu zehnstellige Kontonummer">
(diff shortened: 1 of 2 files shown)


wiese:
Code that consumes the patterns in JS is [here](https://github.com/wmde/FundraisingFrontend/blob/master/skins/10h16/js/lib/reducers/input_validation.js#L11).

The changes to the patterns look fair enough.

But this is the _pure_ client side validation only.

If I am not mistaken there is also one with an AJAX (so server-side in essence) roundtrip which apparently is not catching the faulty zip codes either, but should given the other server-side validation (on form POST) does so. The idea behind the ticket was more about syncing these two server-side validation mechanisms, rather then touching the skin(s) - it might be a nice addon, though.

JeroenDeDauw:
When reading the ticket I though that the AJAX stuff should be used when going through the form and entering something invalid. As far as I can tell this does not happen. I actually do not know when the AJAX stuff is used and was not able to find this within 5 min or so.

We have more such discrepancies in the validation I think. I recall us talking about this and deciding it is not an issue, since you should not run into it unless you do fishy stuff such as edit the JS in your browser or manually create a post request.

So this change is about the user perspective, while the other one is, in my mind, more about refactoring and getting everything nice and consistent.

wiese:
Please find my comment in https://phabricator.wikimedia.org/T180936

JeroenDeDauw:
Ok, so this is not solving the task you had in mind. It's still fixing a bug, so let's merge this?

PR #1030, opened 2017-11-21
T177930 Lose dead code

wiese:
* lose uncalled function implementations incl tracking goals 
  > (14:46:51) Kai_WMDE: ja, schmeiß weg. die zahlen sieht sich niemand an.
* rm the use of pages (left the implementation so it _could_ be reused for [page sections](https://github.com/wmde/FundraisingFrontend/blob/cat17_noGoals/skins/cat17/src/scripts/main.js#L134) - scroll, anchor, tracking, ...)

### DIFF
--- a/skins/cat17/src/scripts/donationForm.js
+++ b/skins/cat17/src/scripts/donationForm.js
@@ -274,7 +274,7 @@ $( function () {
     return validity.paymentData && addressIsValid() && bankDataIsValid();
   }
 
-  function triggerValidityCheckForPersonalDataPage() {
+  function triggerValidityCheckForPersonalData() {
     var formContent = store.getState().donationFormContent;
 
     if ( !addressIsValid() ) {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
+1

skins/cat17/src/scripts/donationForm.js
```diff
@@ -442,12 +379,11 @@ $( function () {
 
   // connect DOM elements to actions
   // fixme don't use interval, use form events instead? Discuss performance & other implications
-  //$( '#continueFormSubmit1' ).click( WMDE.StoreUpdates.makeEventHandlerWaitForAsyncFinish( handlePaymentDataSubmit, store ) );
   $('input').on('click, change', WMDE.StoreUpdates.makeEventHandlerWaitForAsyncFinish( handleGroupValidations, store ) );
   setInterval(handleGroupValidations, 1000);
```
gbirke: Totally unrelated to this PR, but caught my eye because of the proximity to line 447: We should get rid of this interval stuff & handle async validation better, see https://phabricator.wikimedia.org/T181140

PR #1034, opened 2017-11-22
[WIP] T177930 Shy submit button

wiese:
This needs rebasing on T181059 result (#1035) and some more work considering the resetting of _address_ when switching to anonymous (similar to what happens to _bankData_ when not applicable).

### DIFF
--- a/skins/cat17/src/app/tests/view_handlers/test_shy_submit_button.js
+++ b/skins/cat17/src/app/tests/view_handlers/test_shy_submit_button.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var test = require( 'tape' ),
+	sinon = require( 'sinon' ),
+	shyButton = require( '../../lib/view_handler/shy_submit_button' )
+;
+
+test( 'Given all positive values, element gets class removed', function ( t ) {
+	var element = {
+			toggleClass: sinon.stub()
+		},
+		handler = shyButton.createShySubmitButtonHandler( element )
+	;
+
+	handler.update( {
+		a: true,
+		b: true
+	} );
+
+	t.ok( element.toggleClass.withArgs( 'btn-unactive', false ).calledOnce );
+	t.end();
+} );
+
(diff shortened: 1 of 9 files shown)


skins/cat17/src/app/tests/view_handlers/test_shy_submit_button.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+var test = require( 'tape' ),
+	sinon = require( 'sinon' ),
+	shyButton = require( '../../lib/view_handler/shy_submit_button' )
+;
+
+test( 'Given all positive values, element gets class removed', function ( t ) {
```
gbirke: I was a bit stumped by this test, because I had to look at the class again to see what add/removing the class *means* (add=disable, remove=enable, which is also a bit counterintuitive for me).
Wouldn't it be better (i.e. semantic both on the DOM and on the test description side) to enable/disable the button and solve the display by css?

wiese: Do fully agree with all the objections you stated. However that would basically mean leaving no stone unturned - an honorable endeavor that, to me, seems incompatible with our deadline. That's why I went with
* extracting into small, graspable sub-components
* putting these under coverage (even if only to show how weird they behave [test as doc])

If you think this particular component is where you have to draw the line, be my guest.

gbirke: I can see what you're doing and admire your restraint. I'll approve this code.

wiese: Didn't even know [that](https://github.com/wmde/FundraisingFrontend/blob/02b665ae17c2782e437053eb098b4d001a24481a/skins/cat17/src/sass/components/_buttons.scss#L26) is a thing.
https://developer.mozilla.org/de/docs/Web/CSS/pointer-events

wiese:
Rebased

PR #1036, opened 2017-11-22
[WIP] T177930 No interval

wiese:
- [x] no more interval
- [x] refactored into reusable building blocks indicating form progress
- [x] resolves T180215 for the donation process (but let's double-check)
- [ ] inject appropriate validation states, lose validity checks inside view handlers
- [x] tests for proxy (in particular)


wiese:
@gbirke Thanks for the review. Amended per your remarks.

PR #1037, opened 2017-11-23
Enable cat17 skin in Dev environment

gbirke:
* Install npm and composer packages as non-privileged user to avoid npm
  errors.
* Install gulp-cli as a dependency to be able to run gulp.
* Keep dist directory (but not its contents) in repo.
* Enable skin in Vagrant dist configuration and loosen cookie setting to
  be able to store the skin name without using https.

### DIFF
--- a/build/vagrant/config.prod.json
+++ b/build/vagrant/config.prod.json
@@ -76,5 +76,16 @@
 		"theme": "x1",
 		"return-url": "https://spenden.wikimedia.de/show-donation-confirmation",
 		"testmode": true
-	}
+	},
+    "skin": {
+        "options": [
+            "10h16", "cat17"
+        ],
+        "default": "10h16",
+        "cookie-lifetime": 2592000
+    },
+    "cookie": {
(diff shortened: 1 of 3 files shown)


skins/cat17/package.json
```diff
@@ -30,6 +30,7 @@
   "devDependencies": {
     "browser-sync": "^2.18.13",
     "gulp": "^3.9.1",
+    "gulp-cli": "^1.4.0",
```
wiese: This surprises me as I did never install that as a direct dependency, assume it was installed indirectly.

build/vagrant/config.prod.json
```diff
@@ -76,5 +76,16 @@
...
+        "options": [
+            "10h16", "cat17"
+        ],
+        "default": "10h16",
+        "cookie-lifetime": 2592000
+    },
+    "cookie": {
+        "secure": false,
+        "httpOnly": false
+    }
```
JeroenDeDauw: image

PR #1038, opened 2017-11-23
T180339 Fix validation for email 10h16

gbirke:
The email field was always valid after it was filled and then removed,
since there was no client-side validation, only server-side when the
field was filled. Now the email is always checked on the client side
with a (very lenient) regex pattern, that says "field must contain at
least two characters, separated by an at-sign".

The error does not exist on cat17, no need to change that skin.

This is for https://phabricator.wikimedia.org/T180339


wiese:
> The error does not exist on cat17, no need to change that skin.

Not **yet**!

PR #1039, opened 2017-11-24
Gitignore package-lock.json

JeroenDeDauw:


gbirke:
As we're in the process of standardizing on npm 5+, and the [official recommendation is to check in the lock files](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md), let's talk about why we'd want to ignore the file instead.

JeroenDeDauw:
I don't really care about if it is in git or ignored... what I do not want is it not being in git and not being ignored, so that every time I run the build command I get the file showing up as "new file not in git".

PR #1041, opened 2017-11-24
Use new donor validation code from the Donation BC

JeroenDeDauw:
https://phabricator.wikimedia.org/T180936

Uses https://github.com/wmde/fundraising-donations/pull/27


tests/EdgeToEdge/Routes/ValidateAddressRouteTest.php
```diff
@@ -42,7 +42,7 @@ public function testGivenInvalidCompanyAddress_validationReturnsErrorMessage():
 		$expectedResponse = [
 			'status' => 'ERR',
 			'messages' => [
-				'company' => 'field_required'
+				'companyName' => 'missing'
```
JeroenDeDauw: I do not know if this change is fine. I tried finding code using this data but was unable to. Is it used somewhere in the UI?

app/routes.php
```diff
@@ -77,7 +77,7 @@ function( Request $request ) use ( $app, $ffFactory ) {
 $app->post(
 	'validate-address', // Validates donor information. This route is named badly.
 	function( Request $request ) use ( $app, $ffFactory ) {
-		return ( new ValidateAddressHandler( $ffFactory, $app ) )->handle( $request );
+		return ( new ValidateDonorHandler( $ffFactory, $app ) )->handle( $request );
```
gbirke: Just a side note: the `/validate-address` route is also called by the membership form. Since the URL and the underlying code did not change substantially, it will continue to work, but calling `ValidateDonorHandler` could make it weird, especially since the `ApplyForMembershipHandler` does its own address data validation, which should also not be different from `/validate-address`.

tests/EdgeToEdge/Routes/ValidateAddressRouteTest.php
gbirke: This is fine & fixes a potential problem. "Potential" because the JS in the forms does not call "validate-address" before all fields are filled, so we never got the wrong field name back. This is why we never caught this.

app/routes.php
JeroenDeDauw: Ugggggghhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

How about having a `/validate-donor` and a `/validate-applicant`?

gbirke: Yes, two routes would be a clean solution. Although it's not as DRY as I like, I could live with 100% identical code in the two validators, because they are two contexts.

PR #1042, opened 2017-11-24
T177930 Summary widgets

wiese:
Replaces #1036 

- [x] no more interval
- [x] refactored into reusable building blocks indicating form progress
- [x] removed post-submit error field marking - can only submit when valid, so should not happen
- [x] resolves T180215 for the donation process (but let's double-check)
- [x] tests for widgets and proxy
- [x] inject appropriate validation states, lose validity checks inside view handlers

Possible follow-ups (not in this PR anymore)
- [ ] further standardize state_aggregation/* validator result 
- [ ] state_aggregation tests


gbirke:
See https://phabricator.wikimedia.org/T181482

PR #1044, opened 2017-11-27
Fix double title on membership confirmation page

JeroenDeDauw:
https://phabricator.wikimedia.org/T181392

fullName already includes the title

```php
'fullName' => $applicant->getName()->getFullName(),
```

```php
	public function getFullName(): string {
		return join( ', ', array_filter( [
			$this->getFullPrivatePersonName(),
			$this->companyName
		] ) );
	}

	private function getFullPrivatePersonName(): string {
		return join( ' ', array_filter( [
			$this->getTitle(),
			$this->getFirstName(),
			$this->getLastName()
		] ) );
	}
```


JeroenDeDauw:
I did not verify if the issue is actually fixed as I am unable to submit the membership form successfully.

image

PR #1045, opened 2017-11-27
Use integers for storing donation amount

gbirke:
* Change numeric keypress handling:
  * Allow comma, dot and numpad decimal separator (locale-independent)
  * Allow left and right arrow keys
  * Suppress separators that have the wrong locale or are already present
    in the input field
* The amount input field value is now in the configured locale.
* Create converter functions for converting from string to integer and
  back, using a locale.
* Remove simple_currency_formatter, it's superses by teh formatter and
  parser in integer_currency


skins/cat17/src/app/lib/form_components.js
```diff
@@ -70,23 +70,16 @@ var objectAssign = require( 'object-assign' ),
...
+			this.hiddenElement.val( this.numberFormatter.format( formContent.amount ) );
 			if ( formContent.isCustomAmount ) {
 				this.inputElement.parent().addClass( 'filled' );
 				this.selectElement.prop( 'checked', false );
-				this.inputElement.val( formContent.amount );
+				this.inputElement.val( this.numberFormatter.format( formContent.amount ) );
 			} else {
 				this.inputElement.parent().removeClass( 'filled' );
-				this.selectElement.val( [ formContent.amount ] );
+				this.selectElement.val( [ this.numberFormatter.format( formContent.amount ) ] );
```
wiese: I just learned about array as a parameter to `.val()`. 
In order for this to work the elements inside `this.selectElement` (a colletion) are considered a group (even w/o a shared name?!), the one with the specified value being selected, the others unselected.
For this to work, of course, the value given has to match one of the input values.

tl;dr Deep-linking amounts does not work anymore, but we wanted to change something in the templates anyway...

skins/cat17/src/app/lib/form_validation.js
```diff
@@ -209,13 +208,15 @@ var jQuery = require( 'jquery' ),
...
 	 * @param {Function} sendFunction jQuery.post function or equivalent
 	 * @return {AmountValidator}
 	 */
-	createAmountValidator = function ( validationUrl, sendFunction ) {
+	createAmountValidator = function ( validationUrl, amountFormatter, sendFunction ) {
 		return objectAssign( Object.create( AmountValidator ), {
 			validationUrl: validationUrl,
-			sendFunction: sendFunction || jQuery.post
+			sendFunction: sendFunction || jQuery.post,
+			amountFormatter: amountFormatter || { format: function( v ) { return v; } }
```
wiese: I could not find a place where no `amountFormatter` is injected. Can we then please lose the vicious default (`||`)?

skins/cat17/src/app/lib/numeric_input_handler.js
```diff
@@ -2,47 +2,63 @@
...
-						);
-					}, 10
-				);
+			if (this.keyCodeIsOtherDelimiter( keyCode) || this.inputHasDelimiter() ) {
+				event.preventDefault();
 			}
+		},
+		keyCodeIsDelimiter: function ( keyCode ) {
+			return _.has( KEYCODE_DELIMITER_MAP, keyCode ) ||
+				keyCode === 110; // numpad
```
wiese: Is there a way to calculate the numpad code from the original (injected) code? Probably not. But we probably have to create a mapping of delimiter to key, so we may as well have a 1:n mapping there.

skins/cat17/src/app/lib/numeric_input_handler.js
```diff
@@ -2,47 +2,63 @@
...
 				// @see https://stackoverflow.com/a/41656511
 				keyCode = event.keyCode || event.which,
-				keysAllowed = [44, 46, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 0, 8, 9, 13];
+				keysAllowed = [
+					// numbers
+					48, 49, 50, 51, 52, 53, 54, 55, 56, 57,
+					// editing and navigation
+					0, 8, 9, 13, 37, 39, 46,
+					// Delimiters (regular & numpad)
+					188, 190, 110
```
wiese: I thought we'd just disallow all unfit delimiters, like any other key.

skins/cat17/src/app/lib/view_handler/fee_option_switcher.js
```diff
@@ -18,6 +18,7 @@ var objectAssign = require( 'object-assign' ),
 
 		update: function ( state ) {
 			_.each( this.elements, function ( feeOption ) {
+				// TODO no need to parseFloat when fee is converted to integer
 				var shouldBeDisabled = this.minimumFee[ state.addressType ] > feePerMonth( state.paymentIntervalInMonths, parseFloat( feeOption.val() ) ) ;
```
wiese: This has to be touched when we change the amount values in HTML, right?

gbirke: This line needs the same conversion as in donationForm. Also, the membership fee template needs the same "multiply value by 100" shenanigans as in the `payment_preset_amounts.html.twig`. But since we don't have that page yet, there is no way to add a "TODO"

gbirke:
refactored & amended after 1st review by @wiese

PR #1046, opened 2017-11-27
T180718 period radio 

wiese:
* only one set of radio buttons for payment interval (not a dedicated JS component any more)
* tests & some refactoring for form_data_extractor


JeroenDeDauw:
Less code and moar tests \o/

PR #1047, opened 2017-11-28
T180124 Print layout

wiese:
* print layout w/o footer
* lose print.css file
* add html class "client-nojs"/"client-js" to apply script dependend styling
* add a bootstrap-inspired class "hidden-no-js" which hides marked elements on
clients with JS disabled
* add print link to donation confirmation

Depends on https://github.com/wmde/fundraising-frontend-content/pull/54 - not tracked in composer to avoid conflicts


wiese:
Updated implementation. /cc @gbirke

gbirke:
After looking at https://phabricator.wikimedia.org/T180956, I saw that direct debit should get a different link text,  "Diese Bestätigung (inklusive SEPA-Mandat) ausdrucken".

gbirke:
Will merge this, https://phabricator.wikimedia.org/T180956 is its own, unrelated story.

PR #1049, opened 2017-11-28
Keep package-lock.json in version control

gbirke:
An alternative to #1039


wiese:
I like this better. Of course, https://github.com/npm/npm/issues/17722 will be with us more often, but it's the right thing to do.

wiese:
Please update to npm 5.6.0 to avoid problems with optional dependencies in package-lock.json http://blog.npmjs.org/post/167963735925/v560-2017-11-27

PR #1062, opened 2017-11-30
donator => donor

wiese:
for #1059


JeroenDeDauw:
huh... not sure how I missed something. Did a search for "donator" before committing.

PR #1063, opened 2017-11-30
T181273 Make text boxes react faster

gbirke:
Add wrapper function for text components that calls the change and
validation handler function on keypress events.


gbirke:
This solves a problem that was discovered during investigating https://phabricator.wikimedia.org/T181277, but the description there hints more at a problem with overlapping, unclickable elements, probably caused by https://phabricator.wikimedia.org/T181273

skins/cat17/src/app/lib/form_components.js
```diff
@@ -200,6 +201,16 @@ module.exports = {
 		bankDataElements.accountNumberElement.on( 'change', createDefaultChangeHandler( store, 'accountNumber' ) );
 		bankDataElements.bankCodeElement.on( 'change', createDefaultChangeHandler( store, 'bankCode' ) );
 		return component;
+	},
+
+	makeTextComponentMoreSnappy: function( textComponent, debounceDelay ) {
```
wiese: I get the point, but maybe `createEagerlyValidatingTextComponent` would be a better fit.

skins/cat17/src/app/tests/test_form_components.js
```diff
@@ -320,3 +320,29 @@ test( 'Bank data component checks if all elements in the configuration are set',
...
+
+	component.element.on.withArgs( 'keypress' ).args[ 0 ][ 1 ]( fakeEvent );
+
+	setTimeout( function () {
+		t.ok( component.onChange.calledOnce, 'change handler was called once' );
+		t.ok( component.onChange.calledWith( fakeEvent ), 'change handler was called with event' );
+
+		t.ok( component.validator.calledOnce, 'validator  was called once' );
+		t.ok( component.validator.calledWith( fakeEvent ), 'validator was called with event' );
+	}, 5 );
```
wiese: This plays with the hope that 1 < 5 and thus the debunce will always be finished before we test?!
Wouldn't it be nicer to stub out the `_.debounce` (it's not SUT) and perform a synchronous test?
For testing the integration with lodash (if that was the goal) we could [grab a hold of the timer functions](https://lodash.com/docs#runInContext).

skins/cat17/src/app/lib/form_components.js
```diff
@@ -1,9 +1,21 @@
...
 	NumericInputHandler = require( './numeric_input_handler' ),
 
+	/**
+	 * Wrapper around underscore debounce function with a delay of 300 milliseconds
+	 *
+	 * @param {function} f
+	 * @param {Number} milliseconds
+	 * @return {Function}
+	 */
+	defaultDebounce = function ( f, milliseconds ) {
```
JeroenDeDauw: This is never used?

PR #1064, opened 2017-12-01
T180431 Opt out of donation receipts

wiese:
Opt out of donation receipt for donation.

### DIFF
--- a/skins/cat17/src/app/lib/reducers/donation_form_content.js
+++ b/skins/cat17/src/app/lib/reducers/donation_form_content.js
@@ -25,7 +25,8 @@ var formContentLib = require( './form_content' ),
     country: 'DE',
     email: '',
     confirmNewsletter: false,
-    activePresets: false
+    activePresets: false,
+		donationReceipt: false
   };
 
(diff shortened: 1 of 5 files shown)


wiese:
virtualbox_ie10 - win7_01_12_2017_09_43_42

gbirke:
I changed the CSS to use flex again (with `flex-direction: column;`) and changed the `width` and `display` settings for the label.

wiese:
@gbirke aed4431 does resolve the IE10 checkbox issue but destroys the vertical alignment of labels with their radio buttons (current Firefox, Chrome, IE10). As this seems to be on the right trail, can we differentiate `[type="radio"] + label` and `[type="checkbox"] + label`? Maybe not elegant but effective in all our current scenarios?!

JeroenDeDauw:
Rebased

How are you testing this against IE?

gbirke:
https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

wiese:
Very cool. This looks fine now on all the mentioned browsers.

PR #1068, opened 2017-12-04
Remove deprecated template method for errors

gbirke:
Use getLayoutTemplate instead of getIncludeTemplate.

Had to adapt PiwikTest: when debug is enabled in the client,
there seems to be an internal error (which is why the test
worked before, because the donation/new template does not contain
the Piwik inlcude).

### DIFF
--- a/tests/EdgeToEdge/PiwikTest.php
+++ b/tests/EdgeToEdge/PiwikTest.php
@@ -13,13 +13,13 @@ class PiwikTest extends WebRouteTestCase {
 	const PIWIK_SITE_ID = 1;
 
 	public function testPiwikScriptGetsEmbedded(): void {
-		$client = $this->createClient( [], null, self::DISABLE_DEBUG );
+		$client = $this->createClient( [], null, self::ENABLE_DEBUG );
 		$client->request( 'GET', '/' );
 		$this->assertContains( '<!-- Piwik -->', $client->getResponse()->getContent() );
 	}
 
 	public function testConfigParametersAreUsed(): void {
-		$client = $this->createClient( [], null, self::DISABLE_DEBUG );
+		$client = $this->createClient( [], null, self::ENABLE_DEBUG );
 		$client->request( 'GET', '/' );
 
 		$this->assertContains( 'tracking.wikimedia.de', $client->getResponse()->getContent() );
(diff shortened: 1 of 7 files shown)


wiese:
This overlaps with #1066

tests/EdgeToEdge/PiwikTest.php
```diff
@@ -13,13 +13,13 @@ class PiwikTest extends WebRouteTestCase {
 	const PIWIK_SITE_ID = 1;
 
 	public function testPiwikScriptGetsEmbedded(): void {
-		$client = $this->createClient( [], null, self::DISABLE_DEBUG );
+		$client = $this->createClient( [], null, self::ENABLE_DEBUG );
```
JeroenDeDauw: These 3 arguments can now be omitted

JeroenDeDauw:
> Had to adapt PiwikTest: when debug is enabled in the client,
there seems to be an internal error (which is why the test
worked before, because the donation/new template does not contain
the Piwik inlcude).

I am very confoose. You changed it from disabled to enabled and the assertions are about containing something, not about not containing something. This is the opposite of what I expected from what you wrote.

JeroenDeDauw:
> This overlaps with #1066

@wiese note that this PR is a follow up into 1066. (Perhaps this is what you meant, though I first read it more as "this (might) conflict with 1066")

gbirke:
I'm also a bit confused, but from what I've seen I can explain the failing test as follows:

When `self::DISABLE_DEBUG` is set, there is some exception/output somewhere in the code (did not investigate much, but documented in https://phabricator.wikimedia.org/T181990). In the past, this led to the error page being displayed, which included the Base_Layout, which included the Piwik partial template, making the test pass (for the wrong reasons). 
Setting `self::ENABLE_DEBUG` does not cause the exception, displaying the Donation_Form template, which did not contain the Piwik data. 

The changes in this PR make the test pass again, for the right reasons, even when the test itself is silly until the E2E test use the production skin instead of the test one.

In the daily I've talked to the others and we agreed to tackle the underlying issue in the tests later and go with this for now.

wiese:
Realizing that membership confirmation uses `getIncludeTemplate` too, I became a fan of your investigation after all.

PR #1069, opened 2017-12-04
Make donationForm.js indent consistent

JeroenDeDauw:
Not sure if this should use tabs or 2 spaces. Both where used before. Though even with tab set to width 2, the indent was still inconsistent.

### DIFF
--- a/skins/cat17/src/scripts/donationForm.js
+++ b/skins/cat17/src/scripts/donationForm.js
@@ -1,320 +1,320 @@
 $( function () {
-  /** global: WMDE */
+	/** global: WMDE */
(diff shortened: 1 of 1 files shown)


gbirke:
Since we often still need to see where a line is coming from ("git blame" or IDE to see which parts of the file were touched in which commit by the 3rd party) we (i.e. Pablo an me) postponed this cleanup until we consider cat17 to be "done". Until then, we decided to live with inconsistent indentation and use tabs whenever we touch the code.

JeroenDeDauw:
Mawks sense

PR #1072, opened 2017-12-05
T181991 npm 5.6.0 to address ever-changing lock file

wiese:
* resolve deployment problem (changes of lock files on deploy preventing consecutive deployments)
* _should_ resolve lock files changing on either linux or apple systems due to naive handling of optional dependencies (https://github.com/npm/npm/issues/17671)

**Requires all of us** to re-provision our vagrant boxes (or what ever we do to run npm) to use the 5.6.0 [pre-release](https://github.com/npm/npm/releases).


wiese:
> We should revert this line to npm@latest when the new version is released officially.

This is disputable - "latest" is a moving target, essentially the opposite of pinning. I would argue that there should be a process in place to regularly (think security fixes) facilitate orchestrated updates (all projects, all machines), plus update if we realize the version we are using lacks a certain feature. But stumbling into updates when doing unrelated work (running the same old command at a different moment in time) doesn't really help anyone and causes headaches.

wiese:
Deployed merged master to both, test & prod without git diffs inside /tmp/fundraising-build-*

@gbirke How is your macOS dealing with the new package-lock.json?

PR #1078, opened 2017-12-05
Make dev entry point half-broken instead of fully-broken

JeroenDeDauw:


web/index.dev.php
```diff
@@ -81,40 +81,40 @@
...
-
-$app['twig.loader.filesystem'] = $app->extend(
-	'twig.loader.filesystem',
-	function ( $loader ) use ( $ffFactory ) {
-		/** @var \Twig_Loader_Filesystem $loader */
-		$loader->addPath( $ffFactory->getTemplatePath(), 'FunProfiler' );
-		return $loader;
-	}
-);
+//$app->register(
```
gbirke: I don't like commented out code because it is not obvious why it is there and why it was commented out (at least not without using the version history). Could you remove it instead in a single commit that can be referred to and reverted? You could then create an issue "Repair Dev entry point for fundraising frontend" and mention the commit. The issue can then also be referred to in the README.

JeroenDeDauw: I agree in general with not leaving commented code, though in this case for some reason I'm tempted to leave it there. Anyway, not that important so I just removed it now.

PR #1082, opened 2017-12-06
T181231 Error messages from content

wiese:
* error messages for donation
* error messages for membership (branch name is misleading)
* error message for shared component

Depends on (untracked) https://github.com/wmde/fundraising-frontend-content/pull/61

### DIFF
--- a/skins/cat17/src/scripts/donationForm.js
+++ b/skins/cat17/src/scripts/donationForm.js
@@ -163,10 +163,6 @@ $( function () {
         stateKey: 'donationInputValidation.postcode'
       },
       {
-        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '#post-code' ) ),
-        stateKey: 'donationInputValidation.postcode'
-      },
-      {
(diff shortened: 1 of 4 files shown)


skins/cat17/src/scripts/donationForm.js
```diff
@@ -162,10 +162,6 @@ $( function () {
         viewHandler: WMDE.View.createFieldValueValidityIndicator( $('#post-code-company') ),
         stateKey: 'donationInputValidation.postcode'
       },
-      {
-        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '#post-code' ) ),
-        stateKey: 'donationInputValidation.postcode'
-      },
```
JeroenDeDauw: This was unused?

JeroenDeDauw:
+1

skins/cat17/src/scripts/donationForm.js
wiese: Yes an no. There is a second identical call on l158 which should suffice.

PR #1083, opened 2017-12-06
T181732 align amount-typed field with other inputs

wiese:
* align amount-typed field with other inputs
* take currency symbol from template instead of hard-coding in CSS [0]

[0] https://caniuse.com/#feat=css-gencontent

### DIFF
--- a/skins/cat17/src/sass/layouts/_landing.scss
+++ b/skins/cat17/src/sass/layouts/_landing.scss
@@ -136,31 +136,11 @@ form {
                 height: 60px;
                 width: 100%;
                 position: relative;
-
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
+1

skins/cat17/src/sass/layouts/_landing.scss
```diff
@@ -187,12 +167,13 @@ form {
                     }
                 }
 				&:after {
-					content: "€";
+					content: attr( data-currency );
```
gbirke: Ooh, fancy! Did not know something like this is possible! 👍

PR #1085, opened 2017-12-07
T182116 Validate donation amount

wiese:
* fix address validation declaring all validators valid (not just the respective ones)
* validate only donation amount, not payment type, on server side
* use only one (aggregate) truth to determine color of submit button and ability to submit in donation
* mind https://github.com/wmde/FundraisingFrontend/pull/1090/files#diff-0f6e3fca475abf15e74a782294472ab1R342

Surely there will be a lively discussion considering the server-side implementation.

### DIFF
--- a/src/Validation/ConstraintViolationListMapper.php
+++ b/src/Validation/ConstraintViolationListMapper.php
@@ -1,0 +1,37 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Validation;
+
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+
+/**
+ * Map a list of ConstraintViolations into an array of messages grouped by the property causing them
+ *
+ * Class ConstraintViolationListMapper
+ * @package WMDE\Fundraising\Frontend\Validation
+ */
+class ConstraintViolationListMapper {
+
+	private $propertyAccessor;
+
+	public function __construct() {
+		$this->propertyAccessor = PropertyAccess::createPropertyAccessor();
+	}
+
+	public function map( ConstraintViolationListInterface $violations ): array {
+		$errors = [];
+
+		foreach ( $violations as $violation ) {
+			/** @var ConstraintViolationInterface $violation */
+			$entryErrors = (array) $this->propertyAccessor->getValue( $errors, $violation->getPropertyPath() );
+			$entryErrors[] = $violation->getMessage();
+			$this->propertyAccessor->setValue( $errors, $violation->getPropertyPath(), $entryErrors );
+		}
+
+		return $errors;
+	}
+}
--- a/tests/EdgeToEdge/Routes/ValidateAmountRouteTest.php
+++ b/tests/EdgeToEdge/Routes/ValidateAmountRouteTest.php
@@ -1,0 +1,52 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\EdgeToEdge\Routes;
+
+use Symfony\Component\HttpFoundation\Request;
+use WMDE\Fundraising\Frontend\Tests\EdgeToEdge\WebRouteTestCase;
+
+class ValidateAmountRouteTest extends WebRouteTestCase {
+
+	private const PATH = '/validate-donation-amount';
+
+	/**
+	 * @dataProvider getPassingTestData
+	 */
+	public function testGivenValidParameters_successResponseIsReturned( array $parameters ): void {
+		$client = $this->createClient();
+		$client->request( Request::METHOD_POST, self::PATH, $parameters );
+		$this->assertJsonSuccessResponse( [ 'status' => 'OK' ], $client->getResponse() );
+	}
+
+	public function getPassingTestData(): array {
+		return [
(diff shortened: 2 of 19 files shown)


wiese:
Hi @gbirke,
could you have a look at 2874c9a, please and check if I am not completely off (irrespective of _how_ the PoC is implemented)? It appears to me as if AddressValidation spoils all validation fields if it comes back `OK`.
Same code in 10h16, but probably a scenario not relevant because the address fields are reached only after all others are positively validated already...

gbirke:
2874c9a is totally valid and a good catch.

wiese:
For the record, 2874c9a was slightly reworked to c76c6d3.
I looked into how `input_validation` could learn about groups of validators and spent way too much time tinkering with how to define these groups in e.g. `donation_input_validation` and injecting them into the generic validation - but this takes lots of time of time to do right, and even more time to apply to the other cases (amount, bankData, ...) where `input_validation` talks about properties it should not really know about. Left a todo, opted out for the moment.

src/Validation/ConstraintViolationListMapper.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+
+/**
+ * Map a list of ConstraintViolations into an array of messages grouped by the property causing them
+ *
+ * Class ConstraintViolationListMapper
+ * @package WMDE\Fundraising\Frontend\Validation
```
JeroenDeDauw: Autogenerated clutter?

tests/EdgeToEdge/Routes/ValidateAmountRouteTest.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @dataProvider getFailingTestData
+	 */
+	public function testGivenInvalidParameters_matchingFailureResponseIsReturned( array $parameters, array $violations ): void {
+		$client = $this->createClient();
+		$client->request( Request::METHOD_POST, self::PATH, $parameters );
+		$this->assertErrorJsonResponse( $client->getResponse() );
+		$this->assertEquals( $violations, $this->getJsonFromResponse( $client->getResponse() )['messages'] );
+	}
+
+	public function getFailingTestData(): array {
```
JeroenDeDauw: AFAIK the PHPUnit naming convention is to use `failingTestDataProvider`. It certainly is what we have been doing at WMDE since we started using PHPUnit.

gbirke:
One issue that popped into my mind: The validators return english messages. Our translation system is based on language-independent message keys. Can we somehow configure the validators to return message keys?

src/Validation/ConstraintViolationListMapper.php
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	public function map( ConstraintViolationListInterface $violations ): array {
+		$errors = [];
+
+		foreach ( $violations as $violation ) {
+			/** @var ConstraintViolationInterface $violation */
+			$entryErrors = (array) $this->propertyAccessor->getValue( $errors, $violation->getPropertyPath() );
+			$entryErrors[] = $violation->getMessage();
+			$this->propertyAccessor->setValue( $errors, $violation->getPropertyPath(), $entryErrors );
```
JeroenDeDauw: This by-ref mutation is kinda ick

JeroenDeDauw:
Made the suggested CS changes in https://github.com/wmde/FundraisingFrontend/pull/1093

JeroenDeDauw:
Will merge already but please have a look at my earlier comment that is now in bold.

wiese:
Symfony validators are easily translated (of course incl. the option for non-human-friendly robot messages): https://symfony.com/doc/current/validation/translations.html
As there is no (to me) known consumer of the robot messages (only the error count is relevant) I went with the much more illustrative default messages.

> What made you go with this approach?

Honest answer? It was the quickest solution that works under the awareness that any solution short of integration with wmde/fun-validators is a temporary one, and I don't think it's beautiful either.

gbirke:
message keys might be needed when we tackle https://phabricator.wikimedia.org/T182415, that's why I pointed it out.

JeroenDeDauw:
> Honest answer?

Hopefully? :D

> It was the quickest solution that works under the awareness that any solution short of integration with wmde/fun-validators is a temporary one, and I don't think it's beautiful either.

I'm not sure why we would want to integrate this with `wmde/fun-validators` somehow. What makes you bring that up? Do you think we should use the same validation code for all UCs? I think that while this is nice in a way, it is not worth the effort as the gains would be small.

wiese:
> What makes you bring that up?

The server-side validation receiving the donation POST request should definitely be aligned with the validator used to judge the ajax request - so some form of shuffling about in the backend is required.

`wmde/fundraising-donations/contexts/PaymentContext/src/Domain/PaymentDataValidator.php` is conceptually not too far off from `wmde/fun-validators/src/Validators/AmountPolicyValidator.php`. Now we have a third place where a similar subject is dealt with.
If I understood @gbirke correctly he is also in favor of rethinking the validation concept and using _a_ 3rd party library and focus on configuration/instantiation rather than implementation - if that then still needs a custom repository is up for debate once a joint course is decided.

> Do you think we should use the same validation **code** for all UCs?

Code? Yes, sure. Of course with the respective configuration. I see little benefit of maintaining our own implementation(s) of the same concepts over and over while the building blocks are readily available as OSS.

PR #1089, opened 2017-12-08
T182098 - Add descriptions in membership state bar on mobile

gbirke:
Content changes in https://github.com/wmde/fundraising-frontend-content/pull/62

### DIFF
--- a/skins/cat17/src/scripts/membershipForm.js
+++ b/skins/cat17/src/scripts/membershipForm.js
@@ -177,7 +177,7 @@ $( function () {
 		},
 		{
 			viewHandler: WMDE.View.SectionInfo.createMembershipTypeSectionInfo(
-				$( '.member-type' ),
+				$( '.state-bar-lateral .member-type, .state-bar-detailed .member-type' ),
 				{
 					'sustaining': 'icon-favorite',
 					'active': 'icon-flash_on'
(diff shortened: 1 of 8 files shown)


skins/cat17/templates/partials/membership/state_bar_sections.html.twig
```diff
@@ -1,6 +1,6 @@
 <li class="member-type">
 	<a href="#membership-type">
-		<span class="text" data-empty-text="{$ 'membership_section_membership_type_anchor' | trans | e( 'html_attr' ) $}"></span>
+		<span class="text" data-empty-text="{$ membership_type_empty_text | default( 'membership_section_membership_type_anchor' ) | trans | e( 'html_attr' ) $}"></span>
```
wiese: Only if you really share my opinion on second thought... I am not a huge fan of defaults (i.e. optional parameters) in templates, in this case the `membership_type_empty_text`'s name does not even indicate it's a translation key (sounds more like an already translated message), is run through trans again.

skins/cat17/templates/Membership_Application.html.twig
```diff
@@ -92,8 +92,8 @@
 							<div class="wrap-field">
 								<div class="wrap-input">
 									<input type="radio" name="membership_type" id="aktive" value="active">
-									<label for="aktive">
-										<span data-short-name="Aktive">{$ 'membership_membershiptype_option_active' | trans $}</span>
+									<label for="aktive"  data-short-text="{$ 'membership_membershiptype_option_active_short' | trans $}">
```
wiese: That is two spaces, right?

skins/cat17/src/scripts/membershipForm.js
```diff
@@ -187,21 +187,30 @@ $( function () {
...
 			]
 		},
 		{
 			viewHandler: WMDE.View.SectionInfo.createDonorTypeSectionInfo(
-				$( '.donor-type' ),
+				$( '.state-bar-lateral .donor-type, .state-bar-detailed .donor-type' ),
 				{
 					'person': 'icon-account_circle',
-					'firma': 'icon-work',
-					'anonym': 'icon-visibility_off'
```
wiese: For the record: this is a desired change, `anonym` is not an option for memberships.

PR #1090, opened 2017-12-08
T180690 Scroll to next relevant section and display infobox

gbirke:
Replaces #1080 

Related content change: https://github.com/wmde/fundraising-frontend-content/pull/64

### DIFF
--- a/skins/cat17/src/app/lib/scrolling.js
+++ b/skins/cat17/src/app/lib/scrolling.js
@@ -22,9 +22,14 @@ var objectAssign = require( 'object-assign' ),
 
 	AnimatedScroller = {
 		fixedHeaderElements: null,
-		scrollTo: function( $element ) {
+		scrollTo: function( $element, options ) {
+			var topOffset = calculateElementOffset( $element, this.fixedHeaderElements );
+			if ( options && options.additionalOffset ) {
+				topOffset += options.additionalOffset;
+			}
+
 			$( 'html, body' ).stop( true ).animate( {
-				scrollTop: calculateElementOffset( $element, this.fixedHeaderElements )
+				scrollTop: topOffset
 			}, 1000, function () {
 				// Callback after animation
 				// Must change focus!
--- a/skins/cat17/src/app/lib/scrolling.js
+++ b/skins/cat17/src/app/lib/scrolling.js
@@ -107,7 +107,7 @@ module.exports ={
 	},
 	scrollOnSuboptionChange: function( $suboptionInput, $suboptionContainer, scroller ) {
 		$suboptionInput.on( 'change', function ( evt ) {
-			var wrapper = $suboptionContainer.find( '.wrap-field input[value=' + evt.target.value + ']' ).parents( '.wrap-field' );
+			var wrapper = $suboptionContainer.find( '.wrap-field input[value=' + evt.target.value + ']' ).parents( '.wrap-field' ).find( '.info-text' );
 			if (wrapper.length) {
 				scroller.scrollTo( wrapper, { elementStart: ElementStart.ELEMENT } );
 			}
--- a/skins/cat17/src/app/lib/scrolling.js
+++ b/skins/cat17/src/app/lib/scrolling.js
@@ -3,33 +3,65 @@
 var objectAssign = require( 'object-assign' ),
 	_ = require( 'underscore' ),
 
-	calculateElementOffset = function ( $element, $fixedHeaderElements ) {
-		var offset = 0,
-			matchedElemPadding = $element.css( 'padding-top' ).match( /^(\d+)px$/ );
-		_.each( $fixedHeaderElements.get(), function ( elm ) {
-			var $elm = $( elm );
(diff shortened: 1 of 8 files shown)


skins/cat17/templates/Donation_Form.html.twig
```diff
@@ -52,8 +30,17 @@
...
-
 				<div class="form-shadow-wrap col-xs-12 col-sm-8">
+					<section class="introduction-banner hidden">
+						<p>
+							{$ 'donation_banner_deeplink_summary' | trans( {
+							'%frequency%': '<span class="frequency"><span class="text"></span></span>',
+							'%amount%': '<span class="amount"><span class="text"></span></span>',
+							'%payment_method%': '<span class="payment-method"><span class="text"></span></span>'
+							} ) | raw $}
+							<a href="#donation-amount">{$ 'donation_banner_correction_anchor' | trans | raw $}</a>
```
wiese: We could lose the `raw` now. No raw is the best raw.

skins/cat17/src/scripts/donationForm.js
```diff
@@ -333,18 +333,23 @@ $( function () {
...
+
+	// Scroll to first required element that needs to be filled
+	var currentState = store.getState();
+	if ( currentState.validity.paymentData ) {
+		var nextRequired = currentState.donationFormContent.paymentType === 'BEZ' ? $( '#payment-method' ) : $( '#donation-type' );
+		var $introBanner = $('.introduction-banner');
+		$introBanner.insertBefore( nextRequired ).removeClass( 'hidden' );
+		// 24px = 1 em padding of intro-banner that is removed automatically by scroller, plus .5em whitespace above
+		var scrollOptions = { additionalOffset: -24 };
+		scroller.scrollTo( $introBanner, scrollOptions );
```
wiese: Maybe move the `scrollOptions` object right into the .scrollTo? We still have the comment to explain it. OTOH, please see my comment for `scrolling.js`.

skins/cat17/src/scripts/donationForm.js
```diff
@@ -333,18 +333,23 @@ $( function () {
...
-    $introDefault.addClass('hidden');
-  }
-
 	// Non-state-changing event behavior
 
 	var scroller = WMDE.Scrolling.createAnimatedScroller( $( '.wrap-header, .state-bar' ) );
+
+	// Scroll to first required element that needs to be filled
+	var currentState = store.getState();
+	if ( currentState.validity.paymentData ) {
```
wiese: This will probably be too simple after #1085. We will see - it's good to see the prototypical code replaced.

skins/cat17/src/sass/layouts/_introduction.scss
```diff
@@ -39,3 +39,19 @@
 	padding-top: 0;
 }
 
+.introduction-banner {
+	border: 1px solid $brand-primary;
+	padding: 1em 1em 1em 2em;
+	color: #494f51;
```
wiese: Where does this come from? Does it take inspiration from `$gray-line`?

skins/cat17/src/app/lib/scrolling.js
```diff
@@ -22,9 +22,14 @@ var objectAssign = require( 'object-assign' ),
 
 	AnimatedScroller = {
 		fixedHeaderElements: null,
-		scrollTo: function( $element ) {
+		scrollTo: function( $element, options ) {
+			var topOffset = calculateElementOffset( $element, this.fixedHeaderElements );
+			if ( options && options.additionalOffset ) {
```
wiese: I thought this class was to encapsulate the offset calculation. Which circumstances led to it now requiring an external information instead, which would not better be _calculated_ (instead of hard coded) by this class (instead of somewhere else)?

gbirke: `calculateElementOffset` removes the padding of the elements it scrolls to, which is important for the scrolls from the info sections (as the form sections have a big padding above them). The info box has a border and a padding, hence the padding needs to be re-added. 
Possible solutions:

* fiddle with the CSS/Markup (of the form sections, not the infobox) to find better scroll points, dropping the subtraction of padding.
* instead of adding explicit offsets, have a `options.subtractPadding` boolean flag that is given to `calculateElementOffset`
* ???

What do you think?

skins/cat17/src/sass/layouts/_introduction.scss
gbirke: This comes [from Jan](https://phabricator.wikimedia.org/T180690#3822646), who wanted the text to be a bit lighter than the normal text. I've asked in the issue what color values he wants, if we stay in the given palette or not.

skins/cat17/templates/Donation_Form.html.twig
gbirke: Does the up arrow in the msg show up correctly? Otherwise we might have to use the `&uparrow;` entity and need raw. I'll check on my platform, please check on yours.

PR #1091, opened 2017-12-08
T179938 basic layout for supporters list

wiese:
* does NOT work on small screens yet
* does NOT show comment icon (particular not only when comment available)
* is NOT click-based but only reacts to hover

What do you think about CSS `hyphens`?

Depends on (untracked) https://github.com/wmde/fundraising-frontend-content/pull/63


wiese:
Follow-up https://phabricator.wikimedia.org/T182698

PR #1092, opened 2017-12-08
T182121 membership application initial values from donation

wiese:
- [x] ability to convert donation info to membership's initial form values
- [x] set membership's initial form values on `/apply-for-membership` (e.g. `/apply-for-membership?donationId=19&donationAccessToken=ce6d215305f1614b8dd54e691618b8e2&type=sustaining`)
- [x] ensure initial state is rendered nicely in membership application form & widgets
- [x] ensure fields with initial state are shown as validated (`state.membershipInputValidation`)
- [x] ensure form is submittable once fields not part of initial state are provided (`state.validity`)

Time is an object. This works but needs follow-ups, please see my comment.


wiese:
Once initial values are provided there are the following challenges:
- presist the initial state inside the state = `membershipFormContent` values
- present to new state to the user inside the form (incl. the correct select option for e.g. salutation selected - this is based on jcf and requires a change event on the DOM element)
- show validation state of the provided info on the form fields (e.g. check mark and background color next on first name input) = `membershipInputValidation` values
- calculate global validity values based on ajax responses = `validity` values

JeroenDeDauw:
+2 to the first 2 commits. (I was expecting something totally horrendous after reading the PR description... what is there seems reasonable enough to me.)

+0.42 to the 3rd commit. I don't really get the `SelectComponent` part even with the description and now does not seem like a good point to invest a lot of time into that.

wiese:
Linking done in #1108's 04b9253

PR #1106, opened 2017-12-12
T182405 Validate address input length

gbirke:
Validate address fields with patterns.


wiese:
We talked about the `addEagerChangeBehavior` for memberships yesterday. Do you think the same logic we found true for donation applies there?

gbirke:
It's not as bad as in the donation form where the city is the last field and the `addEagerChangeBehavior` is the solution to that. But for consistencies sake I can use this PR to add it. But I have to take care that the message does not pup up too early, e.g. while you're in the process of filling out the postcode field which now validates the postcode length.

wiese:
...and the street/house number (T180042).

gbirke:
Rebased for another PR that will be based on this.

PR #1109, opened 2017-12-13
T182474 Hide icons in mobile bar

gbirke:
Add rule for hiding icons in the status bar on mobile when the state
that's displayed is invalid.


gbirke:
Thanks, I did another commit that removes the opacity statements

PR #1116, opened 2017-12-15
Fix response status check in comment form

JeroenDeDauw:
Apparently I did not push this commit to GH yesterday :)


gbirke:
When I tested it yesterday I still got a success message, I wonder why ...

PR #1117, opened 2017-12-15
DNM - Use path() twig function in comment form

JeroenDeDauw:


JeroenDeDauw:
Apparently I can no longer -2 myself using the review feature... the fuck GH. See https://github.com/wmde/FundraisingFrontend/pull/1118

wiese:
The thoughts on how routes are built inside the view apply to all our templates so it's probably fair to keep it consistent (#1117) until we change it (#1118).

PR #1118, opened 2017-12-15
DNM - use twig path() function

JeroenDeDauw:
While working on this supposed improvement I started thinking this is actually making things worse.

I think we should not have logic in the twig templates. This URL building might not be application or domain logic and be presentation specific (though not as much as the twig template), yet it is still logic. If the template constructs the URL it needs to know which data is in the URL and what the names of the parameters are. There is no need for the template for have this knowledge.

Consider the left bottom part of this Clean Architecture diagram:

Clean Architecture

The Presenter gets the Response Model from the UC (in our case the UC does not give this directly to the Presenter but that does not matter for the topic at hand). In the diagram the Presenter constructs a View Model, which like the Reponse Model of the UC is a Data Transfer Object, and hands it to the View. This object is supposed to contain all the data the View needs, in the form the view needs it. This means the view might contain loops and if statements to choose if some section should be shown or not, but no further logic. One important effect of this is that the views are so simple that they do not require automated testing.

I'm not sure if I ever brought that up already, yet I've been viewing the twig templates used directly by the presenters as having the role of the View in this diagram, and the argument list we provide to them as the View Model. If you follow that, then it is clear URL construction should go in the Presenter, not in the Twig template (View).

### DIFF
--- a/skins/cat17/templates/partials/donation_confirmation/direct_debit.html.twig
+++ b/skins/cat17/templates/partials/donation_confirmation/direct_debit.html.twig
@@ -63,7 +63,11 @@
     <div class="row">
         <nav class="menu menu-footer col-xs-12 col-sm-5 col-sm-offset-7 col-md-offset-8 col-md-4">
             <ul class="list-menu list-unstyled">
-                <li><a href="{$ commentUrl $}"><span>{$ "add_donation_comment" | trans $}</span></a></li>
+                <li>
+                    <a href="{$ path( 'commentUrl', { donation.id: donationId, updateToken: donation.updateToken, accessToken: donation.accessToken } ) $}">
+                        <span>{$ "add_donation_comment" | trans $}</span>
+                    </a>
+                </li>
                 {% include 'partials/share_on_social_networks.html.twig' %}
             </ul>
         </nav>
--- a/skins/cat17/templates/partials/donation_confirmation/external_payment.html.twig
+++ b/skins/cat17/templates/partials/donation_confirmation/external_payment.html.twig
@@ -53,7 +53,11 @@
     <div class="row">
         <nav class="menu menu-footer col-xs-12 col-sm-5 col-sm-offset-7 col-md-offset-8 col-md-4">
         <ul class="list-menu list-unstyled">
-            <li><a href="{$ commentUrl $}"><span>{$ "add_donation_comment" | trans $}</span></a></li>
+            <li>
+                <a href="{$ path( 'commentUrl', { donation.id: donationId, updateToken: donation.updateToken, accessToken: donation.accessToken } ) $}">
+                    <span>{$ "add_donation_comment" | trans $}</span>
+                </a>
+            </li>
             {% include 'partials/share_on_social_networks.html.twig' %}
         </ul>
         </nav>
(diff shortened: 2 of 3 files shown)


JeroenDeDauw:
Improved version of the diagram:

arch

gbirke:
In my understanding it's a tradeoff between ease of testing (not having to set up the whole Twig  environment with its framework integration) and ease of using existing facilities and libraries, making the code more compact. As long as we try to avoid duplicating functionality that the framework/library provides, I'm happy. 

I agree with you that generating the URL in the Presenter is the better way to do it, especially because the tests for the presenter can then check if the right variables are passed on to the URL generator.

One improvement I'd suggest is to use the UrlGenerator that comes from Silex (https://silex.symfony.com/doc/2.0/services.html#core-services) instead of grabbing it from the Twig extension. We could even think about dropping the interface, since Silex/Symfony already provides us with one.

One last comment on the topic of routes: Most routes in the templates are just plain strings relative to `{$ basepath $}`, which is brittle. I've created https://phabricator.wikimedia.org/T183006 to remind us of a better alternative.

JeroenDeDauw:
> In my understanding it's a tradeoff between ease of testing (not having to set up the whole Twig environment with its framework integration) and ease of using existing facilities and libraries, making the code more compact. As long as we try to avoid duplicating functionality that the framework/library provides, I'm happy.

Not using this logic in the template indeed has an advantage for testing, but it is not just this one thing you mentioned and it is also not limited to testing. Beyond the easy of setup in your tests you also have easy of understandability, less brittleness (big point) and faster execution speed (small point). Going beyond testing, the better defining of and separation of concerns has all its usual advantages.

When using framework code one should always be aware of the costs, and evaluate if there really is an advantage in using it. For usecases going beyond the superficial, this is often not the case. So I am very skeptical of trying to use framework code as much as possible if it does not explicitly include this caution. 

The particular case at hand here does not seem to be a question of using framework code or not. It's the difference between using an injected service (that uses framework code) in the Presenter or using a global function in the View.

JeroenDeDauw:
> One improvement I'd suggest is to use the UrlGenerator that comes from Silex (https://silex.symfony.com/doc/2.0/services.html#core-services) instead of grabbing it from the Twig extension.

Isn't that what we are doing? 

```php
    'cancelUrl' => $app['url_generator']->generate(
```

> We could even think about dropping the interface, since Silex/Symfony already provides us with one.

I'm sure I did not introduce this interface just for the sake of it. I did not double check this, but if I had to guess, I would say this interface was introduced to avoid coupling the top level factory to the framework. If so, that is plenty of gain for very little cost: just an interface, and perhaps a trivial Adapter, neither of which you need to care about pretty much anywhere.

gbirke:
>> One improvement I'd suggest is to use the UrlGenerator that comes from Silex (https://silex.symfony.com/doc/2.0/services.html#core-services) instead of grabbing it from the Twig extension.

> Isn't that what we are doing?

For the cancel URL yes, for the comment URL there is `app/UrlGenerator` which uses `Symfony\Bridge\Twig\Extension\RoutingExtension` (which is just a Twig wrapper around `Symfony\Component\Routing\Generator\UrlGenerator`)  instead of using the `Symfony\Component\Routing\Generator\UrlGenerator` provided by Silex in `$app['url_generator']`.

My proposition would be either to 
a) make `WMDE\Fundraising\Frontend\App\UrlGenerator` a wrapper for `Symfony\Component\Routing\Generator\UrlGenerator`
b) Drop `WMDE\Fundraising\Frontend\App\UrlGenerator` and its interface and use `Symfony\Component\Routing\Generator\UrlGeneratorInterface` in places where it's needed.

JeroenDeDauw:
https://github.com/wmde/FundraisingFrontend/pull/1130

PR #1126, opened 2017-12-15
cat all vendor js into one file, add file prefix to complex skin assets

wiese:


skins/cat17/gulpfile.js
```diff
@@ -3,14 +3,12 @@ var sass = require( 'gulp-sass' );
 var sourcemaps = require( 'gulp-sourcemaps' );
 var autoprefixer = require( 'gulp-autoprefixer' );
 var browserSync = require( 'browser-sync' ).create();
-var cssNano = require( 'gulp-cssnano' );
-var useref = require( 'gulp-useref' );
```
gbirke: We should remove these packages from `package.json`

PR #1129, opened 2017-12-16
T182418 Add padding to validated input fields

gbirke:
Add right padding to validated input fields so the checkmark cannot
overlap thee field contents.


JeroenDeDauw:
+1

PR #1133, opened 2017-12-18
T179938 ensure long text does not break layout

wiese:


skins/cat17/src/sass/layouts/pages/supporters.scss
```diff
@@ -28,6 +28,7 @@
 			height: 75px;
 			width: 170px;
 			padding-right: 15px;
+			hyphens: auto;
```
JeroenDeDauw: I had no idea this existed :)

JeroenDeDauw:
+1

PR #1138, opened 2017-12-18
Add error message to amount box in donation form

JeroenDeDauw:
https://phabricator.wikimedia.org/T182415

### DIFF
--- a/skins/cat17/src/scripts/donationForm.js
+++ b/skins/cat17/src/scripts/donationForm.js
@@ -138,68 +138,51 @@ $( function () {
 			stateKey: 'donationFormContent.addressType'
 		},
       {
-        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '#first-name' ) ),
+        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '.field-firstname' ) ),
         stateKey: 'donationInputValidation.firstName'
       },
       {
-        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '#last-name' ) ),
+        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '.field-lastname' ) ),
         stateKey: 'donationInputValidation.lastName'
       },
+		{
+			viewHandler: WMDE.View.createFieldValueValidityIndicator( $('.field-company') ),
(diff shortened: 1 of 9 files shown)


JeroenDeDauw:
+1 to Gabriels changes

wiese:
You're on fire, man - will look at and test-drive all that awesomeness first thing tomorrow.

JeroenDeDauw:
The overflow issue does not seem to be fixed yet:

image

skins/cat17/templates/Donation_Form.html.twig
```diff
@@ -53,6 +53,8 @@
 								<div class="wrap-amount-typed field-grp" data-currency="€">
 									<input type="text" id="amount-typed">
 								</div>
+								<span class="error-text">{$ 'form_amount_error' | trans $}</span>
```
wiese: Key still seems to be missing from the content repo.

skins/cat17/src/scripts/donationForm.js
```diff
@@ -138,68 +138,51 @@ $( function () {
 			stateKey: 'donationFormContent.addressType'
 		},
       {
-        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '#first-name' ) ),
+        viewHandler: WMDE.View.createFieldValueValidityIndicator( $( '.field-firstname' ) ),
```
wiese: I was very excited to see the ids replaced by class selectors - making half the view handlers redundant in the process. OTOH is adds a certain amount of redundancy (separate attributes only used by one view handler/component each) for now. Would you agree that the general direction, step by step, is to reduce this back to one unique way to identify each input, maybe even reuse elements selected once in the main dispatchers (à la page contract in e.g. `donationForm.js`)

skins/cat17/templates/Donation_Form.html.twig
JeroenDeDauw: Added in https://github.com/wmde/fundraising-frontend-content/pull/84

skins/cat17/src/scripts/donationForm.js
gbirke: Yes, that's a direction I'd like to take. Thanks for pointing it out.

PR #1140, opened 2017-12-18
T183114 Skip unhandled PayPal notifications more early

gbirke:
Instead of trying to construct a request with "invalid" data (i.e.
negative euro amounts), ignore requests we can't handle anyways.

To be more clear what kinds of requests are handled,
`HandlePayPalPaymentNotificationUseCase` should be renamed to
`HandlePayPalPaymentCompletionNotificationUseCase`.

### DIFF
--- a/app/RouteHandlers/PayPalNotificationHandler.php
+++ b/app/RouteHandlers/PayPalNotificationHandler.php
@@ -20,6 +20,8 @@ use WMDE\Fundraising\Frontend\PaymentContext\RequestModel\PayPalPaymentNotificat
  */
 class PayPalNotificationHandler {
 
+	private const MSG_NOT_HANDLED = 'PayPal request not handled';
+
 	private $ffFactory;
 
 	public function __construct( FunFunFactory $ffFactory ) {
@@ -38,6 +40,11 @@ class PayPalNotificationHandler {
 			return $this->createErrorResponse( $e );
 		}
 
+		if ( !$this->requestIsForPaymentCompletion( $post ) ) {
+			$this->ffFactory->getPaypalLogger()->log( LogLevel::INFO, self::MSG_NOT_HANDLED, [ 'post_vars' => $post->all() ] );
+			return new Response( '', Response::HTTP_OK );
+		}
+
 		$useCase = $this->ffFactory->newHandlePayPalPaymentCompletionNotificationUseCase( $this->getUpdateToken( $post ) );
 
 		$response = $useCase->handleNotification( $this->newUseCaseRequestFromPost( $post ) );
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Best to merge https://github.com/wmde/FundraisingFrontend/pull/1143 and then rebase this one

app/RouteHandlers/PayPalNotificationHandler.php
```diff
@@ -20,6 +20,8 @@
  */
 class PayPalNotificationHandler {
 
+	const MSG_NOT_HANDLED = 'PayPal request not handled';
```
JeroenDeDauw: DO YOU EVEN PHP 7.1?!

:)

app/RouteHandlers/PayPalNotificationHandler.php
```diff
@@ -55,6 +62,25 @@ private function getValueFromCustomVars( string $customVars, string $key ): stri
 		return !empty( $vars[$key] ) ? $vars[$key] : '';
 	}
 
+	private function requestIsForPaymentCompletion( ParameterBag $post ): bool {
+		return $this->isSuccessfulPaymentNotification( $post->get( 'payment_status' ) ) || (
+				$this->isForRecurringPayment( $post->get( 'txn_type' ) ) &&
+				!$this->isRecurringPaymentCompletion( $post->get( 'txn_type' ) )
```
JeroenDeDauw: I like how this is split, though the specification of the get value here now mixes levels of abstraction. This can be avoided by passing $post to the methods.

tests/EdgeToEdge/Routes/HandlePayPalPaymentNotificationRouteTest.php
```diff
@@ -352,7 +385,7 @@ public function testGivenNegativeTransactionFee_exceptionIsThrown(): void {
 			);
 
 			$this->assertSame(
-				[ 'Unhandled PayPal instant payment notification' ],
```
wiese: Both occurrences of this (unhandled from inside the uses cases) have been converted to tests against the new logic - so there is no more test for the guard inside handleNotification, right?

JeroenDeDauw: If I am understanding what you mean correctly, which I am not very sure of, then this is the test covering the guard:

image

JeroenDeDauw:
> Am I missing a PR on the PaymentContext or is this duplicating the code from PayPalPaymentNotificationRequest?

Huh, apparently it is duplication for now. The already merged PR is just a rename... https://github.com/wmde/fundraising-donations/pull/32/files

So this will need a small follow up PR once merged

JeroenDeDauw:
I'm happy to do a follow up if there is a missing test

PR #1142, opened 2017-12-18
T180110 basic no-js message

wiese:
Depends on https://github.com/wmde/fundraising-frontend-content/pull/82
  
Is more of a PoC to provoke requirements specification.

### DIFF
--- a/skins/cat17/templates/Base_Layout.html.twig
+++ b/skins/cat17/templates/Base_Layout.html.twig
@@ -35,6 +35,8 @@
 		{% include 'partials/' ~ footer_template %}
 	</div>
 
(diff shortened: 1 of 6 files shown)


skins/cat17/templates/Base_Layout.html.twig
```diff
@@ -35,6 +35,19 @@
 		{% include 'partials/' ~ footer_template %}
 	</div>
 
+	{% if page_works_without_js is not defined %}
+		<noscript>
+			<div class="fallback">
+				<div class="content">
+					<div class="container">
+						<h2>No script!</h2>
```
gbirke: The headline also needs to be put in the content repo.

PR #1144, opened 2017-12-19
Update to match latest version of AddCommentUC

JeroenDeDauw:
https://github.com/wmde/fundraising-donations/pull/31

I did not yet manually test this and we do not seem to have tests that cover the integration with the route or JS. Nomnomnom time now


gbirke:
I have temporarily enabled the 2nd option in cat17 and made comments. Both options worked. 10h10 also worked. +1

Why is the 2nd option still commented out in cat17?

JeroenDeDauw:
> Why is the 2nd option still commented out in cat17?

Cause I did not want to make master un-deployable, which it would be if there is a placeholder / not yet translated message. Next step is to re-add that option.

gbirke:
The message placeholder [is translated](https://github.com/wmde/fundraising-frontend-content/blob/test/i18n/de_DE/messages/messages.json#L91), you can activate it.

PR #1146, opened 2017-12-19
T180956 add more information to sepa mandate

wiese:
Depends on (untracked) https://github.com/wmde/fundraising-frontend-content/pull/83


wiese:
As the requirements are trickling in, some content was removed again, so two pages are not reached any more. Question about the NL text is in the ticket.

wiese:
Added some changes, should be finalized now.

PR #1147, opened 2017-12-19
T182790 Fix infobox offset calculation on mobile

gbirke:
Show state bar before scrolling to force its height being calculated in offset.

### DIFF
--- a/skins/cat17/src/scripts/donationForm.js
+++ b/skins/cat17/src/scripts/donationForm.js
@@ -336,6 +336,10 @@ $( function () {
 		var nextRequired = currentState.donationFormContent.paymentType === 'BEZ' ? $( '#payment-method' ) : $( '#donation-type .legend:first' );
 		var $introBanner = $('.introduction-banner');
 		$introBanner.insertBefore( nextRequired ).removeClass( 'hidden' );
+
+		// The fixed horizontal state bar is hidden initially and only shown when scrolling, which leads to wrong offset.
+		// Thus, we have to make it visible for offset calculation.
+		$( '.state-bar' ).addClass( 'active' );
 		scroller.scrollTo( $introBanner, { elementStart: WMDE.Scrolling.ElementStart.MARGIN } );
 	}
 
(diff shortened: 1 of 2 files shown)


skins/cat17/src/scripts/donationForm.js
```diff
@@ -336,6 +336,12 @@ $( function () {
 		var nextRequired = currentState.donationFormContent.paymentType === 'BEZ' ? $( '#payment-method' ) : $( '#donation-type .legend:first' );
 		var $introBanner = $('.introduction-banner');
 		$introBanner.insertBefore( nextRequired ).removeClass( 'hidden' );
+
+		// On mobile, the state bar is hidden initially and only shown when scrolling, which leads to wrong offset.
+		// Thus, we have to make it visible for offset calculation
+		if ( $( window ).width() < 1023 ) {
```
wiese: This could be a little less painful to read (and would spare us from hard-coding the 1023 again) if we just always added the active class without this check - the comment does a good job at pointing out what it's good for, and desktop does not seems to have a problem with the class added while the state-bar is hidden.

skins/cat17/src/scripts/main.js
```diff
@@ -127,14 +127,6 @@
                 if (currentScroll + 70 >= fixBarTop) {
                     $('.state-bar').addClass('active');
                     $('.fixed-button').addClass('active');
-                    $('.menu-main').addClass('under-bar');
```
wiese: This looks sort of scary but I could not find a use either.

skins/cat17/src/scripts/main.js
```diff
@@ -127,14 +127,6 @@
...
-
-                    if ($('.footer').isVisible('top')) {
-                        $('.menu-main').removeClass('under-bar');
-                    } else {
-                        $(".state-bar").addClass('active');
```
wiese: Can not fathom how this ever helped us. Do you have a theory?

gbirke: My theory is that it was meant to avoid the right state "bar" to overlap the footer. But that is already taken care of in another place and this is just a vestige.

PR #1151, opened 2017-12-20
T182963 payment section validity

wiese:
Based on #1149

Could not stand two branches of the if resulting in true, so I took on T182963 and incorporated this.


wiese:
@gbirke Please reassess.

PR #1153, opened 2017-12-21
T182419 - Set header element class during scrolling

gbirke:
The header status bar becomes only visible (with the CSS class `active`)
after a certain scroll position which messes with the calculation of the
final scroll offset.

This change introduces the `scrolling` CSS class that is only active on
an element when scrolling iwth the `scrollTo` method. By giving it the
same CSS properties as `active`, the element has the proper height
before the offset is calculated.

This also improves on the kind of hackish solution of https://phabricator.wikimedia.org/T182790


wiese:
Added some tests for AnimatedScroller and [this particular case](https://github.com/wmde/FundraisingFrontend/compare/cat17-class-during-scrolling...cat17-class-during-scrolling-test?expand=1#diff-a8530160dc88aa6eff76245fda652537R370). #1154

PR #1155, opened 2017-12-21
check data attribute exists to prevent vagueness

wiese:
For #1152

Got scared by the native .getAttribute(), apparently rightfully so
https://developer.mozilla.org/de/docs/Web/API/Element/getAttribute#Notes


wiese:
Rebased

PR #1156, opened 2017-12-21
T182425 - Fix initial validation state

gbirke:
This also solves https://phabricator.wikimedia.org/T182967


src/Presentation/DonationMembershipApplicationAdapter.php
```diff
@@ -51,4 +52,20 @@ private function getMembershipFormBankDataValues( PaymentMethod $paymentMethod )
 			'bankname' => $paymentMethod->getBankData()->getBankName(),
 		];
 	}
+
+	public function getInitialValidationState( Donation $donation ): array {
+		$validationState = [];
```
wiese: Was it a conscious decision not start with null values for the validation keys? I agree that it should not make a difference?

src/Presentation/DonationMembershipApplicationAdapter.php
```diff
@@ -51,4 +52,20 @@ private function getMembershipFormBankDataValues( PaymentMethod $paymentMethod )
...
+		if ( $donation->getPaymentType() !== PaymentType::DIRECT_DEBIT ) {
+			return $validationState;
+		}
+		/** @var DirectDebitPayment $paymentMethod */
+		$paymentMethod = $donation->getPayment()->getPaymentMethod();
+		if ( $paymentMethod->getBankData()->hasIban() ) {
+			$validationState['bankData'] = true;
+		}
+		return $validationState;
+	}
```
JeroenDeDauw: Why did you put the code here? I noticed there is no cohesion with the other code in this class and then could not readily come up with a reason to put it here rather than in the route.

JeroenDeDauw: If the goal was to make testing easier (not needing an e2e test), then how about putting this in a static method in the route? Static is generally not nice though in this case you would not increase the static binding (the only user directly constructs the object), assuming there is no plan to use this in other places (not sure if that even makes sense).

gbirke: > Why did you put the code here? 

I thought of `DonationMembershipApplicationAdapter` as "a class that prepares **all presentation data** for a membership application from a Donation object" rather than "a class that prepares the prepares **only the presentation values** for a membership  application from a Donation object". Everything is conveniently on one place and */* could live with a presentation class that has two methods that do slightly different things.

I don't know what you mean by "putting this in a static method in the route". Do you mean converting the route function into a Handler class, that has a static function that can be tested?

skins/cat17/src/scripts/membershipForm.js
```diff
@@ -344,19 +344,19 @@ $( function () {
...
-
-		store.dispatch( actions.newChangeContentAction( key, value ) );
-	} );
+	if ( typeof initSetup.amount === 'string' ) {
+		initSetup.amount = WMDE.IntegerCurrency.createCurrencyParser( 'de' ).parse( initSetup.amount );
+	}
+	store.dispatch( actions.newInitializeContentAction( initSetup ) );
 
 	// Set initial validation state
+	if ( initSetup.value === 0 ) {
```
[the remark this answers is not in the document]
gbirke: Yes, good catch

src/Presentation/DonationMembershipApplicationAdapter.php
gbirke: I left them out because I only want to override client-side values from the server side that I can be sure of. In the very unlikely case that the client changes its defaults, I don't have to change the server-side defaults.

src/Presentation/DonationMembershipApplicationAdapter.php
JeroenDeDauw: Mhh... sure. No objection from my side. Just a bit triggered by lack of cohesion and having an "all the things" thing, even though no real harm seems to be done.

> I don't know what you mean by "putting this in a static method in the route". Do you mean converting the route function into a Handler class, that has a static function that can be tested?

Yeah. Though I was not clued in to the route being a function and not a class at present :)

JeroenDeDauw:
+1

wiese:
Rebased in #1161 - wanted to keep the original branch just in case. 
@gbirke Please do the honors.

PR #1159, opened 2017-12-22
Purged membership handling

JeroenDeDauw:
https://phabricator.wikimedia.org/T183221 & https://phabricator.wikimedia.org/T183447

Includes adding of missing tests, fixing PHP fatal on database exceptions and refactoring to use dedicated response models rather than a single unified but fractured one.


JeroenDeDauw:
Won't be able to work on this till after the ccc, so feel free to continue with it. To finish the task, the presenter needs to get a new `presentApplicationWasPurged` method and it needs to be called from the UC. The logic that decides to call this new method in the UC (the catch block) still needs to be tested.

JeroenDeDauw:
Even though the task is not done, this should be mergeable into master

JeroenDeDauw:
This PR fixes fatal PHP errors when ...

* access is denied by the authorizer
* accessing a membership application that has been purged
* the data access layer throws an exception

For the later two more work is needed to make the UX nice. Right now the user will just get a blank page with an English message. Since this is not worse then just a blank page, this can be merged and deployed IMO. Still, before the (purged) task is done, these TODOs in the presenter should be done.

phpcs.xml
```diff
@@ -150,6 +150,7 @@
 				@slowThreshold,
 				@noinspection
 			"/>
+            <property name="enableVoidTypeHint" value="false" />
```
wiese: I'm okay with most everything we decide as a team - consciously and transparently.

JeroenDeDauw: I thought you said I could remove this if it bugged me. That was some time ago though, so perhaps my memory on it is foobar

JeroenDeDauw:
> Is this the end to the pattern of emitting response objects from use cases?

Just in cases where the classical Clean Architecture approach (which is what I did here) works better: when the response model would otherwise be very fractured. And even in those cases we can still have one, or multiple, response model DTOs for a UC. (Handed to the presenter methods, not returned from the UC.)

Are we going to have only one presenter for different scenarios (and hard-wire that fact into the use cases)?

The UC binds to the same amount of information as before, now just in the form of a n interface rather than a DTO. The new approach is more explicit, though this is a good thing.

JeroenDeDauw:
> what is temporary and what will remain.

The only temporary stuff is the lack of decent presentation for the cases that have a TODO.

wiese:
> The only temporary stuff is the lack of decent presentation for the cases that have a TODO.

I look forward to a solution in which the _one_ presenter is not instantiated with multiple templates for all possible scenarios all the time only to go unused in the majority of the cases.

phpcs.xml
wiese: @KaiNissen @gbirke Void return type hints are now not enforced any more.

PR #1160, opened 2017-12-28
JS CS for cat17

wiese:
Applied JS CS rules from 10h16.

All of the changes should be cosmetic, without change to prod functionality.

`form_validation.js` is peculiar in that it contains instructions for when "we use jQuery >= 3.0", which is the case now. In the light of it it's tempting to go down the rabbit hole AMD / explicit jQuery dependency / ES / ... but I think this warrants a ticket, is more than a side project, because the outcome definitely needs QA.


wiese:
Will rebase after #1161 and we should be clean.

wiese:
Rebased

PR #1163, opened 2017-12-29
add context specific payment interval notices

KaiNissen:
This uses the context specific payment interval notices introduced in wmde/fundraising-frontend-content#89.

See also: [phab:T183463](https://phabricator.wikimedia.org/T183463)

### DIFF
--- a/skins/cat17/templates/partials/payment_intervals.html.twig
+++ b/skins/cat17/templates/partials/payment_intervals.html.twig
@@ -12,7 +12,7 @@
 } %}
 {% spaceless %}
 {% for interval in intervals %}
-	<div class="wrap-field" data-info-text="{$ ( 'payment_interval_' ~ interval ~ '_info' ) | trans | e( 'html_attr' ) $}">
+	<div class="wrap-field" data-info-text="{$ ( context ~ '_payment_interval_' ~ interval ~ '_info' ) | trans | e( 'html_attr' ) $}">
 		<div class="wrap-input">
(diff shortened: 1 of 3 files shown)


skins/cat17/templates/partials/payment_intervals.html.twig
```diff
@@ -21,7 +21,7 @@
 		</div>
 
 		<p class="info-text" data-info="{$ PAYMENT_INTERVAL_OPTIONS[interval].id $}">
-			{$ ( 'payment_interval_' ~ interval ~ '_info' ) | trans $}
+			{$ ( context ~ '_payment_interval_' ~ interval ~ '_info' ) | trans $}
```
wiese: We don't follow a rule set (don't know a formal one), but would you mind adding the context parameter to the documentation in the head of `payment_intervals.html.twig` and maybe two examples to prevent us from making mistakes / going mad looking for those keys (from here to content repo or the other way around), now that they are so very dynamic...

wiese:
The translation is actually used twice (not ideal but a fact, reason being identical structure but additional HTML in `payment_types.html.twig`), the second time is in the `data-info-text` attribute on the `.wrap-field` element - this probably needs to change, too (used for the summary at the bottom of the page).

KaiNissen:
Thanks, good catch!

skins/cat17/templates/partials/payment_intervals.html.twig
```diff
@@ -1,6 +1,7 @@
-{# paramters:
+{# parameters:
 	intervals List of interval types to actually show
 	fieldName Name of the form fields used to submit the selected interval
+	context The context by which this partial was included (can be either `membership` or `donatioon`)
```
gbirke: small typo

PR #1164, opened 2017-12-29
T183452 ibans containing utf-8 delimiters

wiese:
Depends on https://github.com/wmde/fundraising-donations/pull/35


wiese:
Rebased

PR #1166, opened 2018-01-02
integrate latest donations + fix

wiese:
Apparently https://github.com/wmde/FundraisingFrontend/pull/1140 failed to update to the changes provided by https://github.com/wmde/fundraising-donations/pull/33, i.e. never really integrated them - basically what I meant in [this comment](https://github.com/wmde/FundraisingFrontend/pull/1140#discussion_r157697594); but back then I did not understand why the test would pass...

Fix is in ad36b4e - the fact that the bug was introduced seems to prove the point that the guard notation ([originally here](https://github.com/wmde/fundraising-donations/commit/37df0f35adf8042b7b662b1d944bdf4086658aa5#diff-8a070cff273db3663316ce1fe0a4a2a3L46)) is easier to grasp.
  
This
* constitutes a [prod bug](https://phabricator.wikimedia.org/T184059#3871573)
* blocks #1164
* blocks #1167

### DIFF
--- a/app/RouteHandlers/PayPalNotificationHandler.php
+++ b/app/RouteHandlers/PayPalNotificationHandler.php
@@ -63,10 +63,14 @@ class PayPalNotificationHandler {
 	}
 
 	private function requestIsForPaymentCompletion( ParameterBag $post ): bool {
-		return $this->isSuccessfulPaymentNotification( $post ) || (
-				$this->isForRecurringPayment( $post ) &&
-				!$this->isRecurringPaymentCompletion( $post )
-			);
+		if ( !$this->isSuccessfulPaymentNotification( $post ) ) {
+			return false;
+		}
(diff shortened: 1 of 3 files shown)


app/RouteHandlers/PayPalNotificationHandler.php
```diff
@@ -63,10 +63,14 @@ private function getValueFromCustomVars( string $customVars, string $key ): stri
...
-				!$this->isRecurringPaymentCompletion( $post )
-			);
+		if ( !$this->isSuccessfulPaymentNotification( $post ) ) {
+			return false;
+		}
+		if ( $this->isForRecurringPayment( $post ) && !$this->isRecurringPaymentCompletion( $post ) ) {
+			return false;
+		}
+
+		return true;
```
JeroenDeDauw: \o/ much more readable. The amount of negations (including in the calling predicate) is still rather mindfuck tho

PR #1170, opened 2018-01-04
T183130 CC payment notification w/o template

wiese:


src/Presentation/Presenters/CreditCardNotificationPresenter.php
```diff
@@ -5,27 +5,52 @@
...
+		return $this->render( $result );
 	}
 
+	/**
+	 * Response format expected by 3rd party
+	 *
+	 * @see https://www.micropayment.de/help/documentation/
+	 */
+	private function render( array $result ): string {
+		array_walk( $result, function( & $value, $key ) {
```
gbirke: From a performance standpoint, `array_walk` with is in-place modification is probably faster than the more functional, side-effect free version:

    $values = array_map( 
          function( $value, $key) { return $key . self::VALUE_ASSIGNMENT . $value; },
          $result,
          array_keys( $result )
    );

Feel free to leave the code as it is, it's just me getting twitchy when seeing the `&` modifier in function parameters.

tests/EdgeToEdge/Routes/CreditCardPaymentNotificationRouteTest.php
```diff
@@ -19,31 +20,32 @@
...
-				'GET',
-				'/handle-creditcard-payment-notification',
+				Request::METHOD_GET,
+				self::PATH,
 				[]
 			);
 
 			$this->assertSame( 200, $client->getResponse()->getStatusCode() );
-			$this->assertContains( 'status=error', $client->getResponse()->getContent() );
+			$this->assertContains( "status=error\n", $client->getResponse()->getContent() );
```
gbirke: Good catch!

src/Presentation/Presenters/CreditCardNotificationPresenter.php
wiese: I would have preferred a more array_map-like solution but the extra call/passing of the keys made it not worthwhile.

JeroenDeDauw: image

JeroenDeDauw: After looking at some stuff I realized array_map can take multiple arrays, so it is not that much hassle. I gave that approach a go in https://github.com/wmde/FundraisingFrontend/pull/1176. Initial result was rather verbose till I realized the reduce could be killed.

PR #1171, opened 2018-01-05
T184187 use same translation key in both skins

wiese:
Makes key _donation-info-opt-in_ obsolete, will be removed later to ensure
there is no miss in prod.


JeroenDeDauw:
+1. Would need to double check the `raw` bit for +2

wiese:
`donation_sendinfo_label` contains a `<br>` (not that I like it), so `raw` it is.

PR #1172, opened 2018-01-05
T184183 Anonymous donor summary

wiese:


skins/cat17/src/app/lib/view_handler/section_info.js
```diff
@@ -216,7 +216,7 @@ var objectAssign = require( 'object-assign' ),
 		countryNames: null,
 		update: function ( addressType, salutation, title, firstName, lastName, companyName, street, postcode, city, country, email, aggregateValidity ) {
 			var wrapperTag = '<span>',
-				longtext
+				newLongText
```
JeroenDeDauw: Took me about 10 seconds to figure out this was a variable declaration and not some text. /me grumbles about how evil this evil JS var declaration of evilness is.

skins/cat17/src/app/lib/view_handler/section_info.js
```diff
@@ -227,7 +227,11 @@ var objectAssign = require( 'object-assign' ),
 				this.setSectionStatus( SECTION_STATUS.disabled );
 			}
 
-			this.setIcon( this.getValueIcon( addressType ) );
+			if ( aggregateValidity.dataEntered ) {
+				this.setIcon( this.getValueIcon( addressType ) );
+			} else {
+				this.setIcon( undefined );
+			}
```
JeroenDeDauw: This seemed like something that should go into its own method (to not mix levels of abstraction), so I checked the whole `DonorTypeSectionInfo` class(?). Hello one big method with over9000 parameters :( Not obvious to me how to deal with the long parameter list, though a lot of things could be extracted from the big method. Perhaps something to do after this PR is merged. Idk how to best do such extraction in this type of JS, so am curious to see.

skins/cat17/src/app/tests/view_handlers/test_section_info.js
```diff
@@ -422,3 +423,146 @@ test( 'Opened longtext are shut', function ( t ) {
 
 	t.end();
 } );
+
+test( 'Donor type info without entered data indicated correctly', function ( t ) {
```
JeroenDeDauw: When looking at `DonorTypeSectionInfo` I was wondering why there are multiple classes(?) in a single file. Now when looking at this test I am wondering this once again.

In case of the production code it makes finding a specific class a bit harder, though it can make working on those related classes (especially during creation) easier and perhaps has other benefits. Based on the info I have I would still go for one class per file.

In case of the tests there are multiple functions that test a specific production class in the top level scope of the file, together with functions that test other classes, a downside absent for the production code here. This issue is compounded when extracting code from the test functions (either for sharing or single responsibility and single level of abstraction).

skins/cat17/src/app/tests/view_handlers/test_section_info.js
```diff
@@ -422,3 +423,146 @@ test( 'Opened longtext are shut', function ( t ) {
...
+
+	icon.data.withArgs( 'display-error' ).returns( true );
+	text.data.withArgs( 'empty-text' ).returns( 'nothing entered so far' );
+
+	handler.update( 'person', '', '', '', '', '', '', '', '', 'DE', '', { dataEntered: false, isValid: null } );
+
+	t.ok( container.addClass.withArgs( 'disabled' ).calledOnce, 'no data entered reflected in style' );
+	t.ok( icon.addClass.withArgs( 'icon-error' ).calledOnce, 'icon set per address type' );
+	t.ok( text.text.withArgs( 'nothing entered so far' ).calledOnce, 'fallback address type text is set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].toString(), '<span>', 'long text filled with custom mark-up' );
```
JeroenDeDauw: Is this one logical assertion or several? I tried verifying it is a single one but was unable to do this without investigating the context a bunch more.

If it is not a single one, meaning this test is a `"when scenario x, scenario x results should happen"` kind of test, rather than a set of `"when scenario x, then A should happen", "when scenario x, then B should happen", "when scenario x, then C should happen"`, then I suggest creating those dedicated tests instead. Which would make extracting setup code even more important.

If it is a single one, then putting it in a dedicated function, to make it clear it is a single one, and to not mix levels of abstraction, is probably still a good idea.

skins/cat17/src/app/tests/view_handlers/test_section_info.js
```diff
@@ -422,3 +423,146 @@ test( 'Opened longtext are shut', function ( t ) {
...
+
+	t.ok( container.addClass.withArgs( 'completed' ).calledOnce, 'data entered reflected in style' );
+	t.ok( icon.addClass.withArgs( 'icon-firma' ).calledOnce, 'icon set per address type' );
+	t.ok( text.text.withArgs( 'Firma' ).calledOnce, 'address type text is set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].toString(), '<span>', 'long text filled with custom mark-up' );
+	t.equals( longText.html.args[ 0 ][ 0 ].append.args[ 0 ][ 0 ].text.args[ 0 ][ 0 ].toString(), 'ACME INC', 'name set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].append.args[ 1 ][ 0 ].text.args[ 0 ][ 0 ].toString(), 'acmestr 133b', 'street set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].append.args[ 2 ][ 0 ].text.args[ 0 ][ 0 ].toString(), '12331 Wien', 'address set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].append.args[ 3 ][ 0 ].text.args[ 0 ][ 0 ].toString(), 'Österreich', 'country translated and set' );
+	t.equals( longText.html.args[ 0 ][ 0 ].append.args[ 4 ][ 0 ].text.args[ 0 ][ 0 ].toString(), 'us@acme.com', 'email set' );
```
JeroenDeDauw: There is a bunch duplication between these tests which makes me think code should be extracted.

Another thing I noticed here is the magical strings that need to be the same. Extracting these to whatever we use for constants in this version of JS would be nice.

These observations and suggestion of extraction are what prompted me into writing https://github.com/wmde/FundraisingFrontend/pull/1172/files#r160011395

JeroenDeDauw:
+1 assuming the concerns I wrote about will be taken care of or are not applicable. (no +2 since I do not understand the code well enough)

wiese:
Will no split the section info incarnations.

skins/cat17/src/app/lib/view_handler/section_info/types/donor_type.js
```diff
@@ -0,0 +1,66 @@
...
+					this.getValueText( addressType )
+			);
+		}
+
+		if ( !this.longText ) {
+			return;
+		}
+
+		newLongText = jQuery( wrapperTag );
+		if ( addressType === 'person' || addressType === 'firma' ) {
```
gbirke: I'm a big fan of either "bailing out early" ( inverting the condition, calling `this.longText.html( newLongText )`and  `setLongTextIndication`, and returning) instead of nesting `if`s too much. Alternatively, you could put the update code in a method.

PR #1176, opened 2018-01-09
Make code more functional

JeroenDeDauw:
The below was written for the first commit:

PHP, Y U NO two array arguments for array_reduce?! <o>

If it had, the code would just be

```php
return array_reduce(
    function( string $accumulator, string $key, $value ): string {
        return $accumulator . $key . self::VALUE_ASSIGNMENT . $value . self::ARG_SEPARATOR;
    },
    array_keys( $result ),
    array_values( $result )
);
```

### DIFF
--- a/src/Presentation/Presenters/CreditCardNotificationPresenter.php
+++ b/src/Presentation/Presenters/CreditCardNotificationPresenter.php
@@ -48,17 +48,15 @@ class CreditCardNotificationPresenter {
 	 * @see https://www.micropayment.de/help/documentation/
 	 */
 	private function render( array $result ): string {
-		return array_reduce(
+		return implode(
+			'',
 			array_map(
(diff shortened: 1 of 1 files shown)


wiese:
At this point we definitely have moved from functional or performance arguments to aesthetics altogether - but as this is in the eye of the beholder _I_ still prefer the version in #1170. -0.5.

After all, passing by reference is a perfectly good language feature, which fell out of favor but [may make a return](https://wiki.php.net/rfc/explicit_send_by_ref) if it overcomes its mayor counter-argument (at least for me) - clarity at the call-site.

JeroenDeDauw:
I did not consider performance as it does not seem relevant in this case. What I tried to do is reduce state, reduce state mutation and group concepts more.

PR #1179, opened 2018-01-10
Match TemplateBasedMailer changes in Donation component

JeroenDeDauw:
The current situation is not very clean and the needed work
makes me question is this is a good approach (perhaps a mini-lib
is better). Like this we can use latest Donation BC again, hence
the upload now. Let's do a follow up after talking about this.

Now 
image


gbirke:
Thanks for fixing this.

I'm all for splitting out the functionality as a library/service that is used by the contexts. As a basis for discussion, here are some ideas on how an alternative might look like. Please stop me at the point where you think I'm overcomplicating things 😜 

On a more abstract level, we're looking at a *Notification*. That the notification is sent out as an email and that the email is rendered by a template are unneccessary implementation details.

The simplest possible interface for an injectable service could look like this:

```php
interface Notifyable {
    function notify( array $context ): void;
}
```

However, this would imply that the class implementing `Notifyable` could not make any assumptions that `$context` contains any information on how to send the information to a recipient. If we want to restrict the notifications to email, we could do

```php
interface Notifyable {
    function notify( string $emailAddressOfRecipient, array $context ): void;
}
```

or better


```php
class Recipient {
    // constructor with email address and optional name
}
interface Notifyable {
    function notify( Recipient $recipient, array $context ): void;
}
```

And if we want to venture a bit into YAGNI territory 😱 , we could do

```php
class Recipient {}
class EmailRecipient extends Recipient {
    // constructor with email address
}
// YAGNI
class WebServiceRecipient {
    // constructor with URL endpoint
}
class SMSRecipient {
    // constructor with phone number ...
}
```

One last problem we need to solve is how the classes that implement `Notifyable` know which template to render. `$context` should not contain template names IMHO, because they are presentation layer. I'm thinking of doing an interface like this:

```php
interface Notifyable {
    function notify( string $notificationType, Recipient $recipient, array $context ): void;
}
```

The contexts could then define public (class) constants for the different notification types. My remaining question would be if we need to coordinate type constant uniqueness across contexts and if yes, how. One solution could be to have an abstract `NotificationType` class that each context needs to subclass, where each context adds a prefix to the type string.

One last thing: the calling class might actually care about the notification success. A boolean return type is not sufficient IMHO, I'm undecided between a `NotificationResult` instance or exceptions like `UnknownNotificationType` and (maybe) `RecipientTypeNotSupported`. Both have pros and cons.

JeroenDeDauw:
I'm not so sure about that approach.

If `function notify( array $context )` worked (which it indeed does not), then I'd ask "why not use an event bus". That allows to decouple fully from what is being triggered based on the event, how many things are triggered and when they are triggered.

```php
interface Notifyable {
    function notify( string $emailAddressOfRecipient, array $context ): void;
}
```

This is almost what we have now, with the difference we don't specify the second argument are template parameters. It has always bugged me we do this, since template arguments are a View Model, and the UCs have no business knowing about that, much less constructing it. (Some UCs even don't construct one and just pass along the domain entity :(). ... to be continued ...

JeroenDeDauw:
An event system allows getting rid of this View Model by the UC just emitting an "this donation happened" event. Then a listener can act on the event, pull out the donation id in the event info, and get the needed info from the persistence layer.

The obvious downside to such a system being that you now have an event system to deal with :)

Just replacing the View Model by a UC specific dedicated DTO like the Response Model or DTOs passed to presenters that are passed to UCs would improve things from an architecture PoV. Though while for new code I think we should take care here, I'm not sure if it is worthwhile changing existing code. At current "pretty bad" violations of the architecture here have not bitten us yet...

```php
class Recipient {
    // constructor with email address and optional name
}
interface Notifyable {
    function notify( Recipient $recipient, array $context ): void;
}
```
If `Recipient` wraps an email then I'd stick with email specific names in both the object and the service. Otherwise they just look like they are general while they are not.

```php
class Recipient {}
class EmailRecipient extends Recipient {
    // constructor with email address
}
// YAGNI
class WebServiceRecipient {
    // constructor with URL endpoint
}
class SMSRecipient {
    // constructor with phone number ...
}
```

This is quite fuzzy. How would the UC know what kind of recipient to use? And to know what kind of recipient a certain service supports you need to go look at the implementation. And those implementations need to do an instanceof check, or perhaps something like the constant approach you described (which as far as I can see ends up being the same thing).

PR #1191, opened 2018-01-24
Update composer.json

gbirke:
Revert change made to composer.json in #1188 to track the master of
fundraising-donations again.
Update the other dependencies


gbirke:
Duplicate of #1189

PR #1196, opened 2018-02-06
T183221 Membership confirmation after anonymization

wiese:
Depends on 
https://github.com/wmde/fundraising-memberships/pull/2
https://github.com/wmde/fundraising-frontend-content/pull/99
https://github.com/wmde/fundraising-memberships/pull/4

ApplicationAuthorizer needed to become configurable so I looked into how our next steps down BC-road could look like (in the spirit of symfony bundles).

### DIFF
--- a/tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
+++ b/tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
@@ -75,7 +75,7 @@ class ShowMembershipConfirmationRouteTest extends WebRouteTestCase {
 				]
 			);
 
-			$this->assertAccessIsDenied( $client );
+			$this->assertAccessIsDenied( 'access_denied_membership_confirmation', $client );
 		} );
 	}
 
@@ -99,7 +99,7 @@ class ShowMembershipConfirmationRouteTest extends WebRouteTestCase {
 				]
 			);
 
-			$this->assertAccessIsDenied( $client );
+			$this->assertAccessIsDenied( 'access_denied_membership_confirmation_anonymized', $client );
 		} );
 	}
 
@@ -114,7 +114,7 @@ class ShowMembershipConfirmationRouteTest extends WebRouteTestCase {
 				]
 			);
 
-			$this->assertAccessIsDenied( $client );
+			$this->assertAccessIsDenied( 'access_denied_membership_confirmation', $client );
 		} );
 	}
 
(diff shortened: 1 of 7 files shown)


tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
```diff
@@ -38,7 +38,7 @@ private function newStoredMembershipApplication( FunFunFactory $factory ): Appli
 
 	public function testWhenDonationTimestampCookieIsSet_itIsNotOverwritten(): void {
 		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ): void {
-			$donation = $this->newStoredMembershipApplication( $factory );
+			$membershipApplication = $this->newStoredMembershipApplication( $factory );
```
gbirke: The test method name also contains `Donation`.

tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
```diff
@@ -101,9 +101,28 @@ public function testCallOnAnonymizedRecord_deniedPageIsShown(): void {
...
 	}
 
+	public function testCallOnUnknownApplicationId_deniedPageIsShown(): void {
+		$this->createEnvironment( [], function ( Client $client, FunFunFactory $factory ): void {
+			$client->request(
+				Request::METHOD_GET,
+				self::PATH,
+				[
+					'id' => self::UNKNOWN_APPLICATION_ID,
+					'accessToken' => self::WRONG_ACCESS_TOKEN
```
gbirke: Shouldn't this be `CORRECT_ACCESS_TOKEN` to avoid the impression of testing the wrong thing/two things at once? Not that it matters in the implementation because the key === id+access token ...

tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
wiese: Ohh. The actual scandal is in the method body...

tests/EdgeToEdge/Routes/ShowMembershipConfirmationRouteTest.php
```diff
@@ -36,44 +39,124 @@ private function newStoredMembershipApplication( FunFunFactory $factory ): Appli
...
+		$client = $this->createClient( [], function ( FunFunFactory $factory ) use ( $membershipApplication ): void {
+
+			$factory->setMembershipTokenGenerator( new FixedMembershipTokenGenerator(
+				self::CORRECT_ACCESS_TOKEN
+			) );
+
+			$applicationRepository = $this->getMockBuilder( DoctrineApplicationRepository::class )
+				->setConstructorArgs( [$factory->getEntityManager()] )
+				->setMethods( ['getDoctrineApplicationById'] )
+				->getMock();
```
JeroenDeDauw: What does this do? Does this mean only this method is mocked, so the other real production code methods are still there? If so this seems sane enough.

Beware that there might be a test double that does something similar. Perhaps a ThrowingApplicationRepository or an InMemoryApplicationRepository that has a throwOnRead method or some such. The approach used here has the drawback that you refer to a method name with a string (breaks refactoring and navigation such as find usages). It also makes the test bind to ORMException which on first glance looks to be an implementation detail of a specific implementation of the service. ... Yeah having thought about it a bit this coupling is definitely unfortunate.

PR #1203, opened 2018-03-20
Add composer fix command

aniansson:


composer.json
```diff
@@ -139,6 +139,9 @@
 			"@validate-app-config",
 			"@stan"
 		],
+		"fix": [
+			"vendor/bin/phpcbf"
```
JeroenDeDauw: Do arguments to this get forewarded? So if you do `composer fix --stuff such`, it will become `vendor/bin/phpcbf --stuff such`? IIRC phpcbf needs you to specify stuff, so if not forewarded, then I don't understand how this is useful.

gbirke: Yes, arguments get forwarded. 

When `phpcbf` is called without parameters, it will use the `phpcs.xml` configuration to look up which directories to look into for fixes. So if you have the 3 seconds to spare, you can "fix" the whole project. 

I see this as a convenience function because `phpcbf` is such a weird name, I always have to look it up.

JeroenDeDauw: Thanks for explaining!

That's really cool then :)

PR #1205, opened 2018-03-28
T188637 Display custom amount from deep-link in UI

wiese:
https://phabricator.wikimedia.org/T188637


app/config/config.dist.json
```diff
@@ -180,6 +180,9 @@
 			"membership-enabled": false
 		}
 	},
+	"preset-amounts": {
+		"donation": [ 500, 1500, 2500, 5000, 7500, 10000, 25000, 30000 ]
```
wiese: We still "have to" use this in the templates to remove the duplication.
* https://github.com/wmde/FundraisingFrontend/blob/master/skins/cat17/templates/Donation_Form.html.twig#L51
* https://github.com/wmde/FundraisingFrontend/blob/master/skins/10h16/templates/Donation_Form_Payment.html.twig#L42
* https://github.com/wmde/FundraisingFrontend/blob/master/skins/10h16/templates/Donation_Form_Personal_Data.html.twig#L34

Of course, it would be _again_ a great time to hear that 10h16 does not need to be supported anymore...

src/Validation/IsCustomAmountValidator.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Created by IntelliJ IDEA.
```
wiese: Remove, please.

skins/cat17/templates/partials/payment_preset_amounts.html.twig
```diff
@@ -3,9 +3,9 @@
 	<div class="wrap-radio">
 		{# Amount value needs to be multiplied with 100 because the JavaScript uses integer cent amounts #}
```
wiese: Comment does not reflect the reality anymore.

skins/cat17/templates/partials/payment_preset_amounts.html.twig
```diff
@@ -3,9 +3,9 @@
...
 		{# TODO use proper money objects for predefined amounts in templates #}
-		<input type="radio" id="amount{$ loop.index $}" value="{$ amount * 100 $}">
+		<input type="radio" id="amount{$ loop.index $}" value="{$ amount $}">
 		<label for="amount{$ loop.index $}">
-			<span>{$ amount $} €</span>
+			<span>{$ amount / 100 $} €</span>
```
wiese: We really should not be doing that, no devision in particular.
The cleanest way, in my eyes, would be to pass the amounts into the view as objects (`Euro`), using one of its member functions + a (i18n-aware) function to format it for the view.
Poor man's version for the moment: `{$ amount.getEuroFloat() | number_format( 0 ) $}`
In the future we could build our own filter to format monies.

https://github.com/wmde/Euro/blob/master/src/Euro.php#L115
https://twig.symfony.com/doc/2.x/filters/number_format.html

src/Factories/FunFunFactory.php
```diff
@@ -1562,6 +1568,10 @@ public function getPaymentTypesSettings(): PaymentTypesSettings {
 		return $this->pimple['payment-types-settings'];
 	}
 
+	public function getPresetAmountsSettings(): array {
+		return $this->config['preset-amounts']['donation'];
```
wiese: If we need to convert the amounts to `Euro` instances this would be a good spot - with all the consequences e.g. in `IsCustomAmountValidator`, but we'll never have to think about it again. It caught up on us quickly that we hoped we would get away with a native data type...

skins/10h16/templates/payment_preset_amounts.html.twig
```diff
@@ -0,0 +1,8 @@
+{% for amount in presetAmounts if amount.getEuroCents() != 30000 %}
```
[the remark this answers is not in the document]
wiese: Is share your view, we discussed that briefly, but it could be it feels unintuitive to the uninitiated as it involves a certain amount of "magic".

PR #1207, opened 2018-04-04
Followup custom amounts

gbirke:


skins/cat17/templates/partials/payment_preset_amounts.html.twig
```diff
@@ -2,9 +2,9 @@
 {% for amount in presetAmounts %}
 	<div class="wrap-radio">
 		{# TODO use proper money objects for predefined amounts in templates #}
```
wiese: Sorry I only got to see this now, but this comment is now obsolete (which is not your fault!).

PR #1208, opened 2018-04-06
Add preset amount 300 to donation page in 10h16 skin

tzhelyazkova:
Feature: T191607

Elements from the donation page of 10h16 were moved around after 300 was added
in order to make the template better looking


gbirke:
I don't like those merge commits, did you force-push?

PR #1209, opened 2018-04-06
Add invalid state for the "custom amount" input field

tzhelyazkova:
Feature: T182290

Make the background of the custom amount field red on invalid input

### DIFF
--- a/skins/cat17/src/sass/components/_forms.scss
+++ b/skins/cat17/src/sass/components/_forms.scss
@@ -183,6 +183,9 @@ label, input#amount-typed, .wrap-amount-typed:after{
 		.error-text {
 			display: block;
 		}
+		input#amount-typed {
+			background-color: red;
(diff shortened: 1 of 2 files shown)


skins/cat17/src/sass/components/_forms.scss
```diff
@@ -183,6 +183,9 @@ label, input#amount-typed, .wrap-amount-typed:after{
 		.error-text {
 			display: block;
 		}
+		#amount-typed {
```
gbirke: We should avoid using ids for CSS.
See 

- https://stackoverflow.com/questions/8084555/why-selecting-by-id-is-not-recommended-in-css
- http://oli.jp/2011/ids/

gbirke:
CSS issues will be addressed in https://phabricator.wikimedia.org/T191992

PR #1213, opened 2018-04-18
Update invalid state css rules of donation page input box

tzhelyazkova:
Feature: T182290

The donation amount input box should keep the style of invalid state fields on the donation page.

p.s. Some indentation (mixing spaces and tabs) was fixed along the way.

### DIFF
--- a/skins/cat17/src/sass/common/_variables.scss
+++ b/skins/cat17/src/sass/common/_variables.scss
@@ -19,6 +19,8 @@ $brand-primary-membership-dark: #4E9242;
 $brand-footer: #cfdbe2;
 $gray-light: #ddd;
 $brand-error: #812923;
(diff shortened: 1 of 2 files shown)


skins/cat17/src/sass/common/_variables.scss
```diff
@@ -19,6 +19,7 @@ $brand-primary-membership-dark: #4E9242;
 $brand-footer: #cfdbe2;
 $gray-light: #ddd;
 $brand-error: #812923;
+$brand-error-light: #e6d4d3;
 $brand-error-rgba: rgba(129, 41, 35, 0.2);
```
gbirke: I think we should take the opportunity and rename `$brand-error-rgba` to something more meaningful like `$brand-error-dark`.

PR #1214, opened 2018-04-19
PaymentContext remnants cleanup

JeroenDeDauw:
https://phabricator.wikimedia.org/T188394


contexts/PaymentContext/src/Domain/KontoCheckBankDataGenerator.php
```diff
@@ -58,7 +63,7 @@ public function getBankDataFromAccountData( string $account, string $bankCode ):
 	 * @throws \InvalidArgumentException
 	 */
 	public function getBankDataFromIban( Iban $iban ): BankData {
-		if ( !$this->validateIban( $iban ) ) {
+		if ( $this->ibanValidator->validate( $iban )->hasViolations() ) {
 			throw new \InvalidArgumentException( 'Provided IBAN should be valid' );
```
gbirke: Since we're now not only checking for validity but also if the IBAN is on the block list, I think the exception text should reflect this. Maybe use the violation text as the Exception message?

JeroenDeDauw: Yeah. I was also thinking the message was kinda basic, but did not want to touch that in this PR

PR #1216, opened 2018-04-19
Add planning document for Payment refactoring

gbirke:
Planning document for https://phabricator.wikimedia.org/T192323


timEulitz:
I am still wondering if we should merge this or alternatively make this a ticket in Phabricator - though an argument against this would be that it is too technical, I suppose?.

In either case, I think a Phabricator ticket (even if it is of non-technical nature and only referencing this document) would be good as we would otherwise easily forget about this.

gbirke:
Yes, adding a phab ticket referencing this doc and then merging would be best.

PR #1219, opened 2018-04-24
Use latest version of the Payment BC

JeroenDeDauw:
Removes the classes moved in https://github.com/wmde/fundraising-payments/pull/5


JeroenDeDauw:
Due to the Docker changes I found out that running the tests with MySQL takes 11 minutes on my box, while SQLite takes 18 seconds :)

JeroenDeDauw:
I had the slow behavior because I did not add the new `app/config/config.test.local.json` file, and was thus running with MySQL instead of SQLite. At first I was afraid the Docker stuff was just horribly slow for some reason. Luckily not. Anyway, no problem here. Just found it interesting SQLite is that much faster.

PR #1220, opened 2018-04-24
Clean up the README

JeroenDeDauw:
Two more issues that I did not tackle here:

* Installation instructions are probably misleading. Do you need to run all these commands on a fresh install? I suspect not, and that these are just commands you sometimes need and have nothing to do with usual installation flow.

* Not clear to me how to run composer update. I spend about 5 min and ended up running `docker run --rm --interactive --tty --volume /home/j/workspace/FundraisingFrontend:/app -w /app --volume ~/.composer:/composer --user 1000:1000 composer update --ignore-platform-reqs`. Having something about this in the readme (or at least having a thing in the makefile) is probably good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,6 @@
 
 User facing application for the [Wikimedia Deutschland](https://wikimedia.de) fundraising.
 
-The easiest way to get a working installation of the application is to use [docker-compose](https://docs.docker.com/compose/).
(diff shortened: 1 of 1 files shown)


gbirke:
Thank you for cleaning up, this makes the setup much easier to understand.

While reading the README, I saw some other outdated commands and things that can be made easier (e.g. collecting all the setup commands into one `make` target). I'll do a followup.

PR #1225, opened 2018-05-03
Add preset amount config for membership part

tzhelyazkova:
Feature: T192276

The membership preset amounts config should be and work the same as the one for donations.


app/config/schema.json
```diff
@@ -870,6 +870,15 @@
...
           "required": [ "donations" ]
+        },
+        "membership": {
+          "type": "array",
+          "title": "List of possible membership amounts in cents",
+          "description": "These amounts will be selectable by the user. Any amount that's not in this list is considered a 'custom amount'.",
+          "items": {
+            "type": "integer"
+          },
+          "required": [ "membership" ]
```
gbirke: I just realized that the `required` property check is broken - it needs to be defined for "preset-amounts", not inside each property. It needs to read `"required":  [ "donations", "membership" ]`. Could you correct that in this PR?

PR #1226, opened 2018-05-04
Make BIC output optional in SEPA mandate

gbirke:
Only show BIC in cat17 SEPA mandate if there actually is one.

This is a followup for
https://phabricator.wikimedia.org/T181151
https://phabricator.wikimedia.org/T192921


JeroenDeDauw:
+1

PR #1227, opened 2018-05-04
Improve default development configuration

gbirke:
Reduce the minimum waiting period between donations and membership
applications to 1 second. This allows for "unlimited" donations and
membership applications without having to clear the cookie in between.

### DIFF
--- a/build/app/config.prod.json
+++ b/build/app/config.prod.json
@@ -44,6 +44,8 @@
       "name": "Webmaster"
     }
   },
+  "donation-timeframe-limit": "PT1s",
+  "membership-application-timeframe-limit": "PT1s",
   "purging-secret": "purgeme",
   "paypal-donation": {
     "base-url": "https://www.sandbox.paypal.com/cgi-bin/webscr?",


JeroenDeDauw:
This is development config? The file is named `config.prod.json`...

gbirke:
I'd prefer to keep the test and "prod" configurations separated, otherwise you might accidentally trigger external services when runnnig E2E and you forget to override/disable them in test. 

Dropping the `.prod` suffix would be ok with me.

What do you think about the configuration naming scheme, @aniansson , @timEulitz , @tzhelyazkova ?

tzhelyazkova:
I prefer `config.local.json` because it implies you're not gonna break anything "important". The name `config.json` sounds as "scary" as `config.prod.json` to me.

JeroenDeDauw:
> I'd prefer to keep the test and "prod" configurations separated, otherwise you might accidentally trigger external services when runnnig E2E and you forget to override/disable them in test.

Oh yeah, good point :)

PR #1230, opened 2018-05-14
Rename factory methods in DisplayPageRouteTest

gbirke:
Factory methods that create a new instance should start with "new"
instead of get.


JeroenDeDauw:
@timEulitz FYI: You forgetti'd to click the delete button

timEulitz:
@JeroenDeDauw Woopsie, I'll keep that in mind next time!

PR #1235, opened 2018-05-24
[WIP] T181615 Fix keyboard navigation caused by overloading the native UI elements with jcf

tzhelyazkova:
You can navigate through the preset amount buttons using the arrow keys.
You can navigate through radio buttons using arrow keys.
You can navigate between different groups of components on the page using tab key.

TODO: checkboxes should also receive the focus somehow..

Note: This works best on Chrome, not horrible on Firefox, complete mess on Opera.
Any help is highly appreciated :pray: 

Bug: T181615

### DIFF
--- a/skins/cat17/src/app/lib/form_components.js
+++ b/skins/cat17/src/app/lib/form_components.js
@@ -52,6 +52,12 @@ var objectAssign = require( 'object-assign' ),
 		onChange: null,
 		render: function ( formContent ) {
 			this.element.val( [ formContent[ this.contentName ] ] ); // Needs to be an array
+			this.element.on( 'focus', function () {
+				jQuery( this ).addClass( 'focused' );
+			} );
+			this.element.on( 'focusout', function () {
+				jQuery( '.focused' ).removeClass( 'focused' );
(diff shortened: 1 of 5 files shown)


wiese:
There is an [explicit test](https://github.com/wmde/FundraisingFrontend/blob/keyboard-navigation-by-jcf/skins/cat17/src/app/tests/test_scrolling.js#L355) trying to enforce the opposite - easy fix _if_ the change is really intended.

I am afraid I do not understand how to exactly the requirements in the ticket map to the code change however. Try inspecting the element at the line which you removed - is this the place in application lifecycle you want to modify? Let's discuss in person if it helps.

tzhelyazkova:
@wiese I forgot to mark this as [WIP], hence the confusion.
The line in this PR removes the focus of the text which appears upon click on a radio button. The intent is to fix the scrolling of the page when using arrow keys. It was discussed with Kai.
The problem which persists, and currently I have no idea how to solve, is the navigation through preset amounts. 
I'd highly appreciate it if we can discuss this in person, perhaps in the afternoon?

PR #1239, opened 2018-05-31
Use Subscription bounded context

gbirke:
Use the repository for the Subscription bounded context.
Delete the context from this repo.


gbirke:
Waiting for the bounded context repo to be ready

composer.json
```diff
@@ -18,6 +18,7 @@
 		"wmde/fundraising-donations": "@dev",
 		"wmde/fundraising-memberships": "@dev",
 		"wmde/fundraising-payments": "@dev",
+		"wmde/fundraising-subscriptions": "fix-validations-dev",
```
gbirke: Must be changed to `@dev` when https://github.com/wmde/fundraising-subscriptions/pull/3 is merged.

timEulitz:
I don't see any issues with the contents of this PR, however, the README file should be updated so that it mentions the new application context, specifically in the "Project structure" section.

README.md
```diff
@@ -207,6 +208,7 @@ We moved the Bounded Contexts into their own repositories:
 	* `UseCases/`: one directory per use case
 	* `DataAccess/`: implementations of services that binds to database, network, etc
 	* `Infrastructure/`: implementations of services binding to cross cutting concerns, ie logging
+* `contexts/$ContextName`: Concrete implementations of interfaces from the context.
```
gbirke: @JeroenDeDauw What do you think about moving the few remaining implementations out of `contexts` and into `src`? Then we also could get rid of [that confusing line in composer.json where we mess with the PSR-4 loading for [contexts.](https://github.com/wmde/FundraisingFrontend/pull/1239/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R118)

Regardless of the outcome of this question, this should be done in a followup PR, this one can be merged.

PR #1241, opened 2018-06-01
Implement AJAX-based Matomo opt-out

timEulitz:
https://phabricator.wikimedia.org/T195571

Also: 

https://github.com/wmde/fundraising-frontend-content/pull/109


timEulitz:
> On cat17 the box really stands out, both in terms of margin (border/text not aligned with rest of text) and in terms of font size. Is this intentional and approved by UX?

Yeah, Jan told me to make the data protection box stand out, which is why it has negative side margins. I will add this to the ticket specs.

PR #1246, opened 2018-06-05
Add CI as default make target

gbirke:
When typing "make" without any parameters, it will run the first target,
"install-js".

A better target is running the CI.


JeroenDeDauw:
Cool

PR #1247, opened 2018-06-05
Add dropdown with selectable general topics to contact form

aniansson:
Bug: https://phabricator.wikimedia.org/T194285

### DIFF
--- a/skins/10h16/templates/contact_form.html.twig
+++ b/skins/10h16/templates/contact_form.html.twig
@@ -38,6 +38,21 @@
                   {$ util.form_error('email', errors) $}
                 </div>
 
+                <div class="input-container category">
+                  <label for="category">Thema</label>
+                  <select id="category" name="category" class="category">
+                    <option value="" data-behavior="placeholder">{$ 'contact_topic_title' | trans $}</option>
+                    <option {% if category == 'contact_topic_1' | trans %}selected{% endif %}>{$ 'contact_topic_1' | trans $}</option>
+                    <option {% if category == 'contact_topic_2' | trans %}selected{% endif %}>{$ 'contact_topic_2' | trans $}</option>
+                    <option {% if category == 'contact_topic_3' | trans %}selected{% endif %}>{$ 'contact_topic_3' | trans $}</option>
+                    <option {% if category == 'contact_topic_4' | trans %}selected{% endif %}>{$ 'contact_topic_4' | trans $}</option>
+                    <option {% if category == 'contact_topic_5' | trans %}selected{% endif %}>{$ 'contact_topic_5' | trans $}</option>
+                    <option {% if category == 'contact_topic_6' | trans %}selected{% endif %}>{$ 'contact_topic_6' | trans $}</option>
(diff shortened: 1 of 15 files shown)


skins/10h16/templates/contact_form.html.twig
```diff
@@ -38,6 +38,21 @@
                   {$ util.form_error('email', errors) $}
                 </div>
 
+                <div class="input-container">
+                  <label for="">Thema</label>
+                  <select id="category" name="category" class="category">
```
tzhelyazkova: Looks like the contents of the `<select>` can be represented by using a for statement.
https://twig.symfony.com/doc/2.x/tags/for.html

skins/cat17/templates/contact_form.html.twig
```diff
@@ -26,6 +26,18 @@
 							<label for="email" class="sr-only">{$ 'contact_email'|trans $}</label>
 							{$ util.form_error('email', errors) $}
 						</div>
+						<select class="no-outline personal-title" id="title" name="title"
```
tzhelyazkova: Same here, the `<select>` contents can reduced to fewer lines using for statement.
https://twig.symfony.com/doc/2.x/tags/for.html

skins/10h16/web/_styles/styles.css
```diff
@@ -278,6 +278,11 @@ button.back-button:active { background: #cccccc; }
     -moz-border-radius: 2px;
 }
 
+.input-container .ui-selectmenu-button {
+    width: 350px !important;
```
gbirke: Ok, I see why you use `!important` here - the width is set for the element and using `!important` is the only way to override the element style. But I'd prefer if we found some other way to affect the width of the element, let's have a look at `10h16/web/_js/spenden.wikimedia.de`, there is some width trickery going on there. Let's try to understand what it does, if it's still needed and if we can improve it.

as it is, this selector breaks all the other forms where a select is used (eg.g donation and membership address form).

skins/10h16/templates/contact_form.html.twig
gbirke: Hmm, using a `for n in 1..7` that will lead to errors about missing keys or missing output when the number of topics changes. But you're right, iteration would be good here. We should do a followup where we: 

1) move the categories out of `messages.json` into their own JSON file in `i18n/de_DE/data/contact_categories.json`
2) Add an array parameter for the categories to `GetInTouchHtmlPresenter`. That array should be passed as `contact_categories` on to the template in the `render` call of the presenter.
3) In `FunFunFactory::newGetInTouchHtmlPresenter` load the language-appropriate data file (see https://github.com/wmde/FundraisingFrontend/blob/master/src/Factories/FunFunFactory.php#L476-L477 for inspiration, maybe that can be extracted into a private `loadLocalizedData` method)
4) Use `for topic in contact_topics` in the templates.

@aniansson I hope my description is sufficient, would you like to try doing it?

skins/10h16/web/_styles/styles.css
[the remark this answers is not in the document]
tzhelyazkova: Aah yes, ok, I didn't notice it's 10h16, then I'm fine with using `!important` as opposed to spending too much time figuring out why the rule is not applied on chrome.

PR #1252, opened 2018-06-07
Fix subscription process

tzhelyazkova:
Bug: T177937

Add missing templates for subscription success use case. Fix the broken confirmation link sent in the confirmation mail.

To test locally:
- Go to http://localhost:8082/contact/subscribe?skin=10h16 and type whatever email address in the form.
- Go to http://localhost:8025/ (mail hog) and open the last mail.
- Copy the confirmation link, paste it in the url bar and add `?skin=cat17` at the end.  

You should see the successful confirmation page `Subscription_Confirmed.html.twig`. If you go back and click on the link from the email, you should see the error page `Warning_Page.html.twig`, because it was already confirmed.


skins/cat17/templates/Warning_Page.html.twig
```diff
@@ -0,0 +1,20 @@
+	<div class="container">
+		<div class="row">
+			<div class="content col-xs-12 col-sm-10 col-md-8">
+				<h2>{$ 'error_page_header' | trans $}</h2>
+    			{% if message is defined %}
+					<p>
+						{$ message | trans | raw $}
```
gbirke: The message is already translated in `Confirm_Subscription.twig`. Translating it again here won't change it, (because the translation is not found), but it's a potential future error. I'd remove the `| trans` here, because the one in  `Confirm_Subscription.twig` also knows that the message comes from the validation and uses the right message file.

skins/cat17/templates/Warning_Page.html.twig
```diff
@@ -0,0 +1,20 @@
...
+					</p>
+    			{% else %}
+					<p>
+						{$ 'subscription_warning' | trans $}
+					</p>
+    			{% endif %}
+				<h2>{$ 'donation_title' | trans $}</h2>
```
gbirke: The h2 looks weird and can be removed.

skins/cat17/templates/Warning_Page.html.twig
```diff
@@ -0,0 +1,20 @@
...
+					<p>
+						<a href="{$ path( '/' ) $}" target="_blank">Zurück zur Spendenseite</a>
```
gbirke: Oops, an untranslated string.

PR #1256, opened 2018-06-08
Replace hardcoded messages which already exist in content repo

tzhelyazkova:
Follow up on #1252


skins/cat17/templates/Membership_Application_Cancellation_Confirmation.html.twig
```diff
@@ -37,7 +37,7 @@
 			</div>
 			<div class="back-block col-xs-12 col-sm-5 col-sm-offset-7 col-md-4 col-md-offset-8">
 				<ul class="list-menu link-reverse link-back list-unstyled">
-					<li><a href="{$ path( '/' ) $}" class="">Zurück zur Spendenseite</a></li>
+					<li><a href="{$ path( '/' ) $}" class="">{$ 'back_to_donation' | trans $}</a></li>
 					<li><a href="{$ basepath $}/apply-for-membership">Zurück zum Mitgliedschaftsformular</a></li>
```
gbirke: You could take the opportunity and add a `back_to_membership` key as well.

PR #1257, opened 2018-06-08
Make categories an array instead of a list of options

tzhelyazkova:
Folow up on #1247

### DIFF
--- a/app/routes.php
+++ b/app/routes.php
@@ -320,7 +320,7 @@ $app->post(
 $app->get(
 	'contact/get-in-touch',
 	function() use ( $app, $ffFactory ) {
-		return $ffFactory->getLayoutTemplate( 'contact_form.html.twig' )->render( [ ] );
+		return $ffFactory->getLayoutTemplate( 'contact_form.html.twig' )->render( [ 'contact_categories' => $ffFactory->getGetInTouchCategories() ] );
 	}
 )->bind('contact');
 
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -11,6 +11,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\ORM\EntityManager;
 use FileFetcher\ErrorLoggingFileFetcher;
+use FileFetcher\FileFetchingException;
 use FileFetcher\SimpleFileFetcher;
 use GuzzleHttp\Client;
 use NumberFormatter;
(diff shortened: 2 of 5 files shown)


app/routes.php
```diff
@@ -9,6 +9,7 @@
 
 declare( strict_types = 1 );
 
+use FileFetcher\SimpleFileFetcher;
```
gbirke: This can be removed again.

src/Factories/FunFunFactory.php
```diff
@@ -598,7 +598,15 @@ public function newAddSubscriptionJsonPresenter(): AddSubscriptionJsonPresenter
...
-		return new GetInTouchHtmlPresenter( $this->getLayoutTemplate( 'contact_form.html.twig' ), $this->getTranslator() );
+		return new GetInTouchHtmlPresenter(
+			$this->getLayoutTemplate( 'contact_form.html.twig' ),
+			$this->getTranslator(),
+			$this->getGetInTouchCategories()
+		);
+	}
+
+	public function getGetInTouchCategories(): array {
+		return json_decode( ( new SimpleFileFetcher() )->fetchFile( $this->getI18nDirectory() . '/data/contact_categories.json' ), true ) ?? [];
```
gbirke: Not sure about "masking" the error here by returning an empty array - the user will not be able to submit the contact form if there are no categories, which will be frustrating for her because she will probably have filled out the form. 

I think, from a UX perspective, it might be better to check if the categories array is not empty and throw a RuntimeException if it is. The user will see a nicely formatted error page (and can report to us that our error page is broken) and when https://phabricator.wikimedia.org/T163884 is implemented, we might even get notified.

tzhelyazkova: That's a good point. I've added logic to catch the exception thrown by `SimpleFileFetcher`, and I'd rather throw the `RuntimeException` one, instead of `FileFetchingException`, because `FileFetchingException` throws a default message in English, which looks strange in the middle of an error page in German. For this same reason `RuntimeException` is thrown with no message whatsoever. 
> we might even get notified.  
>
 :pray: :pray:

src/Factories/FunFunFactory.php
```diff
@@ -598,7 +600,24 @@ public function newAddSubscriptionJsonPresenter(): AddSubscriptionJsonPresenter
...
+			$this->getTranslator(),
+			$this->getGetInTouchCategories()
+		);
+	}
+
+	public function getGetInTouchCategories(): array {
+		try {
+			$json = ( new SimpleFileFetcher() )->fetchFile( $this->getI18nDirectory() . '/data/contact_categories.json' );
+			if ( $json === "" ) {
+				throw new RuntimeException();
```
gbirke: Please add an exception description like "Could not load categories for contact form.", it will be easier to debug when looking at a log file.

PR #1258, opened 2018-06-11
Add configuration files for campaigns

gbirke:
This is an example on how to configure feature toggles and use them. The previous "campaign" feature for selecting the Confirmation page is replaced by the new A/B testing feature.

This is for https://phabricator.wikimedia.org/T196337


gbirke:
Hey, @aniansson, @timEulitz , @tzhelyazkova, @KaiNissen: Could you comment on this format? It does not do anything yet, it's only about having an "example" file to work with. 

Things I'd like to have feedback about:

- `end` being optional, allowing for "infinite" campaigns. Or, more accurately, about campaigns whose end date we don't know yet.
- Not making time part mandatory for start and end time and using defaults.
- `active` being an additional necessary toggle (defaulting to `true`). I've copied that from [TemplateTestCampaign.php](https://github.com/wmde/FundraisingFrontend/blob/master/src/Presentation/TemplateTestCampaign.php). I think it's mostly needed during development, but it could have different meanings:
  - Additional flag for date range (Campaign is only active when `active === true` **AND** in date range). Usage: quickly deactivate campaigns during their run. While developing, pick start date in the past. When adding a campaign as PR, make sure that start date is in the future or `active` is `false.`
  - Override for date range (Campaign is only active when `active === true` **OR** in date range). Usage: Activate campaigns quickly for development. When adding a campaign as PR, make sure that `active` is `false.`
  - Override for the date range, dependent on the environment, for dev, test and prod. (Campaign is only active when string value for `active` matches current environment **OR** in date range). Valid values are `dev`, `test` and `stage` (last one not used atm).
  - Remove it altogether and make campaign always active on non-production environments, regardless of date range.
- First group item being the default when campaign is not active. This avoids having to add another `default` setting, but more verbosity could be good.
- Other useful settings not in the file?

timEulitz:
Generally looks very good to me and it definitely looks like it is already in a workable state. I would slightly favor being as explicit as possible so that the intended campaign scope and behavior is clear from just a quick glance and without having to read any lengthy configuration introduction. 

I think the `active` value should behave as you have outlined in your first paragraph (**AND**-based). That way, campaigns could be quickly toggled off and on during development and on production if the need arises without requiring developers to fiddle around with the date ranges. For the sake of being as explicit as possible, I would make `active` a mandatory setting just so that it is always there. The `start` and `end` fields should also always be defined but they should accept a null value so that the intentions are clear.

I would also be in favor of introducing a `default_group` setting which will be used for fallback if the current date is not within the active date range of the campaign or if the active flag is turned off. The idea of having the first group as a fallback is not bad, but I again, I think a more explicit approach is less error-prone and forces users to actively ask "what is the default / fallback behavior of the campaign".

A sample configuration could look like this:
```
example:
  description: Example campaign that was never run
  start: 2018-12-31
  end: 
  active: true
  groups: [ ex1, ex2 ]
  default_group: ex1
```

KaiNissen:
> `end` being optional

Landing page tests tend to run quite long. Forgetting that a test is still running will do less harm than forgetting that a test is expiring before we collected enough data to come to a conclusion. I think, the "infinite" option is the way to go.

> Not making time part mandatory for start and end time and using defaults.

I like that idea. We might not need the time of day setting very often, mostly for tests that are being run one after another.

> `active` being an additional necessary toggle (defaulting to `true`)

I agree on what Tim says. But, err, can there be a default, if it's required?

> First group item being the default when campaign is not active. This avoids having to add another `default` setting, but more verbosity could be good.

Does that mean, that the default skin is being removed from the application configuration? If not, the configuration entry should be ignored if either the `active` setting or the date settings do not apply.

By the way, is the number of test groups limited to 2?

tzhelyazkova:
> end being optional, allowing for "infinite" campaigns. Or, more accurately, about campaigns whose end date we don't know yet.
>
I think we should not allow "infinite" campaigns. I agree with Kai forgetting that a campaign is running is the lesser evil of the two, but I think we can take care of not allowing either evil happen. I think it's reasonable to assume we can define an end date even if we don't have one specified yet (something like worst case scenario). Also it caters to Tim's point about making the config file as explicit as possible.
> First group item being the default when campaign is not active. This avoids having to add another default setting, but more verbosity could be good.
>
My vote here goes to verbosity. Let's have additional default setting for this.

src/Factories/FunFunFactory.php
```diff
@@ -518,6 +523,32 @@ function( string $name, array $parameters = [] ): string {
 		$container['payment-types-settings'] = function (): PaymentTypesSettings {
 			return new PaymentTypesSettings( $this->config['payment-types'] );
 		};
+
+		$container['choice_factory'] = function (): ChoiceFactory {
+			return new ChoiceFactory( $this->getFeatureToggle() );
+		};
+
+		$container['campaign_config'] = function (): array {
+			// TODO move to CampaignFactory?
```
timEulitz: I don't feel like I understand the purpose / scope of this factory well enough to comment on this, maybe elaborate on this, possibly also in person.

timEulitz:
Setting aside the `TODO`s, I don't see any issues with the code in its current state.

On a very minor note, I think we agreed to phase out `@author` doc comments, though I don't fully recall how that conversation concluded. I don't personally care too much if you prefer leaving them in but just wanted to mention it quickly.

PR #1259, opened 2018-06-11
Remove exception message from error page

tzhelyazkova:
Because the error page is in German and the exception message is in English.
Small fix in the error page of `10h16` was added as well.

Follow up on #1257


tzhelyazkova:
Abandoning this solution, because it proves to be too complicated for the purpose of just getting rid of English text in German page.

PR #1260, opened 2018-06-11
Fix docker-compose.yml

gbirke:
Remove user-specific volume mount that was accidentally introduced in #1250


timEulitz:
Sorry about that.

PR #1263, opened 2018-06-13
Prevent composer from using the GitHub API

gbirke:
We might run into API rate limits while requesting our own packages, so
we force composer to actually use the git binary to download the
packages instead of using the GitHub API.

See https://getcomposer.org/doc/05-repositories.md#vcs


timEulitz:
Fixed it real quick. @gbirke

PR #1264, opened 2018-06-14
Add support for migration scripts

timEulitz:
T196333

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,39 @@ These tasks are also performed during the [travis](.travis.yml) runs.
 All emails sent by the application can be inspected via [mailhog](https://github.com/mailhog/MailHog)
 at [http://localhost:8025/](http://localhost:8025/)
 
-## Accessing the database
+## Database
+
+### Database migrations
+
+Out of the box, the database should be in a usable state for local development. For production changes to the database,
+doctrine migration scripts may be used. Migration scripts are stored in the *migrations* directory of
+the [FundraisingStore repository](https://github.com/wmde/FundraisingStore/tree/master/migrations).
+
+The scripts from the FundraisingStore repository can be executed through FundraisingFrontend by running these commands:
+
+To execute a specific script, run the following command and add the version number of the migration script you want to use.
+As an example, executing `migrations/Version20180612000000.php` would look like this:
+
+```
+make migration-execute MIGRATION_VERSION=20180612000000
+```
+
+You can also revert a script (if implemented) through an equivalent `migration-revert` command:
+
+```
+make migration-revert MIGRATION_VERSION=20180612000000
+```
+
+Note that Doctrine creates its own `doctrine_migration_versions` table where it stores the status of individual migrations.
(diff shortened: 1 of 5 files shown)


timEulitz:
You should run the following command after you pull this on your local machine, it's only necessary for existing machines and alters the database tables:

```
make migration-execute MIGRATION_VERSION=20180612000000
```

This change can only deployed to production in maintenance mode as it requires database changes which have to be installed via the following command from the project root of the production server:

```
vendor/doctrine/migrations/bin/doctrine-migrations migrations:execute 20180612000000 --up --configuration=vendor/wmde/fundraising-store/migrations.yml
```

README.md
```diff
@@ -123,25 +123,52 @@ These tasks are also performed during the [travis](.travis.yml) runs.
...
 ### Accessing the database from a Docker image
 
-The database container of the Docker development environment is not exposed to the outside. If you want to connect to 
-the default fundraising frontend database using the MySQL command line client, you need first to find out the Docker 
-network name where the database is running in. With the command
+If you want to access the database via the terminal, the easiest way is to first start the docker containers via `docker-compose up`,
+you can then connect to the database via:
 
-    docker network ls
+    docker exec -it fundraisingfrontend_database_1 /bin/bash
```
gbirke: Unfortunately, the 1st part of the container name is dependent on the folder name of you FF checkout. Without forcing everyone to use the same folder name, this command won't work for everyone.

Also, it's purely accidental that there is a `bash` executable in the the mysql image. I think it's more consistent to directly make the mysql client command the executable command.

README.md
```diff
@@ -123,25 +123,52 @@ These tasks are also performed during the [travis](.travis.yml) runs.
...
 
-you can list all networks. There will be one network name ending in `fundraising_proxy` (the prefix is probably the 
-directory name where you checked out this repository). Copy the full network name and use it instead of the 
-placeholder `__NETWORK_NAME__` in the following command to run the MySQL command line client:  
+This will open a bash terminal in the database container, from there you can connect to MySQL via:
 
-    docker run -it --link spenden2_database_1:mysql --net __NETWORK_NAME__ --rm mysql:5.6 \
```
gbirke: As you can see here, my folder is called 'spenden2' instead of 'FundraisingFrontend'. So we probably need additional instructions on how to determine the container name.

gbirke: Please keep the network setup stuff for now & move it to the Fundraising Operatiosn Center README in a separate PR.

README.md
```diff
@@ -123,25 +123,52 @@ These tasks are also performed during the [travis](.travis.yml) runs.
...
-        sh -c 'exec mysql -h database -P 3306 -u fundraising -p"INSECURE PASSWORD" fundraising'
+    mysql -u fundraising -p"INSECURE PASSWORD" fundraising
 
 ### Accessing the database from your host machine
 
+You will be prompted for a password which you can grab from `config/config.prod.json`.
```
gbirke: Please move that sentence to the bottom of the description, as the password only becomes relevant when the container is set up.

PR #1265, opened 2018-06-15
Make label not part of selectable options in contact form

tzhelyazkova:
Bug: T194285

### DIFF
--- a/skins/10h16/templates/contact_form.html.twig
+++ b/skins/10h16/templates/contact_form.html.twig
@@ -41,7 +41,7 @@
                 <div class="input-container category">
                   <label for="category">Thema</label>
                   <select id="category" name="category" class="category">
-                    <option value="" data-behavior="placeholder">{$ 'contact_topic_title' | trans $}</option>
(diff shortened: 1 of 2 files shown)


skins/10h16/templates/contact_form.html.twig
```diff
@@ -41,7 +41,7 @@
                 <div class="input-container category">
                   <label for="category">Thema</label>
                   <select id="category" name="category" class="category">
-                    <option value="" data-behavior="placeholder">{$ 'contact_topic_title' | trans $}</option>
+                    <option value="" data-behavior="placeholder" disabled selected>{$ 'contact_topic_title' | trans $}</option>
```
gbirke: Always using `selected` (even when another category is selected, in case the form is re-displayed after unsuccessful validation), means slightly invalid HTML. AFAIK browsers will still display the last option with `selected`, but maybe you can wrap it with `{% if !category %}selected{% endif %}`

tzhelyazkova: I don't think this check is necessary. Once an option is chosen it gets the `selected` attribute assigned to it, and removed from the default one. Even on reload you still see the option which was selected.

PR #1266, opened 2018-06-18
Clean cache in docker containers

gbirke:
The docker container uses a shared volume called var-data for the `var`
directory of the application, containing cache and other files. The
`make clear` command now empties the cache files in that directory.

Added the alias `make clean` for `make clear` to avoid a common typo.


gbirke:
Tests will probably pass when #1269 is merged and this is rebased.

PR #1267, opened 2018-06-19
Select skin test group and persist selection

tzhelyazkova:
Feature: T196335

TODOs:
- [x] Fix setting cookie in `bootstrap.php` (during our manual user tests it was not overwritten by the URL parameter, even when the URL parameter did override the group selection)
- [x] remove `Campaign::getDefaultGroup`, use `Group::isDefaultGroup` instead.
- [x] Remove call order dependence: If `FunFunFactory::getSelectedGroups` is called before url and cookie parameters are set for `GroupSelector`, then we will always have the default groups (because selected groups is a "singleton array" (stored with `getSharedObject` )). Accidentally, we got this right for now, but let's make sure future changes won't affect this.
- [x] Change `StringHash` generation (both for requestor and in `CampaignFeatureBuilder`) from `$group->getName()` to `$group->getId()`. `getId` should return `<CAMPAIGN_NAME>.<GROUP_NAME>`. This ensures that group names do not need to be unique inside the campaign configuration.
- [x] Add EdgeToEdge test that checks 
  - if the cookie is set when a user visits a page
  - the cookie is changed when a user visits a page with URLParameters
- [x] Fix tests for `CampaignFeatureBuilder`: Check for nested rules that are generated for active campaigns (we had to change the sequential rules to nested rules)
- [x] Add unit test for `Group::getUrlParameter`. Maybe rename`getUrlParameter` to `getIdArray` or `getParameters` because calling `getUrlParameter` for cookie generation (an DB storage) looks weird. 
- [x] Add unit tests for `Campaign::getGroupByIndex` and `Campaign::getIndexByGroup`
- [x] `GroupSelectorTest`
  - [x] Given URL and Cookie parameters, URL overrides cookie
  - [x] Test sanitizing of URL and cookie parameters
  - [x] Test URL and cookie parameters from the constructor are also sanitized, or get rid of them in the constructor 
  - [x] Refactor. Add blank lines to follow Arrange-Act-Assert pattern. Maybe add more factory functions (for creating and setting groups).
- [x] Remove `CampaignCollection::hasUrlParameter`
- [x] More test cases for `CampaignCollectionTest`. 
- [x] Refactor changing the skin 
   - [x] edit campaign configuration: make it an inactive campaign, add local dev override `campaigns.dev.yml` and activate it for `campaigns.test.yml`.
   - [x] add new code to `ChoiceFactory` that returns a skin name based on campaign
   - [x] Replace old skin changing code in `FunFunFactory::getSkinDirectory`  with calls to the new `ChoiceFactory` method.
   - [x] Delete all classes that are no longer necessary (`SkinSettings`, `SkinServiceProvider`). Also delete their tests.
   - [x] Fix all failing Edge2Edge tests (e.g. by adding a 3rd group called "test" to the skin campaign configuration in `campaigns.test.yml` and making it default )  
   - [x] Add Edge2Egde test that checks changing of the skin. Refactoring `CampaignCookieTest::testWhenUserVisitsThePageWithUrlParams_cookieIsChanged` might be a good candidate for this.
   - [x] Remove skin configuration keys from `config.dist.json` and `schema.json`
- [x] Move classes into Infrastructure subdirectories and namespaces
- [x] Remove PHPStorm-created file comments
- [ ] Think about cached campaign configuration: 
  - Can it be disabled for tests? Its it disabled by default for tests?
  - What happens if the campaign definitions change without the cache being purged?
  - Did we accidentally put request-specific information in there? 
- [x] Replace `array_rand` in `GroupSelector` with `ArraySelectionInterface` and implement `EquallyRandomizedSelection` (with `array_rand`), `FixedSelection` (in tests\Fixtures). This would enable us to be really sure that `testGivenNoParams_groupIsRandomlySelected` selects randomly and is not using parameters
- [x] Pass CI
- [ ] Add `campaigns.dev.yml`
  - [x] Create `campaigns.dev.yml`
  - [ ] Add `campaigns.dev.yml` to `config.prod.json` for test-spenden-2.wikimedia.de at the deploy server


src/Infrastructure/BucketTesting/BucketSelectionServiceProvider.php
```diff
@@ -0,0 +1,62 @@
...
+use Silex\Api\BootableProviderInterface;
+use Silex\Application;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+
+class BucketSelectionServiceProvider  implements ServiceProviderInterface, BootableProviderInterface {
+
+	private const COOKIE_NAME = 'spenden_ttg';
+	private const COOKIE_LIFETIME_IN_SECONDS = 60 * 60 * 24 * 90; // 90 days
```
gbirke: Three months was only an assumption. See https://phabricator.wikimedia.org/T196335#4301120 for necessary changes.

src/Infrastructure/BucketTesting/BucketSelector.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace WMDE\Fundraising\Frontend\Infrastructure\BucketTesting;
+
+class BucketSelector {
+
+	private $campaigns;
+ 	private $cookie;
+ 	private $urlParameters;
+
+	public function __construct( CampaignCollection $campaigns, array $cookie = [], array $urlParameters = [] ) {
```
gbirke: Let's get rid of the constructor parameters and setter and make the url and cookie variables parameters for `selectbuckets`.

src/Infrastructure/BucketTesting/CampaignCollection.php
```diff
@@ -0,0 +1,26 @@
...
+namespace WMDE\Fundraising\Frontend\Infrastructure\BucketTesting;
+
+class CampaignCollection {
+	private $campaigns;
+
+	public function __construct( Campaign ...$campaigns ) {
+		$this->campaigns = $campaigns;
+	}
+
+	public function splitBucketsFromCampaigns( array $params ): array {
```
gbirke: We need a doc comment here that explains what happens.

tests/Unit/Infrastructure/BucketTesting/BucketSelectorTest.php
```diff
@@ -0,0 +1,83 @@
...
+				[]
+			) )->selectBuckets(),
+			$this->logicalOr(
+				$this->equalTo( [ $bucketA ] ),
+				$this->equalTo( [ $bucketB ] )
+			)
+		);
+	}
+
+	public function testGivenAssignedBucket_campaignParametersAreSet() {
```
gbirke: This should be moved into BucketTest.

tests/EdgeToEdge/CampaignCookieTest.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\EdgeToEdge;
+
```
gbirke: We need to add a `@covers` tag.

tests/Unit/Infrastructure/BucketTesting/BucketSelectorTest.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Infrastructure;
+
+use PHPUnit\Framework\TestCase;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\Bucket;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\BucketSelector;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\Campaign;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\CampaignCollection;
+
```
gbirke: We need to add a `@covers` tag.

tests/Unit/Infrastructure/BucketTesting/CampaignCollectionTest.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Infrastructure;
+
+use PHPUnit\Framework\TestCase;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\Bucket;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\Campaign;
+use WMDE\Fundraising\Frontend\Infrastructure\BucketTesting\CampaignCollection;
+
```
gbirke: We need to add a `@covers` tag.

tests/Unit/Infrastructure/BucketTesting/CampaignCollectionTest.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Infrastructure;
```
gbirke: Namespace also needs `BucketTesting`.

tests/Unit/Infrastructure/BucketTesting/CampaignBuilderTest.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Infrastructure;
```
gbirke: Namespace also needs `BucketTesting`.

tests/Unit/Infrastructure/BucketTesting/BucketSelectorTest.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Infrastructure;
```
gbirke: Namespace also needs `BucketTesting`.

src/Infrastructure/BucketTesting/CampaignCollection.php
gbirke:
```
Use urlKey => BucketIndex values to select matching buckets from campaigns. 
Also return all campaigns that were not matched.

@return [ Bucket[], Campaign[] ]
```

PR #1271, opened 2018-06-21
Add skins to a b testing refactor skins

timEulitz:


timEulitz:
@gbirke Sorry about the Dockerfile, I really need to set up a local Git ignore until we come up with a better permanent solution.

src/Factories/ChoiceFactory.php
```diff
@@ -58,4 +58,19 @@ public function getConfirmationPageTemplate( Twig_Environment $twig, array $cont
 		throw new UnknownChoiceDefinition( 'Confirmation Page Template configuration failure.' );
 	}
 
+
+	public function getSkinForUserBucket(): string {
```
gbirke: I'd call this function `getSkinTemplateDirectory` and the other one `buildSkinDirectory`. Reasoning:

* the public function should clearly reflect what's returned.
* We don't have the concept of "Users" yet
* The "Bucketing" is implied by using the `ChoiceFactory` and should not show up in its method names.

src/Factories/FunFunFactory.php
```diff
@@ -1642,7 +1632,7 @@ public function newIsCustomDonationAmountValidator(): IsCustomAmountValidator {
 	}
 
 	public function getSkinDirectory(): string {
-		return 'skins/' . $this->getSkinSettings()->getSkin() . '/templates';
+		return $this->getChoiceFactory()->getSkinForUserBucket();
```
gbirke: 👍 

I like this wrapper! This ensures that when the skin campaign is over and 10h16 finally dead, this method can be reverted to building a string.

tests/EdgeToEdge/Routes/DisplayPageRouteTest.php
```diff
@@ -19,7 +20,17 @@
 class DisplayPageRouteTest extends WebRouteTestCase {
 
 	public function testWhenPageHasCustomTemplate_customTemplateIsRendered(): void {
-		$client = $this->createClient( [ 'skin' => [ 'default' => 'cat17' ] ] );
+		$client = $this->createClient(
+			[],
+			function ( FunFunFactory $factory ): void {
```
gbirke: Would you object to making the code more DRY by creating a private method in the test like `setDefaultSkin( FunFunFactory $factory)`?

tests/Fixtures/FakeBucket.php
```diff
@@ -0,0 +1,32 @@
...
+ * @license GNU GPL v2+
+ */
+class FakeBucket {
+
+	public static function createBucket(
+		string $bucketName = 'test',
+		bool $isDefault = true,
+		$campaignName = 'test_campaign',
+		$campaignUrlKey = 'test_url_key'
+	): Bucket {
```
gbirke: The formatting looks a bit off. I don't know if it's possible with the code standards, I prefer to fit as much parameters in one line and then continue the params on the next line at the start of the bracket. 

But it's totally up to you and I fear it's not consistent in the code base ...

PR #1274, opened 2018-06-25
Improve bucket selection logic

gbirke:
For inactive or expired campaigns, always select the default bucket.

The CampaignFeatureBuilder will now always add a bucket check to all
features.
This enables us to check the feature flags in any order in the
ChoiceFactory:

- If the campaign is inactive, the CampaignFeatureBuilder has built
  rules that ignore the buckets.
- If the campaign is active but expired, the BucketSelector will have
  put the visitor in the default bucket. The date range checks for the
  non-default features will disable them. The default feature has no
  date range check, only a bucket name check.
- If the campaign is active and inside its date range, the normal bucket
  selection rule priorities ( URL > Cookie > Random) apply.

### DIFF
--- a/src/BucketTesting/BucketSelector.php
+++ b/src/BucketTesting/BucketSelector.php
@@ -4,14 +4,16 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\BucketTesting;
 
+use LogicException;
+
 class BucketSelector {
 
 	private $campaigns;
-	private $selectionStrategy;
+	private $fallbackSelectionStrategy;
 
-	public function __construct( CampaignCollection $campaigns, BucketSelectionStrategy $selectionStrategy ) {
+	public function __construct( CampaignCollection $campaigns, BucketSelectionStrategy $fallbackSelectionStrategy ) {
 		$this->campaigns = $campaigns;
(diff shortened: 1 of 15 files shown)


src/BucketTesting/BucketSelector.php
```diff
@@ -27,14 +29,28 @@ private function sanitizeParameters( array $params ): array {
...
 		return $buckets;
 	}
 
+	private function selectBucketWithStrategies( array $selectionStrategies, Campaign $campaign ): Bucket {
+		foreach( $selectionStrategies as $strategy ) {
+			if ( $bucket = $strategy->selectBucketForCampaign( $campaign ) ) {
+				return $bucket;
+			}
+		}
+		throw new InvalidArgumentException( 'Fallback bucket selection must always return a bucket.' );
```
JeroenDeDauw: Maybe https://secure.php.net/manual/en/class.outofboundsexception.php is more appropriate

gbirke: This exception can only be caused if the `fallbackSelectionStrategy` argument passed into the `BucketSelector` constructor does not return a bucket. Not because some out of bounds situation. I see that it can be a bit confusing because the argument was not passed directly to the method, but to the constructor instead. I'll throw a `LogicException` instead.

PR #1275, opened 2018-06-26
Create campaign YAML file validator

timEulitz:
https://phabricator.wikimedia.org/T197067


cli/CampaignConfigValidation/ValidateCampaignConfigCommand.php
```diff
@@ -0,0 +1,86 @@
...
+						new InputArgument( 'environment', InputArgument::OPTIONAL, '', 'prod' )
+					]
+				)
+			);
+	}
+
+	protected function execute( InputInterface $input, OutputInterface $output ): int {
+		$environment = $input->getArgument( 'environment' );
+		$factory = $this->getFactory( $environment );
+		try {
```
gbirke: You might want to move this try...catch block into a `LoggingCampaignConfigurationLoader` that implements the `CampaignConfigurationLoader` interface and is a wrapper around `CampaignConfigurationLoader`. Its `loadCampaignConfiguration` could catch exactly the exceptions `CampaignConfigurationLoader` throws and translate them into error messages (and letting other, unforeseen exceptions bubble up). That way, you can even unit test the error handling by passing in a dummy `ThrowingCampaignConfigurationLoader` in the test.

The Campaign Loader can be set in the factory before calling `$factory->getCampaignCollection();`

cli/CampaignConfigValidation/CampaignValidator.php
```diff
@@ -0,0 +1,66 @@
...
+	private function validate() {
+		if ( $this->hasValidated ) {
+			return;
+		}
+
+		/**
+		 * @var $campaign \WMDE\Fundraising\Frontend\BucketTesting\Campaign
+		 */
+		foreach ( $this->campaignCollection->getIterator() as $campaign ) {
+			foreach ( $this->getRules() as $rule ) {
```
gbirke: Could you move rules iteration into a `validateCampaign` method? that way we can keep the nesting level low.

A small performance improvement over instantiating new rules for every campaign could be moving out the error tracking from the rules and into an `ErrorLogger` that gets passed to every rule.

cli/CampaignConfigValidation/CampaignValidator.php
```diff
@@ -0,0 +1,66 @@
...
+				if ( !$rule->validate( $campaign ) ) {
+					$this->errors[] = $rule->getMessages();
```
gbirke: This creates a nested array, you'd need to call `array_merge` to have a flat stream of error messages for every rule & campaign. Or go with the `ErrorLogger` solution instead. But that's fully optional, just a suggestion.

cli/CampaignConfigValidation/CampaignValidator.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Cli\CampaignConfigValidation;
```
gbirke: From my point of view, the Validator is more of a domain object and would therefore belong to the `Frontend\BucketTesting\Validation` namespace, but let's discuss that.

cli/CampaignConfigValidation/CampaignValidator.php
```diff
@@ -0,0 +1,68 @@
...
+
+		return $this->errorLogger->getErrors();
+	}
+
+	private function validate() {
+		if ( $this->hasValidated ) {
+			return;
+		}
+
+		foreach ( $this->campaignCollection->getIterator() as $campaign ) {
```
gbirke: The `IteratorAggregate` interfaces ensures that you can iterate `CampaignCollection` without calling `getIterator`, the will be used by PHP internally.

cli/CampaignConfigValidation/CampaignValidator.php
```diff
@@ -0,0 +1,68 @@
...
+			$this->validateCampaign( $campaign );
+		}
+
+		$this->hasValidated = true;
+	}
+
+	private function validateCampaign( Campaign $campaign ): void {
+		foreach ( $this->getRules() as $rule ) {
```
gbirke: I think in this case it would be ok to use `$this->rules` directly as a performance optimization.

cli/CampaignConfigValidation/LoggingCampaignConfigurationLoader.php
```diff
@@ -0,0 +1,38 @@
...
+			return $this->actualCampaignConfigurationLoader->loadCampaignConfiguration( ...$configFiles );
+		}
+		catch ( ParseException $e ) {
+			$this->errorLogger->addError( 'Failed to parse YAML: ' . $e->getMessage() );
+		}
+		catch ( \LogicException $e ) {
+			$this->errorLogger->addError( 'Logical error in YAML: ' . $e->getMessage() );
+		}
+		catch ( \RuntimeException $e ) {
+			$this->errorLogger->addError( 'Runtime error while parsing YAML: ' . $e->getMessage() );
```
gbirke: The runtime error is thrown if one of the config files is not found.

cli/CampaignConfigValidation/LoggingCampaignConfigurationLoader.php
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	public function loadCampaignConfiguration( string ...$configFiles ): array {
+		try {
+			return $this->actualCampaignConfigurationLoader->loadCampaignConfiguration( ...$configFiles );
+		}
+		catch ( ParseException $e ) {
+			$this->errorLogger->addError( 'Failed to parse YAML: ' . $e->getMessage() );
+		}
+		catch ( \LogicException $e ) {
```
gbirke: I don't remember `LogicException` being thrown ...

cli/CampaignConfigValidation/LoggingCampaignConfigurationLoader.php
```diff
@@ -0,0 +1,38 @@
...
+			$this->errorLogger->addError( 'Logical error in YAML: ' . $e->getMessage() );
+		}
+		catch ( \RuntimeException $e ) {
```
gbirke: Two Exceptions are missing: `FileFetchingException` (if a file can't be read for some reason, e.g. permissions) and `InvalidConfigurationException` (if the configuration structure is wrong, e.g. required keys missing, additional or misspelled keys)

cli/CampaignConfigValidation/Rule/DefaultBucketRule.php
```diff
@@ -0,0 +1,26 @@
...
+ * @license GNU GPL v2+
+ */
+class DefaultBucketRule implements CampaignValidationRuleInterface {
+
+	public function validate( Campaign $campaign, ValidationErrorLogger $errorLogger ): bool {
+		try {
+			$campaign->getDefaultBucket();
+		}
+		catch ( \LogicException $e ) {
+			$errorLogger->addError( 'Must have a valid default bucket.', $campaign );
```
gbirke: you can scratch `valid` for more clarity.

cli/CampaignConfigValidation/Rule/UniqueBucketRule.php
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @license GNU GPL v2+
+ */
+class UniqueBucketRule {
+
+	public function validate( Campaign $campaign, ValidationErrorLogger $errorLogger ): bool {
+		$buckets = [];
+		$valid = true;
+		foreach ( $campaign->getBuckets() as $bucket ) {
+			if ( in_array( $bucket->getName(), $buckets ) ) {
```
gbirke: As a small performance optimization, you could use `$buckets` as a lookup table:

`isset( $buckets[$bucket->getName] )`

cli/CampaignConfigValidation/Rule/UniqueBucketRule.php
```diff
@@ -0,0 +1,28 @@
...
+				$valid = false;
+				$errorLogger->addError( 'Duplicate bucket ' . $bucket->getName(), $campaign );
+			}
+			$buckets[] = $bucket->getName();
```
gbirke: If you do the lookup table, the this needs to be `$buckets[$bucket->getName()] = true;`

cli/CampaignConfigValidation/ValidateCampaignConfigCommand.php
```diff
@@ -0,0 +1,86 @@
...
+	protected function execute( InputInterface $input, OutputInterface $output ): int {
+		$environment = $input->getArgument( 'environment' );
+		$factory = $this->getFactory( $environment );
+
+		$errorLogger = new ValidationErrorLogger();
+
+		$configLoader = new LoggingCampaignConfigurationLoader(
+			$factory->getCampaignConfigurationLoader(), $errorLogger
+		);
+		$factory->setCampaignConfigurationLoader( $configLoader );
```
gbirke: Could be written in one go, without `$configLoader`.

cli/CampaignConfigValidation/ValidateCampaignConfigCommand.php
```diff
@@ -0,0 +1,86 @@
...
+		$campaignCollection = $factory->getCampaignCollection();
+
+		if ( $campaignCollection ) {
```
gbirke: The CampaignCollection will be initialized with an empty array, but that is still a "truthy" value and the following code is run, resulting in the wrong error code if there were loading errors. Maybe addd `hasErrors` to `ErrorLogger` and check that instead?

cli/CampaignConfigValidation/ValidateCampaignConfigCommand.php
```diff
@@ -0,0 +1,86 @@
...
+
+		foreach ( $errorLogger->getErrors() as $error ) {
+			$output->writeln( $error );
+		}
+		$output->writeln( 'Campaign YAML validation failed.' );
+		return self::ERROR_RETURN_CODE;
+	}
+
+	private function getFactory( string $environment ): FunFunFactory {
+		$prodConfigPath = __DIR__ . '/../../app/config/config.' . $environment . '.json';
```
gbirke: Could you rename this to `$environmentConfigPath`?

tests/Unit/Cli/CampaignValidatorTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\Cli\CampaignConfigValidation\CampaignValidator
+ */
+class CampaignValidatorTest extends \PHPUnit\Framework\TestCase {
+
+	public function testWhenValidCampaignIsSupplied_validationPasses(): void {
+		$errorLogger = new ValidationErrorLogger();
+		$bucket = FakeBucket::createBucket();
+		$campaignCollection = new CampaignCollection( $bucket->getCampaign() );
```
gbirke: I had a look at `FakeBucket` and it creates a Campaign with one bucket. Let's discuss if that is something that's useful (and scenarios where we would use that) or we should add a validation rules for "campaign needs to have at least 2 buckets".

tests/Unit/Cli/CampaignValidatorTest.php
```diff
@@ -0,0 +1,63 @@
...
+	public function testWhenCampaignBucketsAreNotUnique_validationFails(): void {
+		$errorLogger = new ValidationErrorLogger();
+		$bucket = FakeBucket::createBucket( 'test', true );
+		$bucket->getCampaign()->addBucket( new Bucket( 'test', $bucket->getCampaign(), false ) );
+
+		$rule = new UniqueBucketRule();
+		$rule->validate( $bucket->getCampaign(), $errorLogger );
+		$this->assertEquals( ['test_campaign: Duplicate bucket test'], $errorLogger->getErrors() );
+	}
+}
```
gbirke: Not sure if the campaigns and buckets should be initialized with `new` statements to make it more clear what's actually in them.

app/config/campaigns.dev.yml
```diff
@@ -5,10 +5,6 @@
...
 
-
 skins:
-  active: false
-  buckets:
-    - "cat17"
-    - "10h16"
-  default_bucket: "cat17"
\ No newline at end of file
+  end: "2020-10-10"
```
gbirke: This comes from #1274, could you rebase this branch to master?

cli/CampaignConfigValidation/LoggingCampaignConfigurationLoader.php
timEulitz: See `\WMDE\Fundraising\Frontend\BucketTesting\Campaign::getDefaultBucket()`.

gbirke: Yes, but that's not called when loading the configuration, only during testing and when buckets are assigned. If a LogicException occurs, I think it's better to lump it with the handling of the other unexpected exceptions.

timEulitz: Ah alright, thanks for pointing that out.

cli/CampaignConfigValidation/CampaignValidator.php
timEulitz: The main purpose of the method is actually the PHPDoc comment as it signals that the rules should be concrete implementations of `CampaignValidationRuleInterface`. I can still remove it though / change this if you have a better idea.

gbirke: The PHPDoc could be put above the instance variable declaration in line 27. But this kind of micro-optimization is no blocker, no need to change it now.

timEulitz: Found another little thing which I wanted to change so I just threw this change in as well.

PR #1277, opened 2018-06-27
Reorganize components in the contact form

tzhelyazkova:
Feature: T198052

- Contact form input fields are reorganized.
- Submit button and back link are moved to the top in line with `Topics` and `Subject`.
- In mobile view (or very small screens) everything is in 1 column.
- In mobile view the submit button and back link fit the view port.

### DIFF
--- a/skins/cat17/src/sass/layouts/pages/contact.scss
+++ b/skins/cat17/src/sass/layouts/pages/contact.scss
@@ -33,13 +33,32 @@
 			}
 		}
 		.wrap-submit {
-			margin-top: 70px;
 			width: 90%;
 			display: flex;
 			justify-content: center;
+			@include bkp(sm) {
+				margin-top: 70px;
+			}
 		}
 		.list-menu {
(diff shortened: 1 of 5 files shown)


skins/cat17/src/sass/layouts/pages/contact.scss
```diff
@@ -33,13 +33,51 @@
...
+				margin-top: 70px;
+			}
 		}
 		.list-menu {
 			margin-top: 30px;
 		}
 	}
+	@include bkp(smmax) {
+		.col-sm-6 {
+			width: 100%;
```
gbirke: Ok, this works, but does't that go against the spirit of the bootstrap grid? I would have thought that the change would involve adding more column classes to the markup and let Bootstrap do its thing.

timEulitz: Thanks a lot, I really was not familiar with that!

PR #1278, opened 2018-06-29
Move some Silex routes to controller classes

gbirke:
This example shows how the anonymous functions from the routes.php files
can be moved into controller methods. This is a preparation to move to
Symfony, where all routing actions must be placed in controller methods.

This example also shows, how a RouteHandler class can be converted into
a controller class.

The new FundraisingFactoryServiceProvider enables the Silex Application
container to automatically inject the FunFunFactory into all controller
classes.

In the future, the FundraisingFactoryServiceProvider might also provide
the use case classes to the controllers, to make it more clear what the
controller is actually about and to further reduce the access surface.

### DIFF
--- a/app/FundraisingFactoryServiceProvider.php
+++ b/app/FundraisingFactoryServiceProvider.php
@@ -1,0 +1,55 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App;
+
+use Generator;
+use Pimple\Container;
+use Pimple\ServiceProviderInterface;
+use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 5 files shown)


app/FundraisingFactoryServiceProvider.php
```diff
@@ -0,0 +1,54 @@
...
+					$this->ffFactory = $ffFactory;
+				}
+
+				public function supports( Request $request, ArgumentMetadata $argument ): bool {
+					return $argument->getType() === FunFunFactory::class;
+				}
+
+				public function resolve( Request $request, ArgumentMetadata $argument ): iterable {
+					yield $this->ffFactory;
+				}
```
JeroenDeDauw: This method has a funky interface :)

JeroenDeDauw:
+1

PR #1279, opened 2018-06-29
Create code analyzer to find dead campaign code

timEulitz:
T196929

### DIFF
--- a/src/BucketTesting/Bucket.php
+++ b/src/BucketTesting/Bucket.php
@@ -22,7 +22,7 @@ class Bucket {
 	}
 
 	public function getId(): string {
-		return $this->getCampaign()->getName() . '.' . $this->getName();
+		return 'campaigns.' . $this->getCampaign()->getName() . '.' . $this->getName();
 	}
 
 	public function getCampaign(): Campaign {
--- a/src/BucketTesting/Validation/CampaignUtilizationValidator.php
+++ b/src/BucketTesting/Validation/CampaignUtilizationValidator.php
@@ -1,0 +1,118 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\BucketTesting\Validation;
+
+use WMDE\Fundraising\Frontend\BucketTesting\Campaign;
+use WMDE\Fundraising\Frontend\BucketTesting\CampaignCollection;
+
+/**
+ * @license GNU GPL v2+
+ */
+class CampaignUtilizationValidator {
+
+	private $hasValidated = false;
+	private $campaignCollection;
+	private $ignoredBuckets;
+	private $featureToggles;
+	private $errorLogger;
+
+	public function __construct(
+		CampaignCollection $campaignCollection,
+		array $ignoredBuckets,
+		array $featureToggles,
+		ValidationErrorLogger $errorLogger
+	) {
+		$this->campaignCollection = $campaignCollection;
+		$this->errorLogger = $errorLogger;
+		$this->ignoredBuckets = $ignoredBuckets;
+		$this->featureToggles = $featureToggles;
+	}
+
+	public function isPassing(): bool {
+		return empty( $this->getErrors() );
+	}
+
+	public function getErrors(): array {
+		$this->validate();
+
+		return $this->errorLogger->getErrors();
+	}
+
+	private function validate(): void {
+		if ( $this->hasValidated ) {
+			return;
+		}
+
+		$this->findInconsistentFeatureToggleOccurrences();
+
+		$filteredFeatureToggleChecks = array_unique( $this->featureToggles );
+		$configBucketIdList = $this->buildBucketIdsFromCampaignCollection(
+			$this->campaignCollection
+		);
+
+		/** Remove all buckets which were marked as test buckets in TEST_BUCKET_IGNORE_LIST */
+		$filteredFeatureToggleChecks = array_filter( $filteredFeatureToggleChecks, [ __CLASS__, 'filterBucketIds' ] );
+		$configBucketIdList = array_filter( $configBucketIdList, [ __CLASS__, 'filterBucketIds' ] );
+
+		/** Check if either the configuration files or the choice factory has rules not found in the other */
+		$codeConfigDiff = array_diff( $filteredFeatureToggleChecks, $configBucketIdList );
+		$configCodeDiff = array_diff( $configBucketIdList, $filteredFeatureToggleChecks );
+
+		$this->hasValidated = true;
+
+		if ( empty( $codeConfigDiff ) === true && empty( $configCodeDiff ) === true ) {
+			return;
+		}
+
+		foreach ( $codeConfigDiff as $missingBucket ) {
+			$this->errorLogger->addError(
+				'Feature toggle check for ' . $missingBucket . ' is implemented but no campaign configuration can be found.'
+			);
+		}
+		foreach ( $configCodeDiff as $featureToggleCheck ) {
+			$this->errorLogger->addError(
+				'Bucket ' . $featureToggleCheck . ' is configured but no implementation can be found in ChoiceFactory.'
+			);
+		}
+	}
+
+	private function buildBucketIdsFromCampaignCollection( CampaignCollection $campaignCollection ): array {
+		$buckets = [];
+		/** @var Campaign $campaign */
+		foreach ( $campaignCollection as $campaign ) {
+			foreach ( $campaign->getBuckets() as $bucket ) {
+				$buckets[] = $bucket->getId();
+			}
+		}
+		return $buckets;
+	}
+
+	private function filterBucketIds( string $bucketId ): bool {
+		return !( in_array( $bucketId, $this->ignoredBuckets ) );
+	}
+
+	/**
+	 * This method finds 'inconsistent' number of implementations for feature toggles checks
+	 * If you have implemented a check for a specific campaign bucket twice and only once for
+	 * a bucket of the same campaign, the ChoiceFactory is missing a code path for the campaign
+	 */
+	private function findInconsistentFeatureToggleOccurrences(): void {
+		$campaignBuckets = [];
+		foreach ( array_count_values( $this->featureToggles ) as $featureToggle => $featureToggleCount ) {
+			$campaign = $this->getCampaignNameFromFeatureToggle( $featureToggle );
+			if ( isset( $campaignBuckets[$campaign] ) && $campaignBuckets[$campaign] !== $featureToggleCount ) {
+				$this->errorLogger->addError(
+					'Campaign buckets for "' . $featureToggle . '" have not been implemented consistently.'
+				);
+				continue;
+			}
+			$campaignBuckets[$this->getCampaignNameFromFeatureToggle( $featureToggle )] = $featureToggleCount;
+		}
+	}
+
+	private function getCampaignNameFromFeatureToggle( string $featureToggle ): string {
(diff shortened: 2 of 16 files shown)


src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -0,0 +1,137 @@
...
+
+		$featureToggleChecks = $this->getFeatureToggleChecks();
+		if ( $duplicates = $this->getDuplicates( $featureToggleChecks ) ) {
+			foreach ( $duplicates as $duplicate ) {
+				$this->errorLogger->addError(
+					'ChoiceFactory contains multiple checks for feature toggle: ' . $duplicate
+				);
+			}
+			return;
+		}
```
gbirke: If a campaign necessitates multiple code changes (e.g. a change where both presentation and database model need to be changed), this duplicate check would fail. I still find the check useful and necessary and would leave it in, if the situation ever arises, we'll figure something out.

For now, the `return` is not strictly necessary, I'd go on checking even if there are duplicates. Maybe you need to call `array_uniq` for the following logic, but that's ok.

Also you move the `if` section to a private method, e.g. `$this->checkForDuplicateFeatureToggles( $featureToggleChecks )`?

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -0,0 +1,137 @@
...
+			}
+			$featureToggleChecks[] = $featureToggleCheck->args[0]->value->value;
+		}
+		return $featureToggleChecks;
+	}
+
+	private function parseMethodCalls(): array {
+		$parser = ( new ParserFactory() )->create( ParserFactory::PREFER_PHP7 );
+		$nodeFinder = new NodeFinder();
+		$choiceFactoryCode = file_get_contents( self::CHOICE_FACTORY_LOCATION );
```
gbirke: Please inject a [`FileFetcher`](https://packagist.org/packages/jeroen/file-fetcher) implementation for better testability and error handling (exceptions instead of errors).

src/BucketTesting/Validation/ValidationErrorLogger.php
```diff
@@ -22,6 +22,6 @@ public function getErrors(): array {
 	}
 
 	public function hasErrors(): bool {
-		return empty( $this->getErrors() );
+		return !empty( $this->getErrors() );
```
gbirke: Oops. Looks like this function was not tested in the last PR 😊

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -0,0 +1,137 @@
...
+			}
+		);
+	}
+
+	private function buildBucketIdsFromCampaignCollection( CampaignCollection $campaignCollection ): array {
+		$buckets = [];
+		/** @var Campaign $campaign */
+		foreach ( $campaignCollection as $campaign ) {
+			foreach ( $campaign->getBuckets() as $bucket ) {
+				$buckets[] = 'campaigns.' . $bucket->getId();
```
gbirke: This is the second place we the `campaigns` string comes out of the blue (the other, related place is `CampaignFeatureBuilder::getFeatureName`). I suggest we add the `campaigns` prefix in `Bucket::getId` and call that function in both places.

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -0,0 +1,137 @@
...
+
+		/** Remove all buckets which were marked as test buckets in TEST_BUCKET_IGNORE_LIST */
+		$featureToggleChecks = array_filter( $featureToggleChecks, [ __CLASS__, 'filterBucketIds' ] );
+		$configBucketIdList = array_filter( $configBucketIdList, [ __CLASS__, 'filterBucketIds' ] );
+
+		/** Check if either the configuration files or the choice factory has rules not found in the other */
+		$codeConfigDiff = array_diff( $featureToggleChecks, $configBucketIdList );
+		$configCodeDiff = array_diff( $configBucketIdList, $featureToggleChecks );
+
+		if ( empty( $codeConfigDiff ) === false || empty( $configCodeDiff ) === false ) {
```
gbirke: The code will become more readable (and less nested) if you invert the check, return early and then do the foreach iterations.

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -0,0 +1,137 @@
...
+				$this->errorLogger->addError(
+					'Feature toggle check for ' . $featureToggleCheck . ' is implemented but no campaign configuration can be found.'
+				);
+			}
+		}
+
+		$this->hasValidated = true;
+	}
+
+	private function getFeatureToggleChecks(): array {
```
gbirke: This class does too much IMHO: loading/parsing a factory file into bucket IDs **and** comparing the bucket IDs to the list of bucket IDs from the CampaignCollection. Could you introduce a class that only does the parsing and inject it as a dependency into the validator? Name suggestions: `BucketIdDetector`, `BucketIdParser`, etc.

Both classes can then be tested separately and you can inject the test for the validator with a fake.

The test for the parser class could have a constructor with file and method names as arguments (and its own constants as defaults). In the FunFunFactory you'd instantiate it without arguments, in the unit test you could pass in the test file itself and have a dummy method in the test with calls to `featureIsActive`. This way you'll avoid having a seemingly useless PHP file in `Tests\Data`.

tests/Unit/Cli/CampaignUtilizationValidatorTest.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\Cli;
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\BucketTesting\Validation\CampaignUtilizationValidator
+ */
+class CampaignUtilizationValidatorTest extends \PHPUnit\Framework\TestCase {
+
```
gbirke: The test should check if the validator still passes when the buckets and checks are not in the same order (both on campaign and bucket level):

```
campaigns.c1.foo
campaigns.c1.bar
campaigns.c2.baz
campaigns.c2.quux
```

should be similar to

```
campaigns.c2.baz
campaigns.c2.quux
campaigns.c1.foo
campaigns.c1.bar
```

and 

```
campaigns.c1.foo
campaigns.c1.bar
campaigns.c2.baz
campaigns.c2.quux
```

should be similar to

```
campaigns.c1.bar
campaigns.c1.foo
campaigns.c2.quux
campaigns.c2.baz
```

src/BucketTesting/Validation/ValidationErrorLogger.php
timEulitz: Haha, the code was correct but the method name was wrong, I have fixed that in both places now.

src/BucketTesting/Validation/CampaignUtilizationValidator.php
timEulitz: That obviously gets rid of all my worries in a much nicer way, thanks.

tests/Unit/Cli/CampaignUtilizationValidatorTest.php
timEulitz: I think I have incorporated that now in `newTestCampaignCollection()` but please let me know if there tests can be improved in that regard.

tests/Fixtures/ChoiceFactoryMock.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace WMDE\Fundraising\Frontend\Tests\Fixtures;
+
+use WMDE\Fundraising\Frontend\BucketTesting\FeatureToggle;
+use WMDE\Fundraising\Frontend\Factories\UnknownChoiceDefinition;
+
+/**
+ * @license GNU GPL v2+
+ */
+class ChoiceFactoryMock {
```
gbirke: In our test fixture code naming we try to be more precise what the code does, see https://www.entropywins.wtf/blog/2016/05/13/5-ways-to-write-better-mocks/ for the definitions of the different kinds of test doubles. In this case, I'd call the class `FakeChoiceFactory`.

tests/Fixtures/ChoiceFactoryMock.php
```diff
@@ -0,0 +1,40 @@
...
+
+	private $featureToggle;
+
+	public function __construct( FeatureToggle $featureToggle ) {
+		$this->featureToggle = $featureToggle;
+	}
+
+	public function someMockBucketSelection(): void {
```
gbirke: In the spirit of how the real `ChoiceFactory` works, could you return strings instead of `echo`ing stuff? Or is there a specific intention behind not returning anything to stress the fact that this is a test double?

tests/Unit/Cli/CampaignUtilizationValidatorTest.php
gbirke: 👍   The instantiation of buckets in `newTestCampaignCollection` is in a different order than the checks in `ChoiceFactoryMock`, that's good enough.

tests/Fixtures/ChoiceFactoryMock.php
timEulitz: I totally did not keep the spirit of the real ChoiceFactory in mind since I figured it does not really matter for code that is never actually executed and only used for static analysis, but I'll quickly change that anyway!

src/BucketTesting/Bucket.php
```diff
@@ -22,7 +22,7 @@ public function getName(): string {
 	}
 
 	public function getId(): string {
-		return $this->getCampaign()->getName() . '.' . $this->getName();
+		return 'campaigns.' . $this->getCampaign()->getName() . '.' . $this->getName();
```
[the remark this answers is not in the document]
timEulitz: Is the method not completely redundant now since it's just ´bucket->getId();´ or would you say it's cleaner if there is a specific ´getFeatureName()´ method that simply forwards the bucket ID?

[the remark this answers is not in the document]
gbirke: Yes, that's what i meant. `getFeatureName` is no longer necessary.

tests/Unit/Cli/CampaignUtilizationValidatorTest.php
```diff
@@ -0,0 +1,92 @@
...
+		$errorLogger = new ValidationErrorLogger();
+
+		$campaign01 = CampaignFixture::createCampaign();
+		CampaignFixture::createBucket( $campaign01, 'test_bucket_a', Bucket::DEFAULT );
+		CampaignFixture::createBucket( $campaign01, 'test_bucket_b', Bucket::NON_DEFAULT );
+
+		$campaign02 = CampaignFixture::createCampaign( 'another_test_campaign' );
+		CampaignFixture::createBucket( $campaign02, 'test_bucket_c', Bucket::DEFAULT );
+		CampaignFixture::createBucket( $campaign02, 'test_bucket_d', Bucket::NON_DEFAULT );
+
```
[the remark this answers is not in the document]
timEulitz: Hey, unfortunately there is no way to modify the array (without the usual unit-testing hackery which makes it even messier in my opinion), so it's probably best to just leave it as it is right now, since it does not really make sense to introduce any logic which would allow for the buckets to be modified during application runtime.

timEulitz:
@gbirke I have created a new FeatureToggleParser class and implemented a new check for inconsistency, see `findInconsistentFeatureToggleOccurrences()` for reference.

PR #1280, opened 2018-07-03
Update JSON schema validator library

timEulitz:
T194753


deployment/tasks/deploy_to_server.yml
```diff
@@ -23,10 +23,10 @@
       failed_when: decode_json_config.stdout != "0"
 
     - name: Validate config file against schema
-      command: ./console app:validate:config
+      command: ./console app:validate:config app/config/config.dist.json app/config/config.prod.json chdir={{ app_dir }}/{{ release_name }}
```
gbirke: Very good catch! Wow!

PR #1282, opened 2018-07-08
Create xDebug-ready Docker environment

timEulitz:


docker-compose.debug.yml
```diff
@@ -0,0 +1,23 @@
+version: '3.4'
+
+services:
+  proxy:
+      environment:
+        - NGINX_HOST=spenden.wikimedia.de
+        - NGINX_PORT=8080
+        - PHP_FPM_HOST=app_debug
```
gbirke: Remember to put this in the nginx config template.

timEulitz:
Works well, still needs to have some setup documentation / references to https://github.com/gbirke/php-xdebug-phpstorm-playground

gbirke:
+2

I've added the documentation from https://github.com/gbirke/php-xdebug-phpstorm-playground

You can un-WIP & merge this now or improve on the docs.

timEulitz:
Did you leave out the hints regarding the "hotfix" with the `.env` file we found on purpose or was this just an oversight and can still be added to the documentation? Other than that it looks good to me.

gbirke:
We're using the [`target` parameter of docker-compose](https://docs.docker.com/compose/compose-file/#target) to be able to build 2 versions of the same image (Dockerfile), the one with Xdebug building on the other.

I found the installation instructions at https://docs.docker.com/install/linux/docker-ce/ubuntu/ ok to follow. They seem longish, but that's because they cover all possibilites. It boils down 

1. Uninstall existing docker package
2. Add docker repo and install docker-ce *OR* download package and install it.

I can help if you want.

Alternatively, you can use and maintain your own version of docker-compose.yml that does not have the `target` parameter (you'll have Xdebug on by default then).

JeroenDeDauw:
I follow the instructions there and now have

image

Still getting

> ERROR: Version in "./docker-compose.yml" is unsupported.

Also looks like the README is in need of update, cause it is no longer as simple to run stuff as it suggests.

gbirke:
I'm sorry, I forgot that on Linux docker-compose has to be updated separately.

I've created a PR with the updated documentation: #1325

PR #1283, opened 2018-07-11
Implement address change functionality

timEulitz:
T198608

https://github.com/wmde/fundraising-donations/pull/61
https://github.com/wmde/fundraising-frontend-content/pull/112

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -53,6 +53,7 @@ use WMDE\Fundraising\DonationContext\DataAccess\DoctrineDonationTokenFetcher;
 use WMDE\Fundraising\DonationContext\DataAccess\UniqueTransferCodeGenerator;
 use WMDE\Fundraising\DonationContext\Domain\Repositories\CommentFinder;
 use WMDE\Fundraising\DonationContext\Domain\Repositories\DonationRepository;
+use WMDE\Fundraising\DonationContext\Domain\Validation\DonorValidator;
 use WMDE\Fundraising\DonationContext\DonationAcceptedEventHandler;
 use WMDE\Fundraising\DonationContext\Infrastructure\BestEffortDonationEventLogger;
 use WMDE\Fundraising\DonationContext\Infrastructure\DonationConfirmationMailer;
@@ -73,6 +74,8 @@ use WMDE\Fundraising\DonationContext\UseCases\GetDonation\GetDonationUseCase;
 use WMDE\Fundraising\DonationContext\UseCases\HandlePayPalPaymentNotification\HandlePayPalPaymentCompletionNotificationUseCase;
 use WMDE\Fundraising\DonationContext\UseCases\ListComments\ListCommentsUseCase;
 use WMDE\Fundraising\DonationContext\UseCases\SofortPaymentNotification\SofortPaymentNotificationUseCase;
+use WMDE\Fundraising\DonationContext\UseCases\UpdateDonor\UpdateDonorUseCase;
+use WMDE\Fundraising\DonationContext\UseCases\UpdateDonor\UpdateDonorValidator;
 use WMDE\Fundraising\DonationContext\UseCases\ValidateDonor\ValidateDonorUseCase;
 use WMDE\Fundraising\Frontend\BucketTesting\RandomBucketSelection;
 use WMDE\Fundraising\Frontend\Infrastructure\Cache\AllOfTheCachePurger;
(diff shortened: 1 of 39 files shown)


app/Controllers/UpdateAddressController.php
```diff
@@ -0,0 +1,39 @@
...
+namespace WMDE\Fundraising\Frontend\App\Controllers;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+
+/**
+ * @license GNU GPL v2+
+ */
+class UpdateAddressController {
```
gbirke: Please rename it to `UpdateDonorController`, as updating addresses will become its own thing which might or might not overlap with this.

skins/cat17/templates/Donation_Address_Update.twig
```diff
@@ -0,0 +1,26 @@
+{% extends 'Base_Layout.html.twig' %}
+
+{% block page_identifier %}address-update{% endblock %}
+
+{% block main %}
+	<div class="container">
+		<div class="row">
+			<div class="content col-xs-12 col-sm-10 col-md-8">
+				<h1>
+					{% if updateSuccessful %}
```
gbirke: I think moving the condition to the presenter and and assigning different translation keys there would make this template more readable.

It would also make the code more testable, as you could look for the translation keys in the output of the Edge2Edge test for the route and/or the Presenter Unit test.

src/Presentation/Presenters/CancelDonationHtmlPresenter.php
```diff
@@ -21,11 +18,9 @@ public function __construct( TwigTemplate $template ) {
 		$this->template = $template;
 	}
 
-	public function present( CancelDonationResponse $response ): string {
+	public function present( UpdateAddressResponse $response ): string {
```
gbirke: Is this change is accidental?

src/Factories/FunFunFactory.php
```diff
@@ -1130,6 +1140,22 @@ public function newDonationConfirmationPresenter(): DonationConfirmationHtmlPres
...
 	}
 
+	public function newDonorUpdatePresenter(): DonorUpdateHtmlPresenter {
+		return new DonorUpdateHtmlPresenter(
+			$this->getChoiceFactory()->getConfirmationPageTemplate(
+				$this->getSkinTwig(),
+				array_merge(
+					$this->getDefaultTwigVariables(),
+					[
+						'piwikGoals' => [ 3 ],
```
gbirke: What does `3` mean and is it the correct value for this page?

src/Presentation/DonorDataFormatter.php
```diff
@@ -0,0 +1,87 @@
...
+
+use WMDE\Fundraising\DonationContext\Domain\Model\Donation;
+use WMDE\Fundraising\PaymentContext\Domain\Model\BankTransferPayment;
+use WMDE\Fundraising\PaymentContext\Domain\Model\DirectDebitPayment;
+use WMDE\Fundraising\PaymentContext\Domain\Model\PaymentMethod;
+
+/**
+ * @license GNU GPL v2+
+ */
+class DonorDataFormatter {
```
gbirke: Currently, I can't think of a better term than `Formatter` (`PresentationExtractor` or `Converter` are not correct either, and `toArrayConverter` exposes too much of the low-level programming language). But formatters in my mind are more for direct output and not for "converting data for presentation". Maybe you can add a comment to the class explains what it does.

src/Factories/FunFunFactory.php
timEulitz: Goal 3 is the custom goal configured in Matomo for "completed", I assume that we do not want to re-send statistics for that goal so I will remove it here.

src/Presentation/Presenters/DonorUpdateHtmlPresenter.php
```diff
@@ -0,0 +1,74 @@
...
+	public function __construct( TwigTemplate $template, UrlGenerator $urlGenerator ) {
+		$this->template = $template;
+		$this->urlGenerator = $urlGenerator;
+		$this->donationMembershipApplicationAdapter = new DonationMembershipApplicationAdapter();
+		$this->donorDataFormatter = new DonorDataFormatter();
+	}
+
+	public function present( UpdateDonorResponse $updateDonorResponse, Donation $donation, string $updateToken, string $accessToken ): string {
+		return $this->template->render(
+			$this->getConfirmationPageArguments( $updateDonorResponse, $donation, $updateToken, $accessToken )
```
gbirke: Could you split the variables from `getConfirmationPageArguments` that are not part of the original confirmation page presenter into a separate function and use `array_merge` here to make it more clear what belongs where?

tests/EdgeToEdge/Routes/UpdateDonorRouteTest.php
```diff
@@ -0,0 +1,265 @@
...
+					$client,
+					$this->newPrivateDonorData(),
+					$donation->getId(),
+					self::CORRECT_UPDATE_TOKEN,
+					self::CORRECT_UPDATE_TOKEN
+				);
+				$response = $client->getResponse();
+				$this->assertTrue( $response->isSuccessful() );
+				$this->assertContains(
+					'Hans Wurst, Teststraße 123, 12345 Mönchengladbach',
```
gbirke: This can become brittle when the formatting changes (from comma-separated to smth. else, probably during the next ticket). But you can leave it for now.

tests/EdgeToEdge/Routes/UpdateDonorRouteTest.php
```diff
@@ -0,0 +1,265 @@
...
+		$factory->getDonationRepository()->storeDonation( $donation );
+
+		return $donation;
+	}
+
+	private function newPrivateDonorData(): array {
+		return [
+			'addressType' => DonorName::PERSON_PRIVATE,
+			'city' => 'Mönchengladbach',
+			'companyName' => '',
```
gbirke: You should leave out the empty fields, the browsers won't send them and our application should be robust enough to initialize the variables with empty strings.

PR #1285, opened 2018-07-17
ADR for ADRs

gbirke:
Document our decision to use ADRs and the preferred format and file
naming conventions.

Rename existing ADR to convention.


JeroenDeDauw:
> ADR for ADRs

Oh man... this is such a Gabriel title :D

PR #1288, opened 2018-07-20
Fix phpunit configuration code coverage whitelist

timEulitz:
Docs should probably also still be improved as part of this commit.


JeroenDeDauw:
Nice

PR #1289, opened 2018-07-23
[WIP] Use Vue and Vuex for the updateDonor functionality

gbirke:
This is a proof-of-concept for using Vue and Vuex for the client-side code.

TODO
- [X] Implement form HTML template
- [X] Modularize input field wrappers
- [X] Switch between person and company address type
- [X] Add Vuex store for storing the address field data
- [ ] Asynchronous validation:
  - Implement as Vuex action, keep validation results in vuex store. Use ternary state: true, false, null (=not validated yet)
  - Only send validation request when all fields are filled, but then send it in the background every change (maybe debounce it)
  - Cancel previous validation request when a new one is sent and the previous has not returned
  - The form is only valid when no validation requests are running and the most recent request returned positive.
- [ ] Implement house number "validation"/ hint message
- [ ] display field state: empty, filled, validation error
- [ ] Use Vue-native components for nice-looking [checkboxes](https://hamed-ehtesham.github.io/pretty-checkbox-vue/) and [dropdowns](https://github.com/shentao/vue-multiselect)
- [ ] Change button state to "active" when all fields are valid
- [ ] Split Webpack configuration into dev and prod parts:
   - Add eslint to dev
   - add uglify (and possibly extraction of CSS) to prod
- [ ] Add Webpack step into Gulpfile (as a stopgap measure, maybe we'll use Webpack to build all assets in the future)
- [ ] Rebase on current master
- [ ] Evaluate the solution - is it more readable and approachable?
- [ ] (Optional) Refactor CSS to use scoped CSS where it makes sense.

### DIFF
--- a/app/Controllers/UpdateDonorController.php
+++ b/app/Controllers/UpdateDonorController.php
@@ -1,0 +1,45 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\Controllers;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\DonationContext\UseCases\UpdateDonor\UpdateDonorRequest;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+
+/**
+ * @license GNU GPL v2+
+ */
+class UpdateDonorController {
+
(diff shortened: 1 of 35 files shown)


gbirke:
Abandoned, migrating to Vue and Vuex will now be a larger project. See https://phabricator.wikimedia.org/T219130

PR #1292, opened 2018-07-26
Create design for donor update use case

timEulitz:
Use `/?da=1&skin=0` to activate and `/?da=0&skin=0` respectively to deactivate this AB test.


JeroenDeDauw:
At 2:00? :) Should you not be asleep? :D

timEulitz:
You're not my momma!

app/Controllers/UpdateDonorController.php
```diff
@@ -21,11 +22,25 @@ public function updateDonor( Request $request, FunFunFactory $ffFactory ): Respo
 		$responseModel = $ffFactory
 			->newUpdateDonorUseCase( $updateToken, $accessToken )
 			->updateDonor( $this->newRequestModel( $request ) );
-		if ($responseModel->getDonation() === null) {
+		if ( $responseModel->getDonation() === null ) {
 			throw new AccessDeniedException();
 		}
+		if ( $responseModel->isSuccessful() ) {
+			return new RedirectResponse(
+				sprintf(
```
gbirke: Could you use the UrlGenerator from the FundraisingFactory?

skins/cat17/templates/partials/donation/donor_form.html.twig
```diff
@@ -1,28 +1,13 @@
 {% set COUNTRIES = [ 'DE', 'AT', 'CH', 'BE', 'IT', 'LI', 'LU' ] %}
-{% set PAYMENT_INTERVALS = [ 0, 1, 3, 6, 12 ] %}
 
-<div class="introduction container default">
-	<div class="row">
```
gbirke: Is this now intentionally missing from the main form?

tests/EdgeToEdge/Routes/UpdateDonorRouteTest.php
```diff
@@ -255,4 +259,12 @@ private function setDefaultSkin( FunFunFactory $factory, string $skinName ): voi
 			)
 		);
 	}
+
+	private function newValidSuccessRedirectUrl( Donation $donation ): string {
```
gbirke: You should either pass in the finished URL as dependency or the URLGenerator.

skins/cat17/templates/partials/donation/donor_form.html.twig
timEulitz: The actual donation form is located at skins/cat17/templates/Donation_Form.html.twig. This file is only for the confirmation page, I think I will rename it to make that more clear.

gbirke: Ah, I was under the assumption that you're re-using the `donor_form.html.twig` partial on `Donation_Form.html.twig` and that removing the text would break `Donation_Form.html.twig`.

Since I still think re-using the form fields is a good idea, I suggest the following: 

* Split `donor_form` into `partials/donor_update/donor_form.html.twig` and `partials/donation/donor_fields.html.twig`, the latter only has the container elements with the fields
* Replace donor field markup in `Donation_Form.html.twig` with an include of `partials/donation/donor_fields.html.twig`.

app/routes.php
```diff
@@ -1,581 +0,0 @@
-<?php
```
gbirke: Not really part of this PR, more like fixing #1288 where I did not pay attention to the bad interaction between  git and macOS case-insensitive file system. Could you place this deletion at least in a separate commit?

skins/cat17/templates/Donation_Confirmation.html.twig
```diff
@@ -13,10 +13,34 @@
 		{% include 'partials/donation_confirmation/bank_transfer.html.twig' %}
 	{% endif %}
 
-	{% include 'partials/donation_confirmation/membership_advantages.html.twig' %}
-	{% include 'partials/donation_confirmation/social_share.html.twig' %}
+	{% if featureToggle.donorUpdateEnabled == true and updateData.isSuccessful == false %}
```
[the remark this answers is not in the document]
timEulitz: @tzhelyazkova Hey 🅱️onina 🇧🇬, Twig 🌿 actually does 🚫 not 🚫 support the short-hand 🖐 negator ❗️ for variables 🖥. Instead, the correct ✅ approach 🤸‍ would be `if not updateData.isSuccessful` for example. I will change 🤞 these instances so it's more Twig-like 🌿🌿🌿.

timEulitz:
@tzhelyazkova Regarding your Chrome screenshot, I am fairly certain that it's due to the fact that your Chrome had the old CSS cached. The striped table CSS, for example, was completely removed by me and it's still visible in your screenshot. It also looks / works fine on my local Chrome.

skins/cat17/templates/partials/donation/donor_form.html.twig
timEulitz: Hey, I don't think that will work because technically the forms differ, the donor update form does not include the anonymous donor selector and I think your solution would mean that both forms share the exact same structure. Though maybe I am misunderstanding you? Or is that what you meant and I should just hide the anonymous selector with CSS for the donor update use case?

tzhelyazkova:
@timEulitz hmm, yeah, weird that `make ui` didn't clean the cache, as it did for Firefox.

skins/cat17/templates/partials/donation/donor_form.html.twig
gbirke: Ah ok, in this case just move the partial to a different directory.

While I'm not totally against hiding stuff with CSS, I prefer not to generate the markup if it does not complicate the code otherwise.

gbirke:
Not a requirement but nice 🍒 on top: Moving `ShowDonationConfirmationHandler` to `ShowDonationConfirmationController`

app/Routes.php
```diff
@@ -440,14 +440,8 @@ function( Request $request ) use ( $ffFactory ) {
...
-			function( Application $app, Request $request ) use ( $ffFactory ) {
-				return ( new ShowDonationConfirmationHandler( $ffFactory ) )->handle(
-					$request,
-					$app['session']->get( 'piwikTracking', [] )
-				);
-			}
-		)->bind( 'show-donation-confirmation' )
-			->method( 'GET|POST' );
+			ShowDonationConfirmationController::class . '::show'
+		)->bind( 'show-donation-confirmation' )->method( 'GET|POST' );
```
gbirke: Chained method calls can easily get lost so I prefer each method call on a new line.

skins/cat17/templates/Donation_Confirmation.html.twig
```diff
@@ -13,10 +13,34 @@
...
\ No newline at end of file
+{% endblock %}
+
+{% block scripts %}
+	<script src="{$ basepath|e('html_attr') $}{$ '/skins/cat17/scripts/wmde.js'|prefix_file $}"></script>
+	<script id="init-form" src="{$ basepath|e('html_attr') $}{$ '/skins/cat17/scripts/donorUpdateForm.js'|prefix_file $}"
+			data-initial-form-values="{% if initialFormValues %}{$ initialFormValues|json_encode|e('html_attr') $}{% else %}{}{% endif %}"
+			data-violated-fields="{% if violatedFields %}{$ violatedFields|json_encode|e('html_attr') $}{% else %}{}{% endif %}"
+			data-initial-validation-result="{% if validationResult %}{$ validationResult|json_encode|e('html_attr') $}{% else %}{}{% endif %}"
+			data-validate-amount-url="{$ basepath | e( 'html_attr' ) $}/validate-donation-amount"
```
gbirke: This url is not needed, as we're only validating email and address.

skins/cat17/templates/Donation_Confirmation.html.twig
```diff
@@ -13,10 +13,34 @@
...
+			data-validate-address-url="{$ basepath|e('html_attr') $}/validate-address"
+			data-validate-email-address-url="{$ basepath|e('html_attr') $}/validate-email"
+			data-validate-iban-url="{$ basepath|e('html_attr') $}/check-iban"
```
gbirke: These URLs are also not needed.

PR #1294, opened 2018-07-26
Fix donation summary overlay in tablet view when coming from a banner

tzhelyazkova:
The donation overview overlaps the message coming from the banner on the donation front page.
This was observed only on tablet vertical view.

Bug: T199111


timEulitz:
♻️ Reproduced ♻️ and confirmed ✅ fix for 🍎 iPad / iPad Mini 📱.

PR #1297, opened 2018-07-31
Adapt Scrutinizer settings

timEulitz:
https://phabricator.wikimedia.org/T198705

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -16,5 +16,17 @@ filter:
     excluded_paths:
         - 'vendor/*'
         - 'web/skins/*'
-        - 'skins/10h16/web/_js/lib/*'
+        - 'skins/10h16/*'
         - 'node_modules/*'
+
+build:
+    nodes:
+        analysis:
+            tests:
+                override:
+                - php-scrutinizer-run
+    tests:
+        override:
+        - true
+
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,9 @@ matrix:
   fast_finish: true
   include:
   - php: 7.2
-    # Disabled since we hit the 50 minute timeout
-#  - php: 7.1
-#    env: DB=sqlite; TYPE=coverage
-#  allow_failures: # We allow this one to fail so fast_finish causes the build to not wait for it
-#  - env: DB=sqlite; TYPE=coverage
+    env: TYPE=coverage
+  # More PHP instances with different environment variables may be added here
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -103,7 +106,7 @@ migration-revert:
 migration-status:
 	docker-compose run --rm --no-deps app vendor/doctrine/migrations/bin/doctrine-migrations migrations:status --configuration=vendor/wmde/fundraising-store/migrations.yml
 
-ci: covers phpunit cs npm-ci validate-app-config validate-campaign-config stan
+ci: covers phpunit-with-coverage cs npm-ci validate-app-config validate-campaign-config stan
```
gbirke: I often run `make ci` before committing, so I'd like us to have a different CI command without coverage. Maybe `ci` and `ci-with-coverage`?

.travis.yml
```diff
@@ -9,11 +9,7 @@ matrix:
   fast_finish: true
   include:
   - php: 7.2
-    # Disabled since we hit the 50 minute timeout
-#  - php: 7.1
-#    env: DB=sqlite; TYPE=coverage
-#  allow_failures: # We allow this one to fail so fast_finish causes the build to not wait for it
-#  - env: DB=sqlite; TYPE=coverage
+    env: TYPE=coverage
```
gbirke: Thanks for cleaning up the test matrix. Now that we're down to 1 case, could you add a comment outlining possible uses, e.g. PHP 7.3

Makefile
```diff
@@ -105,6 +108,8 @@ migration-status:
 
 ci: covers phpunit cs npm-ci validate-app-config validate-campaign-config stan
 
+ci-with-coverage: covers phpunit-with-coverage cs npm-ci validate-app-config validate-campaign-config stan
```
gbirke: I think you need to add this to the [.PHONY targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

.scrutinizer.yml
```diff
@@ -16,5 +16,17 @@ filter:
...
+
+build:
+    nodes:
+        analysis:
+            tests:
+                override:
+                - php-scrutinizer-run
+    tests:
+        override:
+        - true
```
JeroenDeDauw: What does this change do?

timEulitz: It disables unit testing on Scrutinizer, since we're doing that on Travis which then sends the code coverage results to Scrutinizer, no need to have them run twice.

PR #1298, opened 2018-07-31
Add ADR 005 about the URL Schema

gbirke:


gbirke:
It's 005 because 003 and 004 are in #1287

PR #1302, opened 2018-08-03
Various code quality improvements

timEulitz:
T198705

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -4,7 +4,7 @@ tools:
     external_code_coverage:
         timeout: 3600
     php_code_sniffer: true
-    php_cpd: true
+    php_cpd: false # PHP Copy/Paste Detector and PHP Code Similarity Analyzer cannot both be used simultaneously
     php_cs_fixer: true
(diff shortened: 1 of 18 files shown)


.scrutinizer.yml
```diff
@@ -29,6 +29,4 @@ build:
                 - php-scrutinizer-run
     tests:
         override:
-        - true
-
-
+        - true # Disable unit tests because they are handled by Travis
```
tzhelyazkova: set to `false` :question:

timEulitz: Oddly enough, setting "true" will skip this according to the docs.

app/RouteHandlers/PageDisplayHandler.php
```diff
@@ -26,7 +26,7 @@ public function __construct( FunFunFactory $ffFactory, Application $app ) {
 		$this->app = $app;
 	}
 
-	function handle( $pageName ) {
```
JeroenDeDauw: wow. It surprises me we have a method without visibility o_O

PR #1308, opened 2018-08-21
Emphasize confirmation page info based on context

timEulitz:
https://phabricator.wikimedia.org/T202259

Quick explanation here on why I implemented it the way it is: Since there currently is no real user session or persistence, I decided to implement a session cookie for this particular feature. After some thinking and debating, I felt like using a session cookie would be the best choice here, mostly because it was the least painful solution. 

The alternatives I saw were modifying the URL which I think is kind of ugly and allows users to easily modify site behavior which they are not supposed to be able to control. Modifying the database was another alternative so that we could persist some sort of "address was changed" flag or even an address update timestamp which would allow us to determine if a user had updated their address and returned to the confirmation page.

One advantage of the session cookie is also that if a user bookmarks the URL and comes back to it later to print it or view the payment information, the highlighting around the address is gone and the highlighting around the payment will be back.

I accessed the cookie via Twig instead of accessing it via PHP directly, I added a new wrapper div around the payment methods with the sole purpose of being able to toggle the highlighting states via one CSS class instead of having 5 separate if statements across all the different payment method templates, this seemed like the least painful approach since I would not have to find a way to access the request deep into the PHP logic nor be worried of making this code too error-prone to refactor in the future. However, if you think this is ugly and you know of a better way, let me know.

### DIFF
--- a/app/RouteHandlers/AddDonationHandler.php
+++ b/app/RouteHandlers/AddDonationHandler.php
@@ -12,6 +12,7 @@ use WMDE\Fundraising\DonationContext\Domain\Model\DonationTrackingInfo;
 use WMDE\Fundraising\DonationContext\UseCases\AddDonation\AddDonationRequest;
 use WMDE\Fundraising\DonationContext\UseCases\AddDonation\AddDonationResponse;
 use WMDE\Fundraising\Frontend\App\Controllers\ShowDonationConfirmationController;
+use WMDE\Fundraising\Frontend\App\Controllers\UpdateDonorController;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\Infrastructure\AmountParser;
 use WMDE\Fundraising\PaymentContext\Domain\Model\BankData;
@@ -51,6 +52,7 @@ class AddDonationHandler {
 		}
 
 		$this->sendTrackingDataIfNeeded( $request, $responseModel );
+		$this->resetSessionState();
 
 		return $this->newHttpResponse( $responseModel );
 	}
@@ -220,5 +222,15 @@ class AddDonationHandler {
 	 */
 	private function filterAutofillCommas( string $value ): string {
 		return trim( preg_replace( ['/,/', '/\s{2,}/'], [' ', ' '], $value ) );
+	}
+
+	/**
+	 * Reset session data to prevent old donations from changing the application output due to old data leaking into the new session
+	 */
+	private function resetSessionState(): void {
+		$this->app['session']->set(
+			UpdateDonorController::ADDRESS_CHANGE_SESSION_KEY,
+			false
+		);
 	}
 }
--- a/tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
+++ b/tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
@@ -4,11 +4,15 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\Tests\EdgeToEdge\Routes;
 
+use Silex\Application;
 use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\HttpKernel\Client;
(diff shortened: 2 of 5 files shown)


timEulitz:
Also, I was wondering if we really need tests. The ticket makes this problem sound much more complicated than it really is with this solution. Since the cookie is only set when the address is actually changed, all the scenarios described in the Phabricator task have been correctly implemented since the only way to get the cookie is to actually change the address and the only way to get there is by being in the right test group and the right payment method.

JeroenDeDauw:
screenshot_20180821_204027

timEulitz:
@tzhelyazkova @JeroenDeDauw I just have to cover all my bases because Gabriel will read it will I am on my day off :D.

app/RouteHandlers/AddDonationHandler.php
```diff
@@ -221,4 +222,15 @@ private function newTrackingInfoFromRequest( Request $request ): DonationTrackin
...
 		return trim( preg_replace( ['/,/', '/\s{2,}/'], [' ', ' '], $value ) );
 	}
+
+	/**
+	 * Since the application has no concept of logging out or resetting a session, we reset any session cookies at this point
+	 *
+	 * @param Response $response
+	 * @return Response
+	 */
+	private function resetSessionCookies( Response $response ) {
```
gbirke: I found this method a bit confusing, because it's not clear *why* we're resetting the cookie (because we want to differentiate between two "paths" to the confirmation page)

gbirke:
> Also, I was wondering if we really need tests.

It is an application behavior ("show different highlights, depending on previous actions and state"), so it needs to be tested. Since we don't have browser tests at the moment, I suggest adding test cases to the E2E tests:
- Checking for the existence of the `address-change` CSS class when the session flag is set. I suggest 2 tests in `ShowDonationConfirmationRouteTest`, setting the skin from `test` to `cat17`. See `DisplayPageRouteTest` for an example on how to switch the skin
- Checking the session changes in `AddDonationRouteTest` and `UpdateDonorRouteTest`.

tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
```diff
@@ -228,4 +231,56 @@ public function testWhenDonationTimestampCookiePreexists_itIsNotOverwritten(): v
...
+
+	public function testWhenAddressIsNotUpdated_addressHighlightIsNotShown(): void {
+		$this->createAppEnvironment( [], function ( Client $client, FunFunFactory $factory, Application $app ): void {
+			$this->setDefaultSkin( $factory, 'cat17' );
+			$app['session']->set(
+				UpdateDonorController::ADDRESS_CHANGE_SESSION_KEY,
+				false
+			);
+			$donation = $this->newStoredDonation( $factory );
+			$client->request(
```
tzhelyazkova: This could become `$this->showDonationConfirmation()` since it's used twice.

app/RouteHandlers/AddDonationHandler.php
```diff
@@ -221,4 +223,14 @@ private function newTrackingInfoFromRequest( Request $request ): DonationTrackin
 	private function filterAutofillCommas( string $value ): string {
 		return trim( preg_replace( ['/,/', '/\s{2,}/'], [' ', ' '], $value ) );
 	}
+
+	/**
+	 * Since the application has no concept of sessions or logging in / out, we reset session data here
```
gbirke: The comment has now become contradictory. Can you please state why this exists, e.g. "Unset 'address was changed' state in session, to display the right state on teh confirmation page."

tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
```diff
@@ -228,4 +231,56 @@ public function testWhenDonationTimestampCookiePreexists_itIsNotOverwritten(): v
...
+				'GET',
+				'show-donation-confirmation',
+				[
+					'id' => $donation->getId(),
+					'accessToken' => self::CORRECT_ACCESS_TOKEN
+				]
+			);
+			$crawler = $client->getCrawler();
```
gbirke: `$client->request` returns the `$crawler` instance, you could remove this line.

PR #1310, opened 2018-08-23
Log Buckets

gbirke:
Add BucketLogWriter interface and a file-based implementation.
Add Logging events for adding donations and membership applications.
Explain how to process the log file for getting a CSV file.

Please do a "Squash and merge" for this PR.

### DIFF
--- a/src/BucketTesting/Logging/StreamBucketLogger.php
+++ b/src/BucketTesting/Logging/StreamBucketLogger.php
@@ -20,19 +20,22 @@ class StreamBucketLogger implements BucketLogger {
 	}
 
 	public function writeEvent( LoggingEvent $event, Bucket ...$buckets ) {
-		fwrite( $this->stream, json_encode( [
-			'date' => date( $this->dateFormat ),
-			'eventName' => $event->getName(),
-			'metadata' => $event->getMetaData(),
-			'buckets' => array_reduce(
-				$buckets,
-				function ( array $bucketMap, Bucket $bucket ) {
-					$bucketMap[$bucket->getCampaign()->getName()] = $bucket->getName();
-					return $bucketMap;
-				},
-				[]
-			)
-		]));
+		fwrite(
+			$this->stream,
+			json_encode( [
+				'date' => date( $this->dateFormat ),
+				'eventName' => $event->getName(),
+				'metadata' => $event->getMetaData(),
+				'buckets' => array_reduce(
+					$buckets,
+					function ( array $bucketMap, Bucket $bucket ) {
+						$bucketMap[$bucket->getCampaign()->getName()] = $bucket->getName();
+						return $bucketMap;
+					},
+					[]
+				)
+			] ) . "\n"
+		);
 	}
 
 	public function setDateFormat( string $dateFormat ): void {
--- a/src/BucketTesting/Logging/StreamBucketLogger.php
+++ b/src/BucketTesting/Logging/StreamBucketLogger.php
@@ -2,11 +2,12 @@
 
 declare( strict_types = 1 );
 
-namespace WMDE\Fundraising\Frontend\BucketTesting;
+namespace WMDE\Fundraising\Frontend\BucketTesting\Logging;
 
 use DateTime;
+use WMDE\Fundraising\Frontend\BucketTesting\Bucket;
 
-class StreamBucketLogWriter implements BucketLogWriter {
+class StreamBucketLogger implements BucketLogger {
 
 	private $stream;
 	private $dateFormat = DateTime::RFC3339_EXTENDED;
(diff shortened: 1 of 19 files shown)


src/BucketTesting/StreamBucketLogWriter.php
```diff
@@ -0,0 +1,37 @@
...
+
+	public function writeEvent( string $eventName, array $eventMetadata, Bucket ...$buckets ) {
+		fwrite( $this->stream, json_encode( [
+			'date' => date( $this->dateFormat ),
+			'eventName' => $eventName,
+			'metadata' => $eventMetadata,
+			'buckets' => array_reduce( $buckets, function ( array $bucketMap, Bucket $bucket ) {
+				$bucketMap[$bucket->getCampaign()->getName()] = $bucket->getName();
+				return $bucketMap;
+			}, [] )
```
JeroenDeDauw: One argument per line would be more readable here for me

tests/Unit/BucketTesting/StreamBucketLogWriterTest.php
```diff
@@ -0,0 +1,73 @@
...
+class StreamBucketLogWriterTest extends TestCase {
+
+	public function testLogWriterAddsDate() {
+		$logfile = fopen( 'php://memory', 'a+' );
+		$logWriter = new StreamBucketLogWriter( $logfile );
+		$logWriter->setDateFormat( 'Y-m-d H:i' ); // Ignore seconds to avoid Heisentests
+
+		$logWriter->writeEvent( 'testLogged', [], ...[] );
+
+		$this->assertLogValue( date( 'Y-m-d H:i' ), 'date', $logfile );
```
tzhelyazkova: pretty neat that you extracted the assertion into a function :+1:

tests/Unit/BucketTesting/StreamBucketLoggerTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\BucketTesting\Logging\StreamBucketLogger
+ */
+class StreamBucketLoggerTest extends TestCase {
+
+	public function testLogWriterAddsDate() {
+		$logfile = fopen( 'php://memory', 'a+' );
+		$logWriter = new StreamBucketLogger( $logfile );
+		$logWriter->setDateFormat( 'Y-m-d H:i' ); // Ignore seconds to avoid Heisentests
```
JeroenDeDauw: haha :)

Isn't the issue still there though? Just 60 times less likely to happen?

tests/Unit/BucketTesting/StreamBucketLoggerTest.php
```diff
@@ -0,0 +1,113 @@
...
+use WMDE\Fundraising\Frontend\BucketTesting\Logging\StreamBucketLogger;
+use WMDE\Fundraising\Frontend\Tests\Fixtures\FakeBucketLoggingEvent;
+
+/**
+ * @covers \WMDE\Fundraising\Frontend\BucketTesting\Logging\StreamBucketLogger
+ */
+class StreamBucketLoggerTest extends TestCase {
+
+	public function testLogWriterAddsDate() {
+		$logfile = fopen( 'php://memory', 'a+' );
```
JeroenDeDauw: What about creating this in setUp? I did not check fully if that makes sense but noticed low level details in test method "arrange" part and repetition. Even if setUp is not suitable extracting this to avoid the low level details might be good.

src/BucketTesting/Logging/StreamBucketLogger.php
```diff
@@ -0,0 +1,45 @@
...
+						$bucketMap[$bucket->getCampaign()->getName()] = $bucket->getName();
+						return $bucketMap;
+					},
+					[]
+				)
+			] ) . "\n"
+		);
+	}
+
+	public function setDateFormat( string $dateFormat ): void {
```
JeroenDeDauw: Why is this not a constructor parameter?

src/BucketTesting/Logging/StreamBucketLogger.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	public function __construct( $stream ) {
+		$this->stream = $stream;
+	}
+
+	public function writeEvent( LoggingEvent $event, Bucket ...$buckets ) {
+		fwrite(
+			$this->stream,
+			json_encode( [
+				'date' => date( $this->dateFormat ),
```
JeroenDeDauw: This is a global function that relies on global state that gets changed by the system. As you noticed this causes issues when trying to test this code. What about creating a TimeThinghy service? That way this code no longer binds to that changing global state and also no longer needs to know about the time format.

tests/Unit/BucketTesting/StreamBucketLoggerTest.php
timEulitz: The execution time where this is relevant is just a few milliseconds (maybe even less), so the odds of this being a problem is like a 1 in 60000 chance which should be good enough, otherwise we should start considering ditching Wikipedia and start playing the lottery.

JeroenDeDauw: No need to have this chance at all though

src/BucketTesting/Logging/StreamBucketLogger.php
JeroenDeDauw: I will make a PR into this one with a SuchSuperTimeThinghy

JeroenDeDauw:
I addressed my own comments in https://github.com/wmde/FundraisingFrontend/pull/1315, which is a PR into this one. So please review that one before merging this one, else we'll need to rebase and create a fresh PR, since GH does not handle that situation well.

PR #1312, opened 2018-08-28
Specify language version for Scrutinizer

gbirke:
Specify PHP and Node version, to avoid scrutinizer errors on outdated Node versions


gbirke:
Closing, because scrutinizer uses the most recent language version by default.

PR #1313, opened 2018-08-28
Specify return type in FeatureToggleParser

gbirke:
$nodeFinder->find returns node interface, but we're returning concrete
implementations (MethodCall). This change will hopefully appease syntax
checkers and make it more clear what's returned.


JeroenDeDauw:
All the Scrutinizer things :)

gbirke:
Unfortunately, Scrutinizer ignores the doc block. But I'll merge anyway to make the return type more clear.

JeroenDeDauw:
You sure it ignores it? I thought it just updates its issues stuff when things are merged into master. Not looked at that in a long time though.

PR #1314, opened 2018-08-28
Ignore FunFunFactory in Scrutinizer

gbirke:
FunFunFactory brings down our score, but since it's our central entry
point, and *known* to be "too long" we don't care.


JeroenDeDauw:
Too much fun for scrutinizer :)

PR #1317, opened 2018-08-31
Share bucket log between deployments

gbirke:
Create a shared directory between deployments so A/B testing data is not scattered between deployments.

Please note:
* I have tested the deployment script on the test machine (checked out the `share-bucket-log` branch on the deployment server, ran the new playbook with `--extra-vars 'build_branch=share-bucket-log'`, checked out master again)
* The `shared` directory already existed on test and prod, it was created by earlier versions of the deployment script. [PR 229](https://github.com/wmde/fundraising-infrastructure/pull/229) on the infrastructure repo that codifies the existence of the directory.
* I've added the `shared_dir` variable to the inventories on the deploy server.

This is a followup for https://phabricator.wikimedia.org/T172119


src/Infrastructure/ErrorLoggingStreamOpener.php
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * @return resource
+	 */
+	public function openStream( string $url, string $mode ) {
+		try {
+			return $this->streamOpener->openStream( $url, $mode );
+		} catch ( StreamOpeningError $e ) {
+			$this->logger->error( $e->getMessage(), [ 'url' => $url, 'mode' => $mode ] );
+			return fopen( 'php://temp', $mode );
```
timEulitz: You may want to change that doc comment to `@return resource|bool` since it will return either of those.

JeroenDeDauw:
Is there a reason to inject the `resource $stream` into  `StreamBucketLogger` besides testing? If not, I suspect this is not the best way to abstract things.

* StreamBucketLogger mixes a call to a global function with side effects with logic
* Currently the stream needs to be created during object graph construction, while writes might not happen
* It is nice if StreamBucketLogger can hide the stream details fully
* Nothing is calling fclose. Not sure that is a problem but it seems dodgy

It seems more natural to me to do the fopen inside of StreamBucketLogger when it is needed, and have fclose in __destruct or something. That way the stream is never exposed outside the object and you are not forced to construct it early.

That leaves the problem of mixing the global function calls with logic and making testing a pain. Unless I'm forgetti'ng someting, both can be avoided by wrapping the stream stuff in some StreamStuff interface.

JeroenDeDauw:
Something like this

```php
class StreamStuff {

	private $streamOpeningInfo;
	private $stream;

	public function __construct( Whatever $streamOpeningInfo ) {
		$this->streamOpeningInfo = $streamOpeningInfo;
	}

	public function writeStuff( string $stuff ) {
		$this->openStreamIfNeeded();
		fwrite( $this->stream, $stuff );
	}

	private function openStreamIfNeeded() {
		if ( $this->stream === null ) {
			$this->stream = fopen( $this->streamOpeningInfo );
		}
	}

	public function __destruct() {
		fclose( $this->stream );
	}

}
```

You can have a Stub/Null implementation of the interface and you can have a simple in memory fake where needed. Logging and ignoring of errors are also easy to add, either in the real implementation or a Decorator.

gbirke:
Thanks for your comment, @JeroenDeDauw,  especially about opening the stream too early. I started out with something like this and then got scared of using `fopen` inside a class and got carried away when I tried to make it more resilient and abstracted. Your comment put me back on the right track.

src/BucketTesting/Logging/StreamBucketLogger.php
```diff
@@ -4,35 +4,57 @@
...
+			json_encode( $this->formatData( $event, ...$buckets ) ) . "\n"
 		);
 	}
 
+	private function openStreamIfNeeded() {
+		if ( $this->stream !== null ) {
+			return;
+		}
+		$this->createPathIfNeeded();
+		$this->stream = @fopen( $this->url, 'a' );
```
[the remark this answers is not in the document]
gbirke: I prefer the "silence error, throw exception on false" approach over too much checking. My reasons:

- it keeps the code small
- You can't protect yourself against all the exotic reasons why a call to `fopen` can fail ( see http://man7.org/linux/man-pages/man2/open.2.html ), checking some of them would give a false sense of security.
- If you check beforehand, you can experience race conditions where the system's circumstances changed between the checks and the `fopen`.

An alternative to the evil `@` operator would be "AtEase": https://github.com/wikimedia/at-ease

src/BucketTesting/Logging/BestEffortBucketLogger.php
```diff
@@ -0,0 +1,36 @@
...
+
+	public function writeEvent( LoggingEvent $event, Bucket ...$buckets ): void {
+		if ( $this->caughtException !== null ) {
+			return;
+		}
+		try {
+			$this->bucketLogger->writeEvent( $event, ...$buckets );
+		} catch ( LoggingError $error ) {
+			$this->caughtException = $error;
+			$this->errorLogging->error( $error->getMessage(), [ 'exception' => $error ] );
```
JeroenDeDauw: So this logs that the logging failed? Hahaha... 

Nvm me, just carry on

gbirke:
> Mind if I make a small follow up PR tomorrow?

Go ahead, I'm courious

JeroenDeDauw:
Any reason this was not merged yet? (I always get slightly triggered when needing to work off a branch instead off master)

JeroenDeDauw:
https://github.com/wmde/FundraisingFrontend/pull/1322

gbirke:
Good to know your preferences! I thought I was doing you a favor by leaving it unmerged so you can base your own PRs on it 😬 

I hope "triggered" in this case is more like "oh no, It's so frustrating when someone does this" and not the awful emotional experiences I associate with that word.

gbirke:
Now I'm waiting for the CI to finish, then I'll merge.

JeroenDeDauw:
*/me is triggered again* ? :D

> oh no, It's so frustrating when someone does this

Less severe then that actually :) Just a tiny little bit frustrating

PR #1322, opened 2018-09-07
Split stream writing code from JSON encoding

JeroenDeDauw:

PR #1323, opened 2018-09-07
Close stream in StreamLogWriter

JeroenDeDauw:
PR into https://github.com/wmde/FundraisingFrontend/pull/1322

Not sure how to test this one. Also did not find any good explanation of what can go wrong when the thing is not closed, but assuming there is reason to do so since it is in every fopen tutorial.


gbirke:
https://stackoverflow.com/questions/6038764/why-do-i-need-fclose-after-writing-to-a-file-in-php

PR #1326, opened 2018-09-07
Add PHPMyAdmin docker container example to README

gbirke:
Also fix wrong (and changing) container name in description by using a
placeholder and describing how to get the value for the placeholder.


gbirke:
My reasoning behind choosing 8099 for PHPMyAdmin was that I would like the 808x ports to be reserved for our own projects. I consider PHPMyAdmin to be a "third-party" project and chose 8099 because I consider URLs starting with an 80 as a hint that it might be a HTTP-based service and "99" for "not important".

It's not that important to me, if you prefer 8089 you can merge. Otherwise one of us can remove the commit, force-push and merge.

tzhelyazkova:
`8099` it is. It's just that the text did not match the docker command. I've updated the commit.

PR #1327, opened 2018-09-11
Fix newsletter checkbox id

tzhelyazkova:
Newsletter id attribute should be different for `Privatperson` and `Firma` fieldsets.


tzhelyazkova:
Self-merging because this is a small change meant to fix an issue in an A/B test.

PR #1336, opened 2018-09-21
Update packages

gbirke:


JeroenDeDauw:
Think it is safe for me to merge this one

JeroenDeDauw:
And otherwise there will be conflict with the follow up PR I need to make for the membership BC change

PR #1337, opened 2018-09-24
 Match changes in Membership BC

JeroenDeDauw:
wmde/fundraising-memberships#23

### DIFF
--- a/app/Controllers/ValidateFeeController.php
+++ b/app/Controllers/ValidateFeeController.php
@@ -1,0 +1,60 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\Controllers;
+
+use InvalidArgumentException;
+use Silex\Application;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Euro\Euro;
+use WMDE\Fundraising\MembershipContext\UseCases\ValidateMembershipFee\ValidateFeeRequest;
+use WMDE\Fundraising\MembershipContext\UseCases\ValidateMembershipFee\ValidateMembershipFeeUseCase;
+
+class ValidateFeeController {
+
+	private $app;
+	private $httpRequest;
+
+	public function __construct( Application $app, Request $httpRequest ) {
+		$this->app = $app;
+		$this->httpRequest = $httpRequest;
+	}
+
+	public function validateFee(): Response {
+		try {
+			$fee = Euro::newFromString(
+				str_replace( ',', '.', $this->httpRequest->request->get( 'amount', '' ) )
+			);
+		}
+		catch ( InvalidArgumentException $ex ) {
+			return $this->app->json( [
+				'status' => 'ERR',
+				'messages' => [
+					'amount' => 'not-money'
+				]
+			] );
+		}
+
+		$request = ValidateFeeRequest::newInstance()
+			->withFee( $fee )
+			->withInterval( (int)$this->httpRequest->request->get( 'paymentIntervalInMonths', '0' ) )
+			->withApplicantType( $this->httpRequest->request->get( 'addressType', '' ) );
+
+		$response = ( new ValidateMembershipFeeUseCase() )->validate( $request );
+
+
+		if ( $response->isSuccessful() ) {
+			return $this->app->json( [ 'status' => 'OK' ] );
+		}
+
+		return $this->app->json( [
+			'status' => 'ERR',
+			'messages' => [
+				'amount' => 'too-low'
+			]
+		] );
+	}
+
+}
--- a/app/Controllers/ValidateFeeController.php
+++ b/app/Controllers/ValidateFeeController.php
@@ -24,17 +24,10 @@ class ValidateFeeController {
 
 	public function validateFee(): Response {
 		try {
-			$fee = Euro::newFromString(
-				str_replace( ',', '.', $this->httpRequest->request->get( 'amount', '' ) )
-			);
+			$fee = $this->euroFromRequest();
 		}
 		catch ( InvalidArgumentException $ex ) {
-			return $this->app->json( [
-				'status' => 'ERR',
-				'messages' => [
-					'amount' => 'not-money'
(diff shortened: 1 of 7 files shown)


app/Controllers/ValidateFeeController.php
```diff
@@ -0,0 +1,65 @@
...
+			->withInterval( (int)$this->httpRequest->request->get( 'paymentIntervalInMonths', '0' ) )
+			->withApplicantType( $this->httpRequest->request->get( 'addressType', '' ) );
+
+		$response = ( new ValidateMembershipFeeUseCase() )->validate( $request );
+
+		if ( $response->isSuccessful() ) {
+			return $this->app->json( [ 'status' => 'OK' ] );
+		}
+
+		return $this->newJsonErrorResponse( 'too-low' );
```
gbirke: This relies on the implicit knowledge that the only error response is "amount too low". In the unlikely event that `ValidateFeeResult` gets another negative validation response, we will display the wrong reason here. 

It's a tradeoff between correctness and brevity. This code works and is unlikely to produce wrong output in the foreseeable future. The "safe" approach would be a `createJsonErrorResponse` which maps the the possible error codes of `ValidateFeeResult` to translation messages and throws an exception for all others. What do you think?

app/Controllers/ValidateFeeController.php
```diff
@@ -0,0 +1,65 @@
...
+
+		$request = ValidateFeeRequest::newInstance()
+			->withFee( $fee )
+			->withInterval( (int)$this->httpRequest->request->get( 'paymentIntervalInMonths', '0' ) )
+			->withApplicantType( $this->httpRequest->request->get( 'addressType', '' ) );
+
+		$response = ( new ValidateMembershipFeeUseCase() )->validate( $request );
+
+		if ( $response->isSuccessful() ) {
+			return $this->app->json( [ 'status' => 'OK' ] );
```
gbirke: Since `app` is on its way out for the Silex -> Symfony transition, please initialize the `JsonResponse` directly. See `ValidateDonorController` for an example.

app/RouteHandlers/ApplyForMembershipHandler.php
```diff
@@ -33,34 +34,38 @@ public function __construct( FunFunFactory $ffFactory, Application $app ) {
 		$this->app = $app;
 	}
 
-	public function handle( Request $request ): Response {
-		if ( !$this->isSubmissionAllowed( $request ) ) {
+	public function handle( Request $httpRequest ): Response {
```
gbirke: We could take to opportunity of having to touch this code to convert this "Handler" into a "Controller".

app/RouteHandlers/ApplyForMembershipHandler.php
```diff
@@ -134,35 +164,55 @@ private function isSubmissionAllowed( Request $request ) {
...
-						'show-membership-confirmation',
-						[
-							'id' => $responseModel->getMembershipApplication()->getId(),
-							'accessToken' => $responseModel->getAccessToken()
-						]
-					),
-					Response::HTTP_SEE_OTHER
-				);
-
+				$httpResponse = $this->newDirectDebitResponse( $responseModel );
```
gbirke: 👍  I like the smaller method(s)! 💯

src/Factories/FunFunFactory.php
```diff
@@ -140,6 +140,7 @@
 use WMDE\Fundraising\Frontend\Presentation\Presenters\PageNotFoundPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\DonorUpdateHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\TwigTemplate;
+use WMDE\Fundraising\MembershipContext\UseCases\ValidateMembershipFee\ValidateMembershipFeeUseCase;
```
gbirke: You could use the opportunity to move all the other membership BC classes up here.

app/Controllers/ValidateFeeController.php
JeroenDeDauw: Step one here was having a lookup in a map with two elements. Then one got removed. So as final step I dropped that lookup part. Which indeed leaves that not as explicit. I suppose you can wrap this thing in a guard if and then throw an exception at the end of the function. Not sure that is worth it though.

JeroenDeDauw:
Can this be merged already before follow up?

Might get to follow up before you start working again, might not (WD stuff to do as well)

JeroenDeDauw:
OMG only just noticed
image

Do I get a prize?

app/Routes.php
```diff
@@ -86,19 +89,7 @@ function ( Request $httpRequest ) use ( $app, $ffFactory ) {
...
-					$httpRequest->request->get( 'addressType', '' )
-				);
-
-				if ( $result->isSuccessful() ) {
-					return $app->json( [ 'status' => 'OK' ] );
-				} else {
-					$errors = $result->getViolations();
-					return $app->json( [ 'status' => 'ERR', 'messages' => $errors ] );
-				}
+				return ( new ValidateFeeController( $app, $httpRequest ) )->validateFee();
```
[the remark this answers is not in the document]
JeroenDeDauw: Interesting, thanks for the pointer.

What is the benefits of this approach? One downside is that it breaks "find usages" and automated refactoring. I suspect it is also not possible to use constructor parameters where appropriate, which is another downside.

app/Controllers/ValidateFeeController.php
JeroenDeDauw: Hah. That seems better even if you are not migrating away from Silex.

app/Routes.php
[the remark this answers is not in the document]
gbirke: This approach is a preparation for migrating to Symfony. The benefit (when all routes are converted to this format) is a short and concise configuration of which URLs are handled by which controllers. Then there is exactly one usage of the controller - in the route configuration file. I agree with you that having to manually touch that file on rename is less efficient than automatic refactoring. 

When we've switched to Symfony, we'll have constructor injection back through the DI container.

gbirke:
🍰

PR #1345, opened 2018-10-04
Form validation logging

timEulitz:
https://phabricator.wikimedia.org/T193054

### DIFF
--- a/app/Controllers/AddDonationController.php
+++ b/app/Controllers/AddDonationController.php
@@ -2,18 +2,17 @@
 
 declare( strict_types = 1 );
 
-namespace WMDE\Fundraising\Frontend\App\RouteHandlers;
+namespace WMDE\Fundraising\Frontend\App\Controllers;
 
 use Silex\Application;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use WMDE\Euro\Euro;
 use WMDE\Fundraising\DonationContext\Domain\Model\DonationTrackingInfo;
 use WMDE\Fundraising\DonationContext\UseCases\AddDonation\AddDonationRequest;
 use WMDE\Fundraising\DonationContext\UseCases\AddDonation\AddDonationResponse;
-use WMDE\Fundraising\Frontend\App\Controllers\ShowDonationConfirmationController;
 use WMDE\Fundraising\Frontend\BucketTesting\Logging\Events\DonationCreated;
-use WMDE\Fundraising\Frontend\App\Controllers\UpdateDonorController;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\Infrastructure\AmountParser;
 use WMDE\Fundraising\PaymentContext\Domain\Model\BankData;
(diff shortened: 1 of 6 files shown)


app/Controllers/AddDonationController.php
```diff
@@ -242,4 +238,22 @@ private function resetSessionState(): void {
 			false
 		);
 	}
+
+	/**
+	 * @param ConstraintViolation[] $constraintViolations
+	 */
+	private function logValidationErrors( $constraintViolations ): void {
+		$formattedConstraintViolations = [];
+		foreach ($constraintViolations as $constraintViolation) {
```
JeroenDeDauw: spacing is wrong. I think the CS stuff can be updated to catch this. IIRC, this was disabled since when we started it did not work with PHP 7.0

timEulitz: Thanks, I will look into enabling this for our CS.

timEulitz: The issue was actually that the directory is not included in the CS config.

app/Controllers/AddDonationController.php
```diff
@@ -242,4 +238,23 @@ private function resetSessionState(): void {
...
+			$formattedConstraintViolations[] = sprintf(
+				'Validation field "%s" of value "%s" failed with: %s',
+				$constraintViolation->getSource(),
+				$constraintViolation->getValue(),
+				$this->ffFactory->getTranslator()->trans( $constraintViolation->getMessageIdentifier() )
+			);
+		}
+		$this->ffFactory->getLogger()->warning(
+			'Unexpected server-side form validation errors.',
+			$formattedConstraintViolations
```
gbirke: The `context` parameter has to be a key => value array.

app/Controllers/ApplyForMembershipController.php
```diff
@@ -224,4 +226,19 @@ private function logSelectedBuckets( ApplyForMembershipResponse $responseModel )
...
+		foreach ( $validationResult->getViolations() as $constraintViolationSource => $constraintViolation ) {
+			$formattedConstraintViolations[] = sprintf(
+				'Validation field "%s" of value "%s"',
+				$this->ffFactory->getTranslator()->trans( $constraintViolationSource ),
+				$this->ffFactory->getTranslator()->trans( $constraintViolation )
+			);
+		}
+		$this->ffFactory->getLogger()->warning(
+			'Unexpected server-side form validation errors.',
+			$formattedConstraintViolations
```
gbirke: The `context` parameter has to be a key => value array.

app/Controllers/AddDonationController.php
```diff
@@ -242,4 +238,23 @@ private function resetSessionState(): void {
 			false
 		);
 	}
+
+	/**
+	 * @param ConstraintViolation[] $constraintViolations
+	 */
+	private function logValidationErrors( $constraintViolations ): void {
+		$formattedConstraintViolations = [];
+		foreach ( $constraintViolations as $constraintViolation ) {
```
gbirke: What, no `array_map`? 😛 😸

JeroenDeDauw: I was also thinking that. This way you avoid the $formattedConstraintViolations variable that gets mutated and have the string formatting not be in the same function as the call to the logger (different levels of abstraction). Decided to not comment on it since the current code is hardly bad.

gbirke:
Looks fine now. You can merge when it passes the CI.

PR #1355, opened 2018-10-08
Load FAQ content from wmde vendor

tzhelyazkova:
Feature: [T206320](https://phabricator.wikimedia.org/T206320)


skins/cat17/templates/page_layouts/faq.html.twig
```diff
@@ -3,9 +3,12 @@
 {% block page_identifier %}faq{% endblock%}
 
 {% block main %}
-	<div id="faq"></div>
+	<div id="faq" content="{$ faq_content $}"></div>
```
gbirke: Let's try to keep the HTML valid by using the `data-` prefix for custom attributes. See https://stackoverflow.com/questions/24932391/are-custom-html-attributes-without-data-prefix-a-valid-attribute

skins/cat17/templates/page_layouts/faq.html.twig
```diff
@@ -3,9 +3,12 @@
...
 {% endblock %}
 
 {% block scripts %}
+	<script type="text/javascript">
+		var faqContent = JSON.parse( document.getElementById( "faq" ).getAttribute( "content" ) );
```
gbirke: This creates a global variable that will trigger ESLint on `faq.js` I think it would be better to move the call there.

app/Routes.php
```diff
@@ -308,6 +308,15 @@ function () use ( $ffFactory ) {
 			}
 		)->bind( 'contact' );
 
+		$app->get(
+			'faq',
+			function () use ( $ffFactory ) {
+				return $ffFactory->getLayoutTemplate( 'page_layouts/faq.html.twig' )->render(
```
gbirke: Now that it's a standalone route we should move the template out of `page_layouts`.

skins/cat17/templates/page_layouts/faq.html.twig
tzhelyazkova: I actually named it `data-content` initially, and then on second thought I removed it because the words `data` and `content` sounded like synonyms in this context. 
Anyway, I didn't know it's considered invalid, so I'll change it. Thanks!

skins/cat17/templates/page_layouts/faq.html.twig
tzhelyazkova: but that's in the `web` folder :fearful:
Do you mean the `main.js` in the `faq-page` folder ?

skins/cat17/faq-page/src/Faq.vue
```diff
@@ -71,167 +71,25 @@ export default {
...
-							'Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris ' +
-							'nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in ' +
-							'reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. '
-						}
-					]
-				}
-			]
+			topicTitle: '',
+			content: faqContent,
+			messages: faqMessages,
```
gbirke: I think you can remove the two previous lines. `content` and `messages` are already properties, there is no need to put them in the state, right? Also, where do those variables come from?

skins/cat17/templates/faq.html.twig
```diff
@@ -0,0 +1,15 @@
...
+{% block page_identifier %}faq{% endblock%}
+
+{% block main %}
+	<div id="faq" data-content="{$ faq_content $}" data-messages="{$ faq_messages $}"></div>
+{% endblock %}
+
+{% block scripts %}
+	<script type="text/javascript">
+		var faqContent = JSON.parse( document.getElementById( "faq" ).getAttribute( "data-content" ) );
+		var faqMessages = JSON.parse( document.getElementById( "faq" ).getAttribute( "data-messages" ) );
```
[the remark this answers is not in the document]
tzhelyazkova: Yeah, about that, I tried but couldn't make it work so far.

skins/cat17/faq-page/src/Faq.vue
tzhelyazkova: > content and messages are already properties, there is no need to put them in the state, right?

true
> Also, where do those variables come from?

They come from `faq.html.twig`

skins/cat17/templates/faq.html.twig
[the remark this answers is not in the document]
gbirke: See #1356

skins/cat17/faq-page/src/Faq.vue
```diff
@@ -71,167 +71,25 @@ export default {
...
+			page: []
 		};
 	},
 	mounted: function () {
 		this.populatePageWithPreviewContent();
 	},
 	methods: {
 		populatePageByTopic: function ( topic ) {
-			this.page = topic;
+			this.page = this.content.questions.filter( ( question ) => question.topic.split().indexOf( topic.id ) !== -1 );
```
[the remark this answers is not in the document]
tzhelyazkova: I'm sorry but I disagree on this one :worried: 
The method is already clearly named and does only one thing -> populating the page with content by topic.
The anon function is not really `hasTopic`, as it is meant to filter by topic. Since `Array.filter` is already pretty clear what it does, I think this is readable enough. If we move this one line into a higher order function somewhere outside of the scope of the `Vue` element we're gonna create a small chaos IMHO. I really don't want to do that. If you insist on extracting this, I'd like it to be part of the `methods` object.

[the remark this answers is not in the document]
gbirke: Ok, it was more of a suggestion and readability is a subjective quality.

Don't forget to add the separator character, though.

tests/Unit/Infrastructure/JsonStringReaderTest.php
```diff
@@ -0,0 +1,52 @@
...
+	private function getFileFetcherWithContent( array $content ): InMemoryFileFetcher {
+		return new InMemoryFileFetcher( $content );
+	}
+
+	public function testGivenInvalidJsonFile_exceptionIsThrown() {
+		$this->expectException( \RuntimeException::class );
+		( new JsonStringReader(
+			'invalid.json',
+			'/test/directory/',
+			$this->getFileFetcherWithContent( [ '/test/directory/invalid.json' => '{"test": "is" "broken": "now "}' ] )
```
gbirke: This faulty JSON is very subtle. Maybe you can make the error a bit more obvious (e.g. leaving out quotes and last brace)

src/Infrastructure/JsonStringReader.php
```diff
@@ -0,0 +1,44 @@
...
+	private function isJsonEmpty(): bool {
+		return $this->json === '';
+	}
+
+	private function isJsonValid(): bool {
+		json_decode( $this->json );
+		return json_last_error() === JSON_ERROR_NONE;
+	}
+
+	public function validateJson(): string {
```
gbirke: This method name does not match the class name. Could you name it `readJson` or `readAndValidateJson`?

PR #1358, opened 2018-10-10
Fix tests for bank account validation

timEulitz:
Made this a separate PR into the feature branch in case I did something silly with the tests.


skins/cat17/src/app/tests/test_form_validation.js
```diff
@@ -296,28 +280,3 @@ test( 'Given non-sepa debit type, bank data validation sends account number and
 	} );
 	t.end();
 } );
-
-test( 'Given a non-debit payment type, bank data validation is not applicable', function ( t ) {
```
timEulitz: Deleted because this check dose not exist anymore.

skins/cat17/src/app/tests/test_form_components.js
```diff
@@ -315,10 +313,8 @@ test( 'Bank data component adds change handling function to its elements', funct
 
 	formComponents.createBankDataComponent( store, bankDataComponentConfig );
 
-	assertChangeHandlerWasSet( t, bankDataComponentConfig.ibanElement );
+	assertChangeHandlerWasSet( t, bankDataComponentConfig.ibanElement, 2 );
```
timEulitz: IBAN field has two change handlers now, I am not entirely sure if that's correct, I assume it's because of the two events we attached to it? Will look into this still.

timEulitz: I have tried looking into it but I am really not sure about this change, maybe you have an idea @gbirke @tzhelyazkova?

gbirke: The test cases for the bank data component can be deleted, as it's replaced with a Vue component

PR #1376, opened 2018-10-30
Use Piwik for tracking tooltip interaction

gbirke:
Hovering on a tooltips now tracks the tooltip content after 1 second.
Clicking on the tooltip also counts as an interaction.

Feature: https://phabricator.wikimedia.org/T207635

### DIFF
--- a/skins/10h16/templates/Donation_Form_Personal_Data.html.twig
+++ b/skins/10h16/templates/Donation_Form_Personal_Data.html.twig
@@ -270,9 +270,10 @@
                             </li>
 
                             <li>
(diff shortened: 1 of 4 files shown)


tzhelyazkova:
Tested locally and it works like a :gem: :+1:

JeroenDeDauw:
It works like a diamond? I suppose that does make sense for frontend stuff. It just sits there looking pretty...

tzhelyazkova:
> It works like a diamond? I suppose that does make sense for frontend stuff. It just sits there looking pretty...

No! It works like a gem!

PR #1385, opened 2018-11-13
Remove ImageminPlugin from Webpack

timEulitz:
ImageminPlugin compresses SVG files too much and removed the viewBox
attribute which breaks Internet Explorer 11 compatibility. Even after
manually configuring svgo by setting removeViewBox to false, the viewBox
is still removed. After countless attempts, I am removing this plugin
for now. See https://github.com/svg/svgo for more info on SVG
compression.

As part of this commit, I have manually compressed the SVG and retained
the data which we really need for browser compatibility. Files are now copied straight into the web directory, see line 49 in the diff:

`{ from: 'assets/images/*', context: 'src' }`

Other than that, this only appears to touch two more files actually, namely the favicon and apple-touch-icon.


timEulitz:
@gbirke This is related to this ticket, in case you come across this.

https://phabricator.wikimedia.org/T209209

timEulitz:
Just tested all browsers on Test, all good.

PR #1388, opened 2018-11-19
Prevent Sofort donations from being recurring

timEulitz:
T205297

### DIFF
--- a/skins/cat17/src/app/lib/reducers/form_content.js
+++ b/skins/cat17/src/app/lib/reducers/form_content.js
@@ -21,17 +21,22 @@ function trimValue( value ) {
 function forcePersonalDataForDirectDebit( state ) {
 	if ( state.paymentType === 'BEZ' && state.addressType === 'anonym' ) {
 		return objectAssign( {}, state, { addressType: 'person' } );
-	} else {
-		return state;
 	}
+	return state;
 }
 
 function forceAddressTypeForActiveMembership( state ) {
 	if ( state.membershipType === 'active' ) {
 		return objectAssign( {}, state, { addressType: 'person' } );
-	} else {
-		return state;
 	}
+	return state;
+}
+
+function forceOneTimePaymentForSofort( state ) {
+	if ( state.paymentType === 'SUB' && state.paymentIntervalInMonths > 0 ) {
+		return objectAssign( {}, state, { paymentIntervalInMonths: 0 } );
+	}
+	return state;
 }
 
 module.exports = {
@@ -68,6 +73,7 @@ module.exports = {
 
 				newState = forcePersonalDataForDirectDebit( newState );
 				newState = forceAddressTypeForActiveMembership( newState );
+				newState = forceOneTimePaymentForSofort( newState );
 				return newState;
 			case 'FINISH_BANK_DATA_VALIDATION':
 				if ( action.payload.status !== 'OK' ) {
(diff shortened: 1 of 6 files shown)


skins/cat17/src/scripts/donationForm.js
```diff
@@ -78,6 +78,15 @@ $( function () {
 			viewHandler: WMDE.View.createElementClassSwitcher( $( '#type-donor .wrap-field.anonym .wrap-input' ), /BEZ/, 'disabled' ),
 			stateKey: 'donationFormContent.paymentType'
 		},
+		// Disable recurring intervals when Sofort is selected
+		{
+			viewHandler: WMDE.View.createElementClassSwitcher( $( '#recurrence .wrap-field.recurring .wrap-input' ), /SUB/, 'disabled' ),
+			stateKey: 'donationFormContent.paymentType'
+		},
+		{
+			viewHandler: WMDE.View.createElementClassSwitcher( $( '#payment-method .wrap-field.sofort .wrap-input' ), /[^0]+/, 'disabled' ),
```
timEulitz: This whole part of the code is some seriously weird territory, but the "not 0" regex makes me look like I am playing chess while the rest of the world is still playing checkers 🎓:

`/[^0]+/`

It seems to work fine, though!

skins/cat17/src/app/lib/reducers/form_content.js
```diff
@@ -21,17 +21,22 @@ function trimValue( value ) {
...
 function forceAddressTypeForActiveMembership( state ) {
 	if ( state.membershipType === 'active' ) {
 		return objectAssign( {}, state, { addressType: 'person' } );
-	} else {
-		return state;
 	}
+	return state;
+}
+
+function forceOneTimePaymentForSofort( state ) {
```
tzhelyazkova: I know it's `Sofort Überweisung` but I think we agreed to refrain from German words in the code. 
I'd say `forceOneTimePaymentForSUB` goes more in that right direction.

skins/cat17/src/scripts/donationForm.js
tzhelyazkova: I think this is the shortest regexp that checks for `not 0`. Anything else will be less readable.

skins/cat17/src/app/tests/reducers/test_form_content.js
```diff
@@ -82,6 +82,22 @@ test( 'When CHANGE_CONTENT sets membership type to active, address type is switc
 	t.end();
 } );
 
+test( 'When CHANGE_CONTENT sets payment type to sofort and payment interval is at least monthly, interval is set to once', function ( t ) {
```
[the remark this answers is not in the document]
timEulitz: I agree, but the problem is that the whole directory full of tests uses this pattern and I think it's not bad enough to go through all files and fix them, I think a simple search and replace will also not do it because they describe different content changes. 

🤷‍♂️

skins/cat17/src/scripts/donationForm.js
gbirke: The regex is fine and the ElementClassSwitcher works as intended. 

The only "better" way to do this is converting all those ViewHandlers to Vue.

skins/cat17/src/app/lib/reducers/form_content.js
gbirke: Yes, "Sofort" is the name of the Company/Payment type.

PR #1390, opened 2018-11-20
Validate form upon clicking Submit

tzhelyazkova:
Feature: [T209920](https://phabricator.wikimedia.org/T209920)


gbirke:
Since there is no plan currently to use the page without banners, Kai gave the "go" to merge this. We will follow up the payment data behavior in another ticket.

PR #1392, opened 2018-11-21
FAQ layout changes

tzhelyazkova:
Feature: [T209943](https://phabricator.wikimedia.org/T209943)

Sorry for taking so long to do this :snail: :snail: :snail:


skins/cat17/faq-page/src/Faq.vue
```diff
@@ -12,14 +12,16 @@
  	<h2>{{ messages.page_title }}</h2>
  	<h5>{{ messages.page_subtitle }}</h5>
  	<div class="row">
- 	<div class="col-xs-12 col-sm-9">
+ 	<div class="col-xs-12 col-sm-8">
 	  	<search-bar :messageSearch="messages.search"></search-bar>
 		<div class="form-shadow-wrap">
-			<h2 v-if="!isPreview" class="title space-below underlined">{{ topicTitle }}</h2>
+			<h2 v-if="!isPreview" class="title">{{ topicTitle }}</h2>
```
gbirke: I remember talking about this - "Preview" means "Overview", right? Could you rename this to make it more clear?

skins/cat17/faq-page/src/Faq.vue
```diff
@@ -12,14 +12,16 @@
...
 			<ul>
-				<li v-for="content in page" v-bind:class="[ isPreview ? 'preview' : 'topic', 'underlined' ]">
-					<question :content="content.question"></question>
-					<answer :hiddenContent="content.hidden_text" :visibleContent="content.visible_text" :messages="messages" :isPreview="isPreview"></answer>
+				<li v-for="(content, index) in page" v-bind:class="[ isPreview ? 'preview' : 'topic' ]">
```
gbirke: I suggest removing the `ul/li`tags and rendering the `question` elements directly - except if the lists strongly help with accessibility or you really want to convey the semantics that this is a list of questions.

If you render the questions directly, you can use the the `:key` attribute, as proposed in https://vuejs.org/v2/guide/list.html You can then drop `:idx` and  use `:key` instead, getting the additional benefit of faster list rendering.

skins/cat17/faq-page/src/components/Question.vue
```diff
@@ -1,12 +1,46 @@
...
+		toggle: function () {
+			if ( !this.isOpen ) {
+				this.$emit( 'question-opened', this.idx );
+				this.collapse = false;
+			} else {
+				this.collapse = true;
+			}
+
+		}
+	}
```
gbirke: I fear this "double tracking" of opened state will be confusing to our future selves 😿 

I'd suggest getting rid of the `collapse` state, just emitting `question-opened` on click.

skins/cat17/faq-page/src/Faq.vue
```diff
@@ -12,14 +12,16 @@
...
+					<question v-on:question-opened="setOpenQuestionId( $event )"
+							  :content="content"
+							  :idx="index"
+							  :visible="openQuestionId"></question>
```
gbirke: This property name is a little bit confusing to me because it can be read as a boolean. Maybe call it 
`visibleQuestion` or `visibleQuestionId`?

skins/cat17/faq-page/src/components/Question.vue
tzhelyazkova: Agh yes, I had a temporal dumb moment with this `collapse` state. :woman_facepalming:

skins/cat17/faq-page/src/components/Question.vue
```diff
@@ -1,12 +1,44 @@
...
+		isOpen: function () {
+			return this.$vnode.key === this.visibleQuestionId;
+		}
+	},
+	methods: {
+		toggle: function () {
+			if ( !this.isOpen ) {
+				this.$emit( 'question-opened', this.$vnode.key );
+			} else {
+				this.$emit( 'question-opened' );
```
gbirke: Oooh, that is so clever! Emitting the event without id will close the question, because an empty id won't match the key in `isOpen`. Great way to do things. 

It took me a while to figure that out, though. Not sure if that makes it too clever and how to mitigate - passing an empty string or adding an inline comment perhaps? But not a blocker, you can also leave the code like it is.

skins/cat17/faq-page/src/components/Question.vue
```diff
@@ -1,12 +1,44 @@
...
+	</div>
 </div>
 </template>
 
 <script>
 export default {
 	name: 'question',
-	props: [ 'content' ]
+	props: [ 'content', 'visibleQuestionId' ],
+	data() {
```
gbirke: You can remove the data method now.

skins/cat17/faq-page/src/components/Question.vue
tzhelyazkova: I also thought about passing empty string because the prop the parent sends to the child is an empty string initially. But.. `key` is `Number` so we emit the same type of event with different type of data. I thought that's not super correct, but I do agree with the argument that it's not clear what's going on in the `else`. I'll add a comment for now and if in 6 months from now we find ourselves wondering what does this do, we can change it.

skins/cat17/src/sass/layouts/pages/faq.scss
```diff
@@ -48,22 +69,43 @@
...
+		display: flex;
+		flex-direction: row;
+		flex-wrap: nowrap;
+		justify-content: space-between;
+		align-items: center;
+	}
+	.clickable {
+		cursor: pointer;
+	}
+	.open {
```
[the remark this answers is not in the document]
tzhelyazkova: nowhere

skins/cat17/src/sass/layouts/pages/faq.scss
```diff
@@ -1,4 +1,25 @@
 .page-faq main {
+	.primary-color {
+		color: #7f7f7f;
```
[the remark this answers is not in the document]
tzhelyazkova: I will try my freakin best :woman_teacher:

skins/cat17/src/sass/layouts/pages/faq.scss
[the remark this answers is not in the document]
timEulitz: 18ps27

skins/cat17/src/sass/layouts/pages/faq.scss
```diff
@@ -1,4 +1,25 @@
...
+		border-bottom: 1px solid #ddd;
+		&:hover {
+			border-color: #000;
+		}
+	}
+	.title-primary-color {
+		color: inherit;
+	}
+	.secondary-color {
+		color: #00adee;
```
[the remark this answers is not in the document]
tzhelyazkova: I sort of followed the convention in the rest of the `scss` files across the repo. And there it's primary etc.

PR #1394, opened 2018-11-26
Change design of membership call to action

gbirke:
Change design of membership call to action on the confirmation page

* Use membership color
* Add colored box
* Add more text.

This is for https://phabricator.wikimedia.org/T210083

### DIFF
--- a/skins/cat17/src/sass/layouts/pages/donation_confirmation.scss
+++ b/skins/cat17/src/sass/layouts/pages/donation_confirmation.scss
@@ -128,9 +128,44 @@
 	}
 
 	.membership-benefits-box {
+		h2 {
+			border-bottom: solid 2px $brand-primary-membership;
+			display: inline-block;
+
+			/* Counteract styles from _layouts/memberships */
+			@include bkp(xsmax) {
+				&::after {
+					border-bottom: none;
+				}
+			}
+		}
+
 		.content {
 			padding-top: 0;
 		}
+
+		.call-to-action-box {
+			background: $brand-primary-membership;
+			color: $white;
+			border-radius: 5px;
+			padding: 15px;
+		}
+
+		.highlighted {
+			border: solid 1px $brand-primary-membership;
+		}
+
+		.list-membership {
+			li::before {
+				color: $brand-primary-membership;
+			}
+		}
+
+		.btn {
+			background: $brand-primary-membership;
+			margin-bottom: 25px;
+		}
+
 		@include bkp(md) {
 			.action-block {
 				margin: 0;
--- a/skins/cat17/templates/partials/donation_confirmation/membership_advantages.html.twig
+++ b/skins/cat17/templates/partials/donation_confirmation/membership_advantages.html.twig
@@ -1,12 +1,23 @@
 <div class="container hidden-print">
 	<div class="row">
-		<div class="content col-xs-12 col-md-8 content col-xs-12 col-md-8">
+		<div class="content highlighted col-xs-12 col-md-8">
+			<h2>{$ 'membership_call_to_action_title' | trans $}</h2>
+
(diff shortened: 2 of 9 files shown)


skins/cat17/templates/partials/donation_confirmation/membership_advantages.html.twig
```diff
@@ -1,12 +1,21 @@
...
+			</div>
+
 			{% include 'partials/membership/benefit_list.html.twig' with {
 				'showActiveMembershipBenefits': true
 			} only %}
-		</div>
-		<div class="action-block col-xs-12 col-sm-offset-0 col-sm-12 col-md-8">
-			<a id="membership-advantages-button" href="{$ basepath $}/apply-for-membership?donationId={$ donation.id $}&amp;donationAccessToken={$ donation.accessToken $}&amp;type=sustaining" class="btn btn-introduction bk-blue text-center text-uppercase">
+
+			<a href="{$ basepath $}/apply-for-membership?donationId={$ donation.id $}&amp;donationAccessToken={$ donation.accessToken $}&amp;type=sustaining"
```
timEulitz: `/apply-for-membership?donationId={$ donation.id $}&amp;donationAccessToken={$ donation.accessToken $}&amp;type=sustaining`

The two cases of `&amp;` in the URL look suspicious to me, I think this should be `&`.

skins/cat17/src/sass/layouts/pages/donation_confirmation.scss
```diff
@@ -128,9 +128,42 @@
...
 
 	.membership-benefits-box {
+		h2 {
+			border-bottom: solid 2px $brand-primary-membership;
+			display: inline-block;
+
+			/* Counteract styles from _layouts/memberships */
+			@include bkp(xsmax) {
+				&::after {
+					border-bottom:none;
```
timEulitz: Ultra nitpick, but there should be a space after the colon character, doesn't our code style toolset check for this?

skins/cat17/src/sass/layouts/pages/donation_confirmation.scss
```diff
@@ -128,9 +128,42 @@
...
+			color: $white;
+			border-radius: 5px;
+			padding: 15px;
+		}
+
+		.highlighted {
+			border: solid 1px $brand-primary-membership;
+		}
+
+		.list-membership li::before {
```
timEulitz: Don't need to change this but quickly wanted to comment: I would personally split this into three blocks but I think we have no agreement for this and I am not sure what would be considered "best practice", I can see arguments for either way.

```
.list-membership { 
    li {
        &::before {
            color: $brand-primary-membership;
    }
}
```

gbirke: I prefer the "only nest when there are actual sub-elements" approach because it's closer to the original css notation. There is no clear rule for our frontend code.

&lt;rant&gt;
I had hoped that the original designers of the frontend would use a systematic, component/styleguide-based approach to css (like Vue components or [BEM](http://getbem.com/introduction/)). Instead, they created seemingly random css snippets, nested too deep. We did not clean up their CSS mess because we only had time for the JS mess. Instead, we continue
&/lt;rant&gt;

Let's discuss our CSS code style in the next in-person daily ...

gbirke:
Thanks for the Screenshot, Tonina. I saw this yesterday, showed it to Kai and he said that having the right-aligned menu is ok for the moment.

I've experimentally removed the column classes for `sm` resolution, but then discovered that the lines between the links are absolutely-positioned, fixed-width `::after` elements, 😝 😨 😠 
I don't want to break other list menus and I don't want to pile on more overriding CSS, so I'd say until someone complains we leave it ugly like that.

PR #1403, opened 2018-12-10
Provide different amount options on donation form in 10h16

tzhelyazkova:
Feature: [T211545](https://phabricator.wikimedia.org/T211545)


src/Factories/ChoiceFactory.php
```diff
@@ -52,6 +52,21 @@ public function getMembershipCallToActionTemplate(): string {
...
+		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.amount_options.5to100' ) ) {
+			return '5to100';
+		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.amount_options.15to250' ) ) {
+			return '15to250';
+		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.amount_options.30to250' ) ) {
+			return '30to250';
+		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.amount_options.50to500' ) ) {
+			return '50to500';
+		}
+		throw new UnknownChoiceDefinition( 'Amount option selection configuration failure.' );
```
timEulitz: To prevent any more :baby::baby_bottle::baby: major boo-boos :baby::baby_bottle::baby:, I think we should return some default fallback value from the list above here. Probably also change it for all other functions in this file to be on the safe side.

tzhelyazkova: Since this A/B test is for the email campaign we don't have to configure `start date` in two different places => throwing an exception here is ok.

PR #1406, opened 2018-12-11
Improve testing for BucketTest feature

gbirke:
https://phabricator.wikimedia.org/T211716


gbirke:
Setting to "DNM" because the campaign end dates are at times where the FUN team probably won't work. When the campaign is over, we can merge this.

PR #1408, opened 2018-12-16
Also run tests PHP 7.3

JeroenDeDauw:


JeroenDeDauw:
This is probably wrong since the stuff is run in Docker. Not seeing where the version for that is defined though

gbirke:
Yes, we build the docker images (with the kontocheck extension) on every CI run. I've already created ticket https://phabricator.wikimedia.org/T206807 to change that. When that ticket is done, we need to change the travis file to allow different build matrices for docker images with different PHP versions.

I'm very sorry for closing this for now, but I've created a task for implementing this properly: https://phabricator.wikimedia.org/T212220

PR #1419, opened 2019-01-03
Conclude form design change AB test

timEulitz:
https://phabricator.wikimedia.org/T212859


skins/cat17/templates/partials/Page_Header.html.twig
```diff
@@ -20,13 +20,26 @@
 					</div>
 					<nav class="menu menu-main">
 						<ul class="list-unstyled list-inline">
-							{% for menu_item in main_menu %}
```
gbirke: I rather liked the menu array constructed in PHP and rendered in the template as a loop. IMHO it made the template much cleaner added uniformity to the markup. 
You could make the argument that the menu is skin-specific and does not belong in the PHP code. In that case I'd be fine with just reverting the template, as having "skin-specific template data" would be overkill. 

If you revert the template, at least use the `path` function for the URL (with the route name) instead of hard-coding them.

timEulitz: As discussed, since we don't really have a good place to place the `main_menu` items right now, we're leaving the template as is (with the addition of the `path` function).

PR #1420, opened 2019-01-07
Add reference to contact form on missing 10h16 FAQ

timEulitz:
https://phabricator.wikimedia.org/T212140

### DIFF
--- a/skins/10h16/templates/faq.html.twig
+++ b/skins/10h16/templates/faq.html.twig
@@ -1,0 +1,14 @@
+{% extends 'Base_Layout.html.twig' %}
+
+{% block main %}
+	<div>
+		<p>
+			Diese Seite steht im Moment nicht zur Verfügung.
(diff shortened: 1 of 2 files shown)


skins/10h16/templates/faq.html.twig
```diff
@@ -0,0 +1,14 @@
...
+
+{% block main %}
+	<div>
+		<p>
+			Diese Seite steht im Moment nicht zur Verfügung.
+			Wenn Sie eine Frage haben, nutzen Sie bitte unser <a href="{$ path( 'contact' ) $}">Kontaktformular</a>.
+		</p>
+
+		<p class="icon-hand-right">
+			<a href="{$ path( '/' ) $}" target="_blank">{$ 'back_to_donation' | trans $}</a>
```
gbirke: I know we have `target="_blank"` all over the place, but we use it mostly to ensure people are not losing their inputs on the donation page. In this case i'd say it's better to not have it.

timEulitz: Copa-pasta mistake :spaghetti:.

PR #1422, opened 2019-01-09
Address Form Frontend

tzhelyazkova:
This PR became a :japanese_ogre: monster.
The reviewer can start at commit `Create Address Form using VueCLI` because everything before that is just attempts of gluing `Vue`, `TypeScript` and `FF`.
*--------------------------------------------------------*
Feature: [T213407](https://phabricator.wikimedia.org/T213407)


skins/cat17/src/pages/AddressForm.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+	<span>Hello World!</span>
```
JeroenDeDauw: image

skins/cat17/address-form/src/store.ts
```diff
@@ -0,0 +1,141 @@
...
+		},
+		validateInput({commit}, contentName, newValue, pattern, optionalField) {
+			return commit('VALIDATE_INPUT', {
+				contentName: contentName,
+				value: newValue,
+				pattern: pattern,
+				optionalField: optionalField || false
+			});
+		},
+		markEmptyFieldsInvalid({commit}, requiredFields, neutralFields) {
```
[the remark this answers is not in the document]
tzhelyazkova: Yes, the [vuex docs](https://vuex.vuejs.org/guide/mutations.html) mention that mutations must be synchronous because otherwise it's virtually impossible to track the state mutation.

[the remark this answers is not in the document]
gbirke: In the past, I've always mixed and matched mutations and actions, using actions only when i needed the async behavior. But now I think always using actions is more consistent and future-proof, at the small cost of "duplicating" synchronous mutations.

skins/cat17/address-form/src/store.ts
```diff
@@ -0,0 +1,138 @@
...
+			state.isValidating = false;
+		}
+	},
+	getters: {
+		validity(state, contentName: string) {
+            return state.form[contentName].isValid;
+		}
+	},
+	actions: {
+        storeAddressFields({ commit }, payload: Payload) {
```
gbirke: I imagined this more as `storeAddressField` with the payload of a single field. Is there any advantage to always sending all the field values from vue when one field value changes?

skins/cat17/address-form/src/store.ts
```diff
@@ -0,0 +1,138 @@
...
+            commit('STORE_VALUES', payload.formData);
+            commit('VALIDATE_INPUT', payload.formData);
+            commit('MARK_EMPTY_FIELD_INVALID', payload.formData);
+            if (Object.keys(this.state.form).filter(field => this.state.form[field].isValid === Validity.INVALID ).length === 0 ) {
```
gbirke: The purpose if this line of code is to start the validation when all the required fields are filled, right? I think the logic is more complex than what's here and should go into a getter called `allAddressFieldsAreFilled`.

That getter should have an internal list of required fields for person/company address type and return true/false if all the required field contain a non-empty `value`. The `isValid` property is not relevant here IMHO.

Also, this part of the code is prone to concurrency errors at the moment: when multiple `storeAddress` actions come in, multiple `BEGIN_ADDRESS_VALIDATION` and `END_ADDRESS_VALIDATION` will be triggered and might resolve in the wrong order. There are several ways on how to solve this:
* Check `isValidating` property and cancel all running requests
* Keep track of the running requests with an internal counter (incremented and decremented by `BEGIN_ADDRESS_VALIDATION` and `END_ADDRESS_VALIDATION`) and make `isValidating` a getter.

skins/cat17/address-form/src/components/Name.vue
```diff
@@ -0,0 +1,44 @@
...
+    </div>
+    <div v-else>
+        <div class="field-grp field-firstname col-xs-12 col-md-6">
+            <label for="first-name">{{ messages.firstname_label }}</label>
+            <input type="text" 
+                    id="first-name" 
+                    name="firstName" 
+                    :placeholder="messages.firstname_label" 
+                    data-pattern="^.+$"
+                    @blur="validateInput('firstName')">
```
gbirke: the `data-pattern` was an attribute that catered more to the jQuery-based code. With Vue and Vuex you can use the validation pattern as a parameter to `validateInput`.

skins/cat17/address-form/src/store.ts
```diff
@@ -0,0 +1,127 @@
...
+const MARK_EMPTY_FIELD_INVALID: string = 'MARK_EMPTY_FIELD_INVALID';
+const BEGIN_ADDRESS_VALIDATION: string = 'BEGIN_ADDRESS_VALIDATION';
+const FINISH_ADDRESS_VALIDATION: string = 'FINISH_ADDRESS_VALIDATION';
+const ADDRESS_FIELD_VALIDATOR_NAMES: Array<string> = [
+	'addressType',
+	'salutation', 'title', 'firstName', 'lastName',
+	'companyName',
+	'street', 'postcode', 'city', 'country'
+]
+let transport = new JQueryTransport();
```
gbirke: This will make the store hard to test, because now it depends on a networking component. Please inject a  TransportInterface in the constructor instead and initialize the transport and store in `main.js`

skins/cat17/address-form/src/store.ts
```diff
@@ -0,0 +1,127 @@
...
+
+Vue.use(Vuex);
```
gbirke: Maybe put this line and the last line in this file into `main.ts`? That way you can test this store component more independently.

tzhelyazkova:
Closing this because all the relevant commit history was transferred to  #1432 and properly re-based and merged to `master`.

PR #1423, opened 2019-01-09
Fix DateTimeInterface-related crash on production

timEulitz:
Production servers are still running on PHP 7.1, however the
DateTimeInterface only received class constants in PHP 7.2 causing the
code to throw a fatal error on production as the class constant cannot
be found.

### DIFF
--- a/src/BucketTesting/CampaignDate.php
+++ b/src/BucketTesting/CampaignDate.php
@@ -4,7 +4,6 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\BucketTesting;
 
-use DateTimeInterface;
 use DateTimeZone;
 
 /**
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Waaaat!!!! No PHP 7.3 yet? :(

gbirke:
@JeroenDeDauw We kind of missed the opportunity to update to PHP 7.2 last year (suddenly the campaign started 😝 ). But we're planning to update until March when Ubuntu 14.04 is EOL

As for PHP 7.3, we're waiting for PHP 7.3.1

timEulitz:
@gbirke PHP 7.3.1 was actually released just a few hours ago :fire::

http://www.php.net/ChangeLog-7.php#7.3.1

Xdebug 7.3.0 is also rather important though and that's still in beta with some bugs.

PR #1426, opened 2019-01-24
Validate address backend update

timEulitz:
Back-end changes for `validate-address` branch. 

CI failures should only be the remaining ones from `validate-address` branch.


tests/EdgeToEdge/Routes/ValidateAddressRouteTest.php
```diff
@@ -48,6 +48,27 @@ public function testGivenInvalidCompanyAddress_validationReturnsErrorMessage():
 		$this->assertJsonSuccessResponse( $expectedResponse, $response );
 	}
 
+	public function testGivenAnonymousAddress_validationReturnsErrorMessage(): void {
```
gbirke: Shouldn't that be `testGivenAnonymousAddress_validationReturnsSuccess` (with matching code)? Why should the validator give an error on anon? Although the behavior is technically correct when validating membership applications and address changes. We might add a parameter for that.

timEulitz: Yeah, true, I accidentally moved the logic from the old [UpdateDonorValidator](
https://github.com/wmde/fundraising-donations/blob/5e9b4ec264fca9aae5e0d1b7d5dd095e799074ac/src/UseCases/UpdateDonor/UpdateDonorValidator.php) here even though we should differentiate between these cases.

PR #1437, opened 2019-02-04
Use types for validity constants

gbirke:
Define enum for validity and object for server-side validation result
codes.
Using constant enum will speed up the compiled end result.


skins/cat17/address-form/src/store.ts
```diff
@@ -93,12 +92,12 @@ const store: StoreOptions<AddressState> = {
...
 		},
 		[FINISH_ADDRESS_VALIDATION](state, payload) {
-            if (payload.status === 'ERR') {
+            if (payload.status === ValidationStates.ERR) {
                 ADDRESS_FIELD_VALIDATOR_NAMES.forEach(name => {
                     if (state.form[name].dataEntered === true) {
                         state.form[name] = {
                             ...state.form[name],
-                            isValid: state.form[name].isValid !== false && !payload.messages[name] ? Validity.VALID : Validity.INVALID
+                            isValid: state.form[name].isValid = !payload.messages[name] ? Validity.VALID : Validity.INVALID
```
tzhelyazkova: this doesn't seem right. Did you mean:
`isValid:  !payload.messages[name] ?  Validity.VALID : Validity.INVALID`
which will be better off as:
`isValid:  payload.messages[name] ?  Validity.INVALID : Validity.VALID`
or is `=` supposed to be `===` ?

skins/cat17/address-form/src/types.ts
```diff
@@ -1,6 +1,18 @@
+export const enum Validity {
+    INVALID,
+    VALID,
+    INCOMPLETE
```
tzhelyazkova: As far as I understand that means INCOMPLETE is assigned with the value of 2. I think this is kind of strange but it does work for our use case.
Before, we had this mean `null` which is evaluated to `false`, while `2` is `true`. But again, it works, so no objections.

gbirke:
Implemented during #1438

PR #1439, opened 2019-02-05
Fix input fields model

tzhelyazkova:
`formData` model should be an object, not an array, because it makes the component more verbose and is easier to access values.


gbirke:
I like this improvement, it makes the code much more readable.

PR #1440, opened 2019-02-05
Prepare Address Form for Unit Tests

tzhelyazkova:
Define `JQueryTransport` interface and pass the library to the main `Vue` component.
This way we can easily test the behavior by passing a fake `JQueryTransport` in the unit tests.

### DIFF
--- a/skins/cat17/address-form/src/AddressForm.vue
+++ b/skins/cat17/address-form/src/AddressForm.vue
@@ -23,7 +23,7 @@
 	import Vue from 'vue';
 	import Name from './components/Name.vue';
 	import Postal from './components/Postal.vue';
-	import {inputField, ValidationResult} from './types';
+	import {InputField, ValidationResult} from './types';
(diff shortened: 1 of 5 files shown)


skins/cat17/address-form/src/types.ts
```diff
@@ -41,4 +42,7 @@ export interface PostData {
     [key: string]: string
 }
 
-
+export interface JQueryTransport {
```
gbirke: Let's lose the `JQuery` here. See the original code where there was a `Transport` interface (although only defined through JSDoc).

skins/cat17/address-form/src/AddressForm.vue
```diff
@@ -107,6 +107,7 @@
 			}
 		},
 		props: {
+            transport: Object,
```
gbirke: Should be a `Transport` interface.

PR #1443, opened 2019-02-07
Frontend ADRs

gbirke:


doc/adr/006_Vue.js.md
```diff
@@ -0,0 +1,37 @@
...
+
+We will write all new features in Vue and ECMAScript 2015 (and later), refactoring and cleaning up the existing code base. The redux "reducers" and "state aggregators" will move to Vuex modules, while the Twig templates, "view handlers" and "components" of the old code base become Vue components. 
+
+
+## Consequences
+
+* We are able to develop new features and A/B tests faster
+* We can re-use individual components and page elements easier 
+* The onboarding experience of new developers improves
+
```
tzhelyazkova: Another argument in my opinion:
We are taking one step towards standardizing our stack

doc/adr/006_Vue.js.md
```diff
@@ -0,0 +1,37 @@
...
+## Decision
+
+Going forward, we will use Vue to render the frontend. We chose it for the following reasons:
+
+* It's a mature, tested, well-documented widely used framework with an open source license and an active ecosystem
+* There is already some knowledge about Vue in the developer and UX teams
+* Vue is used in wikidata, making knowledge sharing easier. 
+
+We will write all new features in Vue and ECMAScript 2015 (and later), refactoring and cleaning up the existing code base. The redux "reducers" and "state aggregators" will move to Vuex modules, while the Twig templates, "view handlers" and "components" of the old code base become Vue components.
```
tzhelyazkova: I thought we were not that sure if we should replace the `Twig` templates with `Vue` ones but yay! \o/

doc/adr/007_Typescript.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Context
+
+During the modernization of the client-side code outlined in "[ADR 006 - Using Vue.js](006_Vue.js.md)" and the first prototype for that project - the Address Change Form - we also tried out using [TypeScript](https://www.typescriptlang.org/). After the project we evaluated if we want to continue to use TypeScript for client side code. We collected following arguments. 
+
+### Learning Curve
+Nobody of us has experience with TypeScript, there will be a learning curve and possible slowdowns while setting up features.
+
+### Migration Pain
+Migrating the existing JavaScript code looks daunting, we don't know if TypeScript will make the migration harder by having type requirements. However, since we're already migrating the code to Vue and Vuex, adding TypesScript into the mix will probably one be a minor hurdle.
```
tzhelyazkova: My opinion is the opposite. We are going to move markup and logic to Vue but also move all store logic to Vuex. That's already quite the commitment, given we have many `.js` and `.twig` files.  Adding TypeScript to the loop adds more complexity to the task and I wouldn't consider it "minor hurdle". But since we already established this is going to be a rewrite and not refactoring we should come out of it with something fully done, which won't need any more iterations of improvements in the near future. Hence we do all the things (vue, vuex, typescript).

^ That doesn't sounds very ADRy :smile:  I'm sure you can formulate it better in case you agree to change this part.

doc/Common_Practices_Typescript.md
```diff
@@ -0,0 +1,8 @@
+# Common practices for client-side code
+
+These guidelines are agreed-upon practices by the FUN team and reflect their current understanding of Vue.js and TypeScript. 
+
+* We collect all new types in the `types.ts` file. When it becomes sufficiently large, we'll have enough code "examples" to refactor them into smaller cohesive units/modules.
+* When writing a Vuex store module, don't split mutations, actions, getters, etc in separate files. Doing this makes the code harder to navigate. If a Vuex module file becomes too big, ask the team members if splitting is ok.
```
wiese: We did that split. Beyond smaller files it also gives you to ability to import only the actionTypes for example in order to perform an action without having to load all of the other store stuff along with it.
Examples:
* https://gerrit.wikimedia.org/r/plugins/gitiles/wikibase/termbox/+/master/src/store/language/ (w/ store modules)
* https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/WikibaseLexeme/+/master/resources/store/ (a monostore)

gbirke: We took inspiration from your code and did not like the "split style". 

At the moment we don't have actionTypes (using plain strings), but as soon as we do, splitting them into separate files makes total sense.

JeroenDeDauw:
Yay TypeScript! Have some kittens

image

doc/adr/006_Vue.js.md
timEulitz: As far as I understand this, `Twig` still remains our template engine, we just move twig files into components from now. But maybe @gbirke can clarify.

gbirke: The full answer to this question is long, as there are many implications and considerations. See Section "Implications for moving from Server-Side to Client-Side Rendering" ff of our [JavaScript Epic](https://phabricator.wikimedia.org/T205707) Please read it and give me feedback.

Short answer: IMHO, the strategy with the least engineering effort is to keep Twig for most pages for now. The partial templates for the donation and membership pages will become `template` sections in Vue components and the page templates will become empty `<div>` elements with data attributes. This is what I meant by the sentence above.

doc/adr/007_Typescript.md
gbirke: I've changed the wording. Is it what you wanted to say?

tzhelyazkova: yep, thanks!

PR #1444, opened 2019-02-11
Address Form unit tests

tzhelyazkova:

### DIFF
--- a/skins/cat17/address-form/src/__tests__/components/AddressForm.spec.ts
+++ b/skins/cat17/address-form/src/__tests__/components/AddressForm.spec.ts
@@ -1,7 +1,7 @@
 import { shallowMount, mount, createLocalVue } from '@vue/test-utils'
 import Vuex from 'vuex'
 import AddressForm from '../../AddressForm.vue'
-import { Validity } from '../../lib/validation_states'
+import { Validity } from "@/types";
 
 const localVue = createLocalVue();
 
@@ -34,48 +34,18 @@ describe('AddressForm.vue', () => {
         state = {
             isValidating: false,
             form: {
-                salutation: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                title: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                firstName: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                lastName: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                companyName: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                street: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                postcode: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                city: {
-                    dataEntered: false,
-                    isValid: Validity.INCOMPLETE
-                },
-                country: {
-                    dataEntered: false,
-                    isValid: Validity.VALID
-                },
-                addressType: {
-                    dataEntered: false,
-                    isValid: Validity.VALID
-                }
+                salutation: Validity.INCOMPLETE,
+                title: Validity.INCOMPLETE,
+                firstName: Validity.INCOMPLETE,
+                lastName: Validity.INCOMPLETE,
+                companyName: Validity.INCOMPLETE,
+                street: Validity.INCOMPLETE,
+                postcode: Validity.INCOMPLETE,
+                city: Validity.INCOMPLETE,
+                country: Validity.VALID,
+                addressType: Validity.VALID
             }
-        }
+        },
         actions = {
             validateInput: jest.fn(),
             storeAddressFields: jest.fn()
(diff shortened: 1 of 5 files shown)


skins/cat17/address-form/src/__tests__/store.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+	it('returns a list of the invalid fields', () => {
+		const state = {
+			form: {
+				title: Validity.INVALID,
+				street: Validity.INVALID,
+				city: Validity.INVALID
+			}
+		}
+		expect( getters.invalidFields( newMinimalStore( state ), null, null, null ) ).toStrictEqual( ['title', 'street', 'city'] );
```
gbirke: Could you add one or more line breaks here?

skins/cat17/address-form/src/__tests__/store.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+				addressType: Validity.VALID
+			}
+		},
+		overrides
+	);
+}
+
+describe('Store', () => {
+
+	it('returns no invalid fields on render', () => {
```
gbirke: The phrase "on render" is confusing to me, since the store itself does not render anything, we're only using the getters in the render part of the Vue component, but that is independent from the store.

Maybe something like "it returns no invalid fields" or "it returns no invalid field names when internal state does not contain them".

You could also add another level of `describe` for each setter.

skins/cat17/address-form/src/store.ts
```diff
@@ -16,78 +16,85 @@ const ADDRESS_FIELD_VALIDATOR_NAMES: Array<string> = [
...
+		return Promise.resolve();
 	}
 }
 
+const store: StoreOptions<AddressState> = {
+	state,
+	mutations,
+	getters,
+	actions
+}
```
gbirke: I like the split into individual constants, that gets rid of some of the nesting and can make things more visible and navigatable in the IDE.

skins/cat17/address-form/src/__tests__/store.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+		expect(getters.invalidFields( newMinimalStore( {} ), null, null, null ) ).toStrictEqual( [] );
```
gbirke: You probably can't get rid of those `null` values, right?

tzhelyazkova: Nope, because Vuex's `Getter` type expects `(state: S, getters: any, rootState: R, rootGetters: any)`

skins/cat17/address-form/src/__tests__/store.spec.ts
tzhelyazkova: Right, "on render" is incorrect indeed.
> You could also add another level of `describe` for each setter.

Yep will do.

skins/cat17/address-form/src/__tests__/components/AddressForm.spec.ts
```diff
@@ -1,71 +1,321 @@
...
+		firstName.setValue('Testina');
+		let lastName = wrapper.find('#last-name');
+		lastName.setValue('von Testinson');
+
+		expect(wrapper.vm.$data.formData.title.value).toBe('Prof. Dr.');
+		expect(wrapper.vm.$data.formData.firstName.value).toBe('Testina');
+		expect(wrapper.vm.$data.formData.lastName.value).toBe('von Testinson');
+	});
+
+	it('sets hidden field addressType "firma" for company form', () => {
```
gbirke: I'm sorry that we have to perpetuate the German identifiers here, for the sake of the server-side code. I've created a ticket to translate the identifiers, but it's a major effort for very little gain, so I fear this improvement might never happen: https://phabricator.wikimedia.org/T221044

skins/cat17/address-form/src/__tests__/store.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+			}
+		},
+		overrides
+	);
+}
+
+describe( 'Store', () => {
+
+	describe( 'Getters/invalidFields', () => {
+		it( 'returns no invalid fields on initialization', () => {
```
gbirke: I 💜 the new descriptions.

skins/cat17/address-form/src/__tests__/store.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+			const commit = jest.fn( function() { return Promise.resolve() } );
+			const action = actions[ 'storeAddressFields' ] as any;
+			action( { commit, getters }, payload );
+			expect( commit ).toBeCalledWith(
+				'MARK_EMPTY_FIELD_INVALID',
+				fakeFormData
+			);
+		} );
+
+		it( 'does not start address validation if there are invalid fields', () => {
```
gbirke: I know that incomplete fields are considered invalid inside this check, so technically this is correct, but reading it without that knowledge made me wonder why address validation is not started when there are invalid fields. And how fields can ever become valid when the validation is never started once they become invalid.

tzhelyazkova: Good that you mentioned it because while changing the description I noticed the next test has the exact same setup but the two assert different things. So something's wrong..

tzhelyazkova: Fixed ⚒

PR #1446, opened 2019-03-19
Update PHP and JavaScript dependencies

gbirke:


timEulitz:
@tzhelyazkova That's a deprecation warning that was apparently introduced with Composer 1.8. If you don't get this warning on your local, check your composer version via `composer --version` and then possibly upgrade it via `composer self-update`. 

I have just pushed a new composer.json and composer.lock file made with Composer 1.8.4 which should hopefully fix Scrutinizer. Either way, it's probably good practice that we all try to make sure to stay on the same release of Composer and maybe even make it part of our maintenance updates (even though it's technically not an infrastructure-related update).

gbirke:
>  Either way, it's probably good practice that we all try to make sure to stay on the same release of Composer and maybe even make it part of our maintenance updates (even though it's technically not an infrastructure-related update).

I did not realize that my local composer image was at 1.7.3. What do you think about pinning the composer and node versions in the Makefile (at least to a minor version)? That way, we can make sure our environment (both for npm and composer and the resulting lock files) is the same for all devs? It does introduce another step for the infra update.
Or should we keep it informal, i.e. everyone runs `docker pull composer` etc once a month?

timEulitz:
@gbirke Sure, if we can pin the version of Composer like that, I think it would be a good idea.

The actual issue that Composer has on Scrutinizer is this one:

>ocramius/proxy-manager 2.3.1 requires php ^7.4.0

I personally think that's kind of a weird move by the developer, but it seems to be on purpose:

https://github.com/Ocramius/ProxyManager/issues/454

The fix would be to lock that package to version 2.2.

PR #1450, opened 2019-03-22
Use special container for composer

gbirke:
Instead of using the official composer image, create an app image with
composer.

This makes sure we always have a matching PHP version and pin the
composer version.

Fix package name in composer.json

Update PHP dependencies.

This is for https://phabricator.wikimedia.org/T218861


tzhelyazkova:
ping @gbirke Travis complains

PR #1451, opened 2019-03-27
fix donation amount validation in UI

moiikana:
The donation amount will now be checked by newPaymentDataValidator()
so the donation maxAmount in the UI will be validated correctly and the same
way as in the backend.

https://phabricator.wikimedia.org/T210162

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ install-js:
 	docker run --rm $(DOCKER_FLAGS) --user $(current_user):$(current_group) -v $(BUILD_DIR):/data:delegated -w /data -v $(TMPDIR)/home:/var/nodehome:delegated -v $(TMPDIR)/passwd:/etc/passwd node:8 npm install $(NPM_FLAGS)
 
 install-php: build_app_composer
-	docker run --rm $(DOCKER_FLAGS) --volume $(BUILD_DIR):/app -w /app --volume ~/.composer:/composer --user $(current_user):$(current_group) wmde/fundraising-frontend-composer composer install $(COMPOSER_FLAGS)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -40,8 +40,8 @@ update-js:
 	docker run --rm $(DOCKER_FLAGS) --user $(current_user):$(current_group) -v $(BUILD_DIR)/skins/cat17:/data:delegated -w /data -v $(TMPDIR)/home:/var/nodehome:delegated -v $(TMPDIR)/passwd:/etc/passwd node:8 npm update $(NPM_FLAGS)
 	docker run --rm $(DOCKER_FLAGS) --user $(current_user):$(current_group) -v $(BUILD_DIR)/skins/10h16:/data:delegated -w /data -v $(TMPDIR)/home:/var/nodehome:delegated -v $(TMPDIR)/passwd:/etc/passwd node:8 npm update $(NPM_FLAGS)
 
-update-php: build_app_composer
```
gbirke: Oops, looks like the merge conflict resolution did not work. Please only add the `tmp` volume and leave the rest as is.

app/Routes.php
```diff
@@ -66,20 +66,15 @@ public static function initializeRoutes( Application $app, FunFunFactory $ffFact
...
-						'allowExtraFields' => false,
-						'fields' => [
-							'amount' => $ffFactory->newDonationAmountConstraint()
-						]
-					]
-				);
+				$rawAmount = $httpRequest->request->get('amount','');
 
-				$violations = Validation::createValidator()->validate( $httpRequest->request->all(), $constraint );
+				if(!ctype_digit($rawAmount)){
```
gbirke: This might trigger the code style check - we usually put spaces after opening brackets.

PR #1460, opened 2019-04-03
Set up wmde19 skin with dev server

gbirke:
Make wmde19 a usable skin in the Fundraising App and demonstrate how Twig templates, Vue components and JavaScript entry points can fit together.


skins/wmde19/templates/Base_Layout.html.twig
```diff
@@ -0,0 +1,42 @@
...
+
+    <meta name="format-detection" content="telephone=no" />
+    <meta name="title" content="{$ block('title') $}" />
+    <meta name="description" content="Wenn jeder nur einen kleinen Beitrag leisten würde, wäre unsere Spendenkampagne in einer Stunde vorüber." />
+
+    <meta property="og:title" content="Wenn jeder nur einen kleinen Beitrag leisten würde, wäre unsere Spendenkampagne in einer Stunde vorüber." />
+    <meta property="og:type" content="non_profit" />
+    <meta property="og:url" content="https://spenden.wikimedia.de/" />
+    <meta property="og:image" content="https://upload.wikimedia.org/wikipedia/commons/thumb/1/13/Wikipedia_svg_logo-de.svg/200px-Wikipedia_svg_logo-de.svg.png" />
+    <meta property="og:site_name" content="Gemeinnnützige Wikimedia Fördergesellschaft mbH" />
```
timEulitz: >Gemeinnnützige

Should be

>Gemeinnützige

README.md
```diff
@@ -298,6 +298,12 @@ To do the same for the 10h16 skin, run the following Docker command correspondin
 
 If you want to debug problems in the Redux data flow, add the parameter `-e REDUX_LOG=on` to the command line above
 
+### Automatic recompilation of assets during development (wmde19)
+
+If you are working on the JavaScript files of the wmde19 skin and need automatic recompilation when a files changes, add the following setting to your `config.dev.json` file:
+
+	"assets-path": "http://localhostz:7072"
```
timEulitz: >localhostz
Probably a typo.

docker-compose.yml
```diff
@@ -61,6 +62,18 @@ services:
       - database
     command: database:3306
 
+  skin_server_wmde19:
```
timEulitz: Why does this need its own Docker service? Is this only for local development or would it also look like this on a production-like environment?

skins/wmde19/templates/Base_Layout.html.twig
gbirke: OMG, I copied that 1:1 from cat17!

README.md
gbirke: Yeah, it's a typo, I'm no `1337 haxx0r` 😝

docker-compose.yml
gbirke: The docker-compose.yml file is only for local development. I thought starting the dev server together with the rest would save us some trouble during development, because we don't have to remember (and document) starting another server. When the app is deployed, we compile the assets and then there is no need for the dev server.

skins/wmde19/src/error.ts
```diff
@@ -0,0 +1,22 @@
...
+
+if ( dataElement ) {
+	applicationVars = JSON.parse( dataElement.dataset.applicationVars || '{}' );
+	messages = JSON.parse( dataElement.dataset.applicationMessages || '{}' );
+}
+
+new Vue( {
+	render: h => h( Error, {
+		props: {
+			errorMessage: applicationVars.message,
```
tzhelyazkova: where is `message` coming from?

skins/wmde19/src/error.ts
```diff
@@ -0,0 +1,22 @@
...
+			messages: messages,
```
tzhelyazkova: Isn't this property holding the error page messages? If yes, what's `errorMessage`? 🤔

README.md
tzhelyazkova: Is this `8082 minus 1010` based on some actual convention or is it just a joke? I'm curious 😄

docker-compose.yml
```diff
@@ -32,6 +32,7 @@ services:
     links:
       - database
       - mailhog
+      - skin_server_wmde19
```
tzhelyazkova: minor nitpick: since none of the rest is named `server` can we drop it from the title here as well `skin_wmde19`

gbirke:
> I suggest we rethink the naming convention of the `twig` files as it differs the components one.

The template names are hard-coded in PHP. Changing them would mean changing them for all skins. Not a horrendous task, but definitively some effort.

skins/wmde19/src/error.ts
gbirke: It's assigned to the template in `InternalErrorHtmlPresenter` and `ErrorPageHtmlPresenter`.

skins/wmde19/src/error.ts
gbirke: `messages` holds the translated i18n strings from `messages.json` in the content repository.
To avoid the confusion between `message` (i.e. error message) and `messages` (i.e. i18n string) I've called the property `errorMessage` in the component.

When we have an i18n mechanism, we can rename `errorMessage` to `message` again for greater consistency and less verbosity.

tzhelyazkova:
> > I suggest we rethink the naming convention of the `twig` files as it differs the components one.
> 
> The template names are hard-coded in PHP. Changing them would mean changing them for all skins. Not a horrendous task, but definitively some effort.

oooh right, ok nevermind then

README.md
gbirke: For all the FUN web applications I'm trying to use ports in the 808x range (8082 for FundraisingFrontend, 8083 for the Fundraising, 8084 for Banners) for the PHP side of things. And since I liked the repeating 8, I chose a repeating 7. It makes sense in my mind, but I fear I'm alone here :neckbeard: 🤓

docker-compose.yml
gbirke: What about `skin_provider_wmde19` 😈  *\*runs away\**

tzhelyazkova: triggered

skins/wmde19/src/pages/error.ts
```diff
@@ -1,23 +1,21 @@
...
 Vue.config.productionTip = false;
 
-const dataElement = document.getElementById( 'app' );
-let applicationVars: any = {}; // TODO Create interface
-let messages = {};
-
-if ( dataElement ) {
-	applicationVars = JSON.parse( dataElement.dataset.applicationVars || '{}' );
-	messages = JSON.parse( dataElement.dataset.applicationMessages || '{}' );
+interface ErrorModel {
```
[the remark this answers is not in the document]
gbirke: I was thinking the same. I wouldn't call it `types`, but maybe `models`, `viewModels` or `domain` - there will be mostly type definitions / interfaces in them and they will be referenced from Vue components and Vuex modules, so they are a very low-level dependency. Calling them `types` sounds too generic and "we're making special affordances for TypeScript" to me.

skins/wmde19/src/components/App.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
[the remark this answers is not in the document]
gbirke: I envision all `.vue` files in the `components`. `App` is basically the `Root.vue` I talked about yesterday, that's why I put it in the "root" of `components` and not in `components/layout`, `components/pages` etc. What do you think?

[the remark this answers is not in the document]
tzhelyazkova: `Vue-CLI` creates the root component outside of `/components` but I can't find a concrete reason for that. The [Vue Style Guide](https://vuejs.org/v2/style-guide/) doesn't mention anything about project structure. Looking at how `Vuex` structures their code I agree we should keep `App.vue` in `/components`.

PR #1464, opened 2019-04-08
New skin base layout

timEulitz:


skins/wmde19/src/components/layout/Content.vue
```diff
@@ -1,11 +1,33 @@
 <template>
-	<main>
-		<slot></slot>
-	</main>
+	<div class="container has-margin-top-112 has-margin-bottom-48">
+		<div class="columns">
+			<main class="column is-two-thirds has-background-white">
```
tzhelyazkova: not a fan of  `-white`. Does this come from bulma and can we change it?
Generally we shouldn't be concerned with the color itself, therefor it shouldn't be in the name of the css class.

gbirke: image

It's a Bulma thing.

skins/wmde19/src/components/layout/Content.vue
```diff
@@ -1,11 +1,33 @@
...
+				<slot></slot>
+			</main>
+			<aside class="column is-one-third">
+				<p>Verschlüsselte Datenübertragung</p>
+				<br>
+				<p><strong>Frage zum Spenden?</strong></p>
+				<p>Rufen Sie uns an oder nutzen Sie das Kontakformular</p>
+				<p><a class="phone-number" href="tel:0049030219158260">030 / 21 91 58 26 - 0</a></p>
+			</aside>
```
gbirke: Now that the translations are merged, you should use `$t()` here.

skins/wmde19/src/components/pages/Error.vue
```diff
@@ -14,8 +13,25 @@
...
+			<br><br>
+			In pellentesque massa placerat duis. Pellentesque eu tincidunt tortor aliquam nulla facilisi. Vulputate eu scelerisque felis imperdiet proin fermentum leo vel orci. Magna ac placerat vestibulum lectus mauris. Ipsum nunc aliquet bibendum enim facilisis gravida neque convallis a. Aenean euismod elementum nisi quis eleifend quam adipiscing vitae proin. Sagittis nisl rhoncus mattis rhoncus urna neque. Arcu odio ut sem nulla pharetra diam sit amet. Sit amet mattis vulputate enim nulla aliquet. A condimentum vitae sapien pellentesque habitant morbi tristique. In tellus integer feugiat scelerisque varius morbi enim nunc faucibus. Faucibus pulvinar elementum integer enim neque volutpat ac tincidunt vitae. Scelerisque varius morbi enim nunc.
+		</p>
+	</div>
```
gbirke: What is this for? Please remove before merging this branch.

skins/wmde19/src/scss/modifiers.scss
```diff
@@ -0,0 +1,8 @@
+//Margins are generally multiples of 16
```
gbirke: 16 or 18?

skins/wmde19/src/components/pages/Error.vue
timEulitz: Of course, just wanted to have content that would stretch to the bottom of the page. Though there is a bit of a problem on pages with very little content, the footer will not be aligned to the bottom of the screen right now (but it also does not do that on production right now).

skins/wmde19/src/components/layout/Header.vue
```diff
@@ -1,11 +1,56 @@
...
+				</a>
+				<a role="button" class="navbar-burger" aria-label="menu" aria-expanded="false">
+					<span aria-hidden="true"></span>
+					<span aria-hidden="true"></span>
+					<span aria-hidden="true"></span>
+				</a>
+			</div>
+
+			<div class="navbar-menu" id="navMenu">
+				<div class="navbar-start">
```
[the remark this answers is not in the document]
gbirke: Especially, if we later want to have a feature that highlights the current page. Although that one can be achieved with css classes instead of using `v-if` in the loop.

skins/wmde19/src/pages/error.ts
```diff
@@ -1,9 +1,12 @@
 import Vue from 'vue';
 import VueI18n from 'vue-i18n';
-import Error from '@/components/pages/Error.vue';
-import App from '@/components/App.vue';
 import PageDataInitializer from '@/page_data_initializer';
 import { DEFAULT_LOCALE } from '@/locales';
+import App from '@/components/App.vue';
+
+import Component from '@/components/pages/Error.vue';
```
tzhelyazkova: Why `Component` and not `Error` ?

skins/wmde19/src/components/App.vue
```diff
@@ -1,13 +1,22 @@
...
-			<Header></Header>
-			<Content>
-				<slot></slot>
-			</Content>
-			<Footer></Footer>
-		</div>
-	</section>
+	<div>
+		<header>
+			<Header :imagePath=imagePath></Header>
```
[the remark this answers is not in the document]
timEulitz: Makes sense to me, I will change it.

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -1,11 +1,81 @@
 <template>
```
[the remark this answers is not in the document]
timEulitz: Yeah, it's unfortunately needed to vertically align the footer items with the items in the main content DIV. It's a bit much but I think it's still in "okay" territory.

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -1,11 +1,81 @@
...
+		</div>
 	</div>
 </template>
 
 <script>
 export default {
 	name: 'Footer',
+	props: [ 'imagePath' ],
+	computed: {
+		iban: function () {
```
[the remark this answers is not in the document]
timEulitz: Didn't know  that, thanks.

skins/wmde19/src/pages/error.ts
timEulitz: The idea was to keep the file as "general" as possible because this file will need to be copy-pasted for each entry point. That way, all you would need to change in between files is the "Error.vue" file name while the rest of the file can stay the same. It's a bit weird and I really don't like the redundancy, but I cannot think of a good way to alleviate this right now.

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -1,11 +1,81 @@
 <template>
-	<div class="footer">
-		Footer
+	<div class="footer has-background-footer">
```
tzhelyazkova: The footer does not stick to the bottom of the page for big resolutions.
`margin-top: 80px` on `.footer` fixes that

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -1,11 +1,81 @@
...
+								<p>Bank für Sozialwirtschaft</p>
+								<p><strong>{{ $t( 'account_number')  }}:</strong> {{ $t( 'operator_account_number')  }}</p>
+								<p><strong>{{ $t( 'blz')  }}:</strong> {{ $t( 'operator_account_blz')  }}</p>
+							</div>
+						</div>
+					</div>
+				</div>
+				<div class="column is-one-third">
+					<ul class="list-menu">
+						<li v-for="link in footerMenu">
```
tzhelyazkova: `eslint` complains there's no `:key` passed for the different nodes.
https://vuejs.org/v2/guide/list.html#key
Also on the other places with `v-for`

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -1,11 +1,81 @@
...
 <script>
 export default {
 	name: 'Footer',
+	props: [ 'imagePath' ],
+	computed: {
+		iban: function () {
+			return this.$t( 'operator_account_iban' ).split( ' ' );
+		},
+	},
+	data: function () {
```
tzhelyazkova: Since we committed to TypeScript we'll have to define an interface for the object returned by `data`.
I propose we have a separate folder for typescript interfaces 'cause there will be a lot of them.
I know Gabriel is not a fan of `\types` so probably `models` or `datamodels` or something will be better.
Since the layout components are simple we can have the `ts` interfaces in the same file, something like `Layout.ts`

skins/wmde19/src/components/layout/Footer.vue
timEulitz: I fixed this with flex-boxes, should be good now.

skins/wmde19/src/components/App.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<header>
-			<Header :imagePath=imagePath></Header>
+			<Header :page-identifier="pageIdentifier" :imagepath="imagepath"></Header>
```
tzhelyazkova: this should be hyphenated `image-path`

skins/wmde19/src/components/layout/Footer.vue
```diff
@@ -7,15 +7,15 @@
 						<div class="level-left">
 							<div class="logo-footer">
 								<a href="https://www.wikimedia.de/">
-									<img :src="imagePath + '/images/logo-vertical-wikimedia.svg'" alt="Wikimedia Deutschland">
+									<img :src="imagepath + '/images/logo-vertical-wikimedia.svg'" alt="Wikimedia Deutschland">
```
tzhelyazkova: camel case was fine here

PR #1468, opened 2019-04-15
Change javascript action to vue action

moiikana:
does this look too hacky? is the data function right at this point?

### DIFF
--- a/skins/wmde19/src/components/layout/Header.vue
+++ b/skins/wmde19/src/components/layout/Header.vue
@@ -1,54 +1,50 @@
 <template>
-	<div class="navbar is-fixed-top has-background-bright has-shadow">
-		<div class="container">
-			<div class="navbar-brand">
-				<a class="navbar-item" href="/">
-					<img :src="imagePath + '/images/logo-horizontal-wikimedia.svg'" alt="Wikimedia Deutschland" width="144" height="29">
-				</a>
-				<a role="button" class="navbar-burger" aria-label="menu" aria-expanded="false">
-					<span aria-hidden="true"></span>
-					<span aria-hidden="true"></span>
-					<span aria-hidden="true"></span>
-				</a>
-			</div>
+    <div class="navbar is-fixed-top has-background-bright has-shadow">
+        <div class="container">
+            <div class="navbar-brand">
+                <a class="navbar-item" href="/">
+                    <img :src="imagePath + '/images/logo-horizontal-wikimedia.svg'" alt="Wikimedia Deutschland"
+                         width="144" height="29">
+                </a>
+                <a role="button" aria-label="menu" aria-expanded="false" @click="showNavbarBurger = !showNavbarBurger"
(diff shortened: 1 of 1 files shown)


skins/wmde19/src/components/layout/Header.vue
```diff
@@ -1,54 +1,48 @@
...
-			</div>
+    <div class="navbar is-fixed-top has-background-bright has-shadow">
+        <div class="container">
+            <div class="navbar-brand">
+                <a class="navbar-item" href="/">
+                    <img :src="imagePath + '/images/logo-horizontal-wikimedia.svg'" alt="Wikimedia Deutschland"
+                         width="144" height="29">
+                </a>
+                <a role="button" class="navbar-burger" aria-label="menu" aria-expanded="false"
+                   v-bind:class="{ 'is-active': showNavbarBurger }" v-on:click="showNavbarBurger = !showNavbarBurger">
```
tzhelyazkova: you can list all classes together in the vue directive like
`v-bind:class="[{ 'is-active': showNavbarBurger }, 'navbar-burger']"`
Also `v-on:click` can be `@click`

skins/wmde19/src/components/layout/Header.vue
```diff
@@ -1,54 +1,48 @@
...
-					</a>
-					<a class="navbar-item">
-						Spenden
-					</a>
-				</div>
-			</div>
-		</div>
-	</div>
+            <div class="navbar-menu" id="navMenu"
+                 v-bind:class="{ 'is-active': showNavbarBurger }" v-on:click="showNavbarBurger = !showNavbarBurger">
```
tzhelyazkova: Same for this `div`

skins/wmde19/src/components/layout/Header.vue
```diff
@@ -1,54 +1,48 @@
...
-		$navbarBurgers.forEach( el => {
-			el.addEventListener( 'click', () => {
-				el.classList.toggle( 'is-active' );
-				document.querySelector('.navbar-menu').classList.toggle('is-active');
-			} );
-		} );
-	}
-} );
-
+	export default {
```
gbirke: If you're feeling adventurous, have a peek at other components and do a `<script type="ts">` and see what happens and how you need to modify this component to become a TypeScript component.

tzhelyazkova:
> Please restore the package.json and packa-lock.json files, they should not shop up in this commit. There are many ways to do this, my approach would be
> 
> ```
> git checkout HEAD^ skins/10h16/package-lock.json
> git checkout HEAD^ skins/10h16/package.json
> # more checkouts for the other files here ...
> git add skins/*/package*.json
> git commit --amend
> ```
> 

It's kinda my fault. I gave @moiikana wrong advice. 
The idea was to revert the changes in the `package(-lock)?.json` files as they are not relevant for this change. You still might wanna do that with `git reset HEAD .` and then adding only `Header.vue` and amending the commit.
Git is not easy 😞

gbirke:
Merging because the failing CI is unrelated to this. Fixed in #1469

PR #1470, opened 2019-04-18
Combine CSS into one file

gbirke:
Configure webpack to combine the CSS from all entry points into one
file.

Unfortunately, due to a [webpack bug][1], this also generates an
additional JS file that must be included.

This is for https://phabricator.wikimedia.org/T220997

[1]: https://github.com/webpack/webpack/issues/7300


gbirke:
I've created a followup ticket https://phabricator.wikimedia.org/T221601

PR #1474, opened 2019-04-25
Rename imagePath property and attribute across new skin

tzhelyazkova:


skins/wmde19/src/pages/address_update_success.ts
```diff
@@ -25,12 +25,12 @@ const i18n = new VueI18n( {
 	},
 } );
 
-// TODO imagePath needs to be dynamically established
```
gbirke: Since the path is now established dynamically, please remove the comment - either in this PR (with a new commit) or in a separate one.

PR #1475, opened 2019-04-25
Add ADR for client side rewrite

gbirke:
This is for https://phabricator.wikimedia.org/T221000

### DIFF
--- a/doc/adr/008_Client_Side_Rewrite.md
+++ b/doc/adr/008_Client_Side_Rewrite.md
@@ -16,6 +16,13 @@ The engineering team, planning the modernization of the frontend code base as ou
 
 By applying small, incremental changes to the existing design, we are following good agile practices - always delivering value, always being able to stop what we're doing and always having a running application.  
 
+We don't risk breaking something that already works (ui/ux wise). [A rewrite always carries the risk of throwing away past bugfixes, "microfeatures", lessons learnt and implicit organizational knowledge](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i/).
+
+We can improve the skin "step by step" so to speak. First implement the Vue components, then get rid of Bootstrap 3 and replace it with Bulma possibly, then transform our CSS structure into a more meaningful one.
+
+This approach doesn't stop us from developing new features if need be. If we have more than 1 or 2 skins, the cost of UI changes due to new features becomes prohibitive.
+
+We have already invested effort in experiments (i.e. the BankData component in `cat17`) that show the evolution from the old code base to Vue components.    
 
 We are avoiding the temptation to test 3 different designs (`10h16`, `cat17` and `new`) against each other. From a UX perspective, testing 3 designs does not make sense, however experience shows that if the option is there, it might be done.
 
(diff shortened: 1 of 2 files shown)


gbirke:
Obvious joke on the side:
rewrite_frontend

timEulitz:
I have thought about it quite a bit and I do think that the benefits of a new skin outweigh the downsides in this particular case. It basically comes down to this section of the ADR:

> A rewrite that does not have to integrate with existing client-side code can be done faster than transitioning to Vue.js through incremental refactorings. A transition would require additional compatibility layers in the architecture during the transition phase. Those layers would affect the performance and make the code base even harder to understand. When the transition is finished, the compatibility layers would have to be removed, making the effort spent on them "wasted". The compatibility layers would also affect the overall architecture: They could improve it by making everything decoupled and testable, on the other hand could make it "non-standard" and avoid the easy integration of data model and view thet VueX and Vue offer.

In addition to the JavaScript-related parts of that section, I also think it would be much harder to implement something like Bulma into the already existing skin as the site would basically be completely broken unless all necessary changes would be implemented in one go - which also would be hard to do while some parts of the front-end still depend on JavaScript to be rendered correctly.

At the same time, the components would need to be written with Bulma's structure in mind, so it becomes a bit of a chicken and egg problem because we could not properly implement Bulma without the components supporting it yet we also could not implement the components without the general page structure following what Bulma expects (components may expect parent nodes to be flex-boxed, for example). It's not impossible to work around this, though it's definitely one of many little things that would add to an increased overall effort.

However, I do think we're now in this weird situation that @tzhelyazkova describes:

>I think creating a new skin puts too much on our hands. Like it or not we are now responsible for 3 skins. And those 3 skins will keep existing simultaneously until it's proven one performs the best (and that takes quite the time as we know).
>One skin is super old, 2/3rds of us don't know what's going on there. No new features have been implemented there.
>One skin is a weird mix of vue, redux, jquery, typescript, javascript, bootstrap and some framework we use for dropdowns.

I don't really see much benefit in keeping `cat17` around as it has been established that it has failed, so there should not really be the need to perform ABC tests during the campaign, though I can imagine that the product owners will consider this a necessity (first I thought it would make sense to keep it as a backup but if `10h16` performs better, there is really no reason to keep `cat17` around?!).

It's also odd that we now basically have two deprecated skins that will continue to diverge in terms of feature parity - `10h16` is already super off the rails when it comes to feature parity with `cat17` and I assume that we are now not implementing any new features into `cat17` anymore which will result in three divergent tracks of feature sets. I assume this will at somehow hurt the newer skins as I assume this can only negatively affect some statistics: More features will will likely negatively affect performance metrics. More features will also lead to more users visiting the website not for the purpose of donating but for other reasons (e.g. they may only visit to look at the FAQ). I think this kind of stuff is hard to track, so I am worried we'll draw one conclusions here later.

doc/adr/008_Client_Side_Rewrite.md
```diff
@@ -0,0 +1,50 @@
...
+
+Accepted
+
+## Context
+
+The mobile-friendly, more aesthetically pleasing design of the software (called `cat17`) did not gather as much donations as the "old" design, called `10h16`. UX engineering, product management and stakeholders from the fundraising department decided against going "back" to the mobile-unfriendly and hard-to-maintain `10h16` design and instead improving the existing `cat17` design by applying UX best practices and making educated guesses at what made `10h16` so attractive to donors.
+
+The engineering team, planning the modernization of the frontend code base as outlined in "[ADR 006 - Using Vue.js](006_Vue.js.md) and "[ADR 007 - Using Typescript](007_Typescript.md) now had to choose between an *evolution* of `cat17` and a *rewrite* of the client-side code. The benefits of one approach would be the drawbacks of the other, so the following sections will describe each approach through the lens of benefits.
+
+### Benefits of an evolution of `cat17`
```
tzhelyazkova: - We don't risk breaking something that already works (ui/ux wise).
- We can improve the skin "step by step" so to speak. First implement the Vue components, then get rid of bootstrap and replace it with bulma possibly, then transform our css structure into a more meaningful one.
- This approach doesn't stop us from developing new features if need be.

PR #1477, opened 2019-04-26
Improve credit card handling controller

gbirke:
Move Credit Card notification handling in a controller class

Adapt to new "return model" of the CreditCardNotificationUseCase: Check
Response instead of catching "Response Exceptions".

This also corrects an architecture violation where a use case Response
object was initialized in the controller.

Log errors that occurred while handling the request. Unsuccessful
sending an email only counts as warning.

This is for https://phabricator.wikimedia.org/T221970

### DIFF
--- a/app/Controllers/CreditCardPaymentNotificationController.php
+++ b/app/Controllers/CreditCardPaymentNotificationController.php
@@ -1,0 +1,49 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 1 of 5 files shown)


gbirke:
This PR needs https://github.com/wmde/fundraising-donations/pull/77 to be merged, plus an update of the fundraising-donation dependency.

PR #1480, opened 2019-04-29
Faq page content

moiikana:


skins/laika/src/components/Question.vue
```diff
@@ -0,0 +1,41 @@
...
+import { Vue, Component, Prop } from 'vue-property-decorator';
+import { Question as QuestionModel } from '@/view_models/faq';
+
+@Component
+export default class Question extends Vue {
+	@Prop() content!: QuestionModel;
+	@Prop() visibleQuestionId!: string;
+
+	get isOpen() {
+		return this.$vnode.key === this.visibleQuestionId;
```
tzhelyazkova: So apparently `$vnode` is a `Vue` internal object and it's probably not very nice that we're using it. Even if we decide to go with the decorators I'll feel better if we pass the question id as another property instead of accessing it on the node.

skins/laika/src/view_models/faq.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export interface FaqContent {
+	topics : Topic[],
+	questions : Question[],
+}
+
+/**
+ * Convert from json format (with snake case) to proper typescript format in interfaces ( camel case )
+ */
+export function faqContentFromObject( obj:any ): FaqContent {
```
tzhelyazkova: pls add space `obj: any`

skins/laika/src/view_models/faq.ts
```diff
@@ -0,0 +1,30 @@
...
+	return { topics: obj.topics,
```
tzhelyazkova: new line after `{`

PR #1488, opened 2019-05-06
Move endpoint URLs to template context

gbirke:
Move endpoint URLs from template markup to template context. This will enable the laika skin JavaScript to use those endpoints URL without adding data attributes to each template.

Also contains some cleanup in the routes.


timEulitz:
There's a notice in Travis:

>Notice: Undefined index: assets-path in /usr/share/nginx/www/spenden.wikimedia.de/current/src/Presentation/TwigEnvironmentConfigurator.php on line 63

PR #1490, opened 2019-05-07
Use wikimediade docker images

gbirke:
Don't build our own images, use the prebuilt environment.

This is for https://phabricator.wikimedia.org/T206807


gbirke:
Travis errors are unrelated and fixed by #1492

PR #1493, opened 2019-05-08
Fix Address Update URL generation

gbirke:
The address update URL can't be generated independently from the current
route in Routes.php, as it needs the parameter addressToken, which is
only available when the address-update route is active.

Undo the rename of a deprecated route.

This is a followup to #1488


gbirke:
Travis errors will be fixed by #1492

PR #1500, opened 2019-05-10
Test payment amount component

gbirke:


skins/laika/tests/unit/components/pages/donation_form/PaymentAmount.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+		const expectedAction = action( NS_PAYMENT, setAmount );
+		const expectedPayload = {
+			amountValue: '2342',
+			validateAmountURL: 'https://example.com/amount-check',
+		};
+
+		expect( store.dispatch ).toBeCalledWith( expectedAction, expectedPayload );
+	} );
+
+	it( 'sends "empty string to store when custom amount is invalid', () => {
```
timEulitz: Stray double quotes here.

skins/laika/tests/unit/components/pages/donation_form/PaymentAmount.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+			},
+			store: createStore(),
+			mocks: {
+				$t: () => {},
+			},
+		} );
+		const store = wrapper.vm.$store;
+		store.dispatch = jest.fn();
+
+		const customAmountInput = wrapper.find( '.custom-amount' );
```
tzhelyazkova: I suppose class `custom-amount` won't hold any css rules so why not use the id instead of the class to select this element? This way we can still stick to our and buefy's css naming schema.

skins/laika/src/components/pages/donation_form/PaymentAmount.vue
```diff
@@ -50,19 +50,18 @@ export default Vue.extend( {
 		formatAmount: ( amount: string ) => Number( amount ).toFixed( 0 ),
 	},
 	methods: {
-		toCents: ( amount: string ) => Number( amount ) * 100,
+		toCents: ( amount: string ) => Math.trunc( Number( amount ) * 100 ),
```
tzhelyazkova: :+1:

skins/laika/tests/unit/components/pages/donation_form/PaymentAmount.spec.ts
gbirke: Oh yes, this is an important separations of concerns and consistency issue. 

In the past, I've tried to avoid IDs, because the 10h16 skin duplicated parts of the page and IDs were not working. Also having IDs in the markup encouraged people to use them in CSS, which [some people](https://dev.to/claireparkerjones/reasons-not-to-use-ids-in-css-4ni4) and a [CSS linter](http://csslint.net/about.html) consider to be a bad thing. That thinking led to me avoiding IDs in the markup and using class names both for CSS and JavaScript. As you (and @moiikana in a discussion) pointed out, this practice leads to problems because it violates the single responsibility principle, making class names do two things.

With Vue.js the only instance where we have to reference inner html elements with a selector is in tests, but still we need to think about how we would like to do it.

I see several options going forward:
* Keep class names as designators for behavior and components, accepting the inconsistency with "class-names-as-designators-for-style". The SRP violation can be remedied by a rule that no markup element can have both JS and CSS-related classes. This will lead to nested markup.
* Use IDs. This might not be feasible for cases where elements are repeated.
* Use data attributes. To my surprise and against my knowledge, [using data attributes as selectors is not slower in modern browsers than using class names](https://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles/)
* Use [`$refs`](https://vuejs.org/v2/api/#ref). To me this looks like it exposes the "inner workings" of a component as an API to the outside, but maybe that's not a bad thing.

I think both data attributes and $refs are good candiates. What do you think, @tzhelyazkova , @timEulitz , @moiikana ?

timEulitz: I am in favor of using IDs as selectors in JavaScript. I think we set high enough standards for ourselves that we will not misuse them as CSS selectors and linting should back that up. Data attributes would also be fine but why over-complicate things when IDs are specifically made to solve this type of problem?

gbirke: Followup: After a brief discussion we decided on using ids for selectors in tests.

PR #1501, opened 2019-05-13
Payment page design

timEulitz:


timEulitz:
In the last commit, I have also moved some of the SCSS contents around as there were some issues with the order and how CSS rules would override each other. Now the order should be right and the files should be somewhat sensible once you take a look at them. We basically only need to include custom.scss in the .vue files now.

skins/laika/src/scss/overrides.scss
```diff
@@ -1,63 +1,42 @@
...
 // Miscellaneous overrides
 
 .title {
 	margin-top: 30px;
 }
 
+.b-radio.radio {
+	height: 4em;
+	width: 320px;
+	border-bottom: 2px solid $fun-color-off-bright;
```
gbirke: Since the lines repeat in the navigation and maybe in the FAQ page, I'm wondering if it makes sense to introduce a "is-underlined" or "has-bottom-line" class. But I'm still unsure of our CSS language, so I don't know if it fits.

skins/laika/src/components/pages/donation_form/PaymentAmount.vue
```diff
@@ -21,14 +21,13 @@
...
-							maxlength="9">
+							maxlength="9"
+							:placeholder="$t('custom_amount')">
 					<label for="amount-custom" class="is-sr-only">{{ $t('payment_amount_legend') }}</label>
 				</div>
 			</div>
 		</div>
 		<span class="help is-danger" v-if="hasErrors">{{ $t('donation_form_amount_error') }}</span>
-
-		<label for="amount-custom" class="sr-only">{{ $t('payment_amount_legend') }}</label>
```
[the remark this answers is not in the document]
timEulitz: It's still there, I just deleted the duplicate.

skins/laika/src/scss/overrides.scss
timEulitz: I see what you mean, we could have some sort of class for 2px primary color underlines. I am wondering if it makes things unnecessary complicated, though I guess that's what we signed up for?

For example, the custom amount input field gets the same type of underline when it's valid but there are some additional rules attached to it, so if we had to attach an is-valid and is-underlined class, it would unnecessarily bloat up the class list. I guess a mixin would make more sense?!

We can also discuss this properly in person some time and come to an agreement where to draw the line with these classes, I am also not sure what the best practices are in this area.

PR #1502, opened 2019-05-14
Test Payment Interval component

gbirke:


skins/laika/tests/unit/components/pages/donation_form/PaymentType.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const testPaymentMethods = [
+	{ id: 'debit-card', type: 'BEZ' },
+	{ id: 'paypal', type: 'PPL' },
+	{ id: 'bank-transfer', type: 'UEB' },
+];
+
+describe( 'PaymentType', () => {
+
+	it( 'sends new interval to store when amount is selected', () => {
```
tzhelyazkova: sends new type/method to store when it is selected

skins/laika/tests/unit/components/pages/donation_form/PaymentType.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+				$t: () => {},
+			},
+		} );
+		const store = wrapper.vm.$store;
+		store.dispatch = jest.fn();
+
+		wrapper.find( '#debit-card' ).trigger( 'click' );
+		const expectedAction = action( NS_PAYMENT, setType );
+
+		expect( store.dispatch ).toBeCalledWith( expectedAction, 'BEZ' );
```
tzhelyazkova: I think it might be better if we do `wrapper.find( '#' + testPaymentMethods[0].id)` and then `expect( store.dispatch ).toBeCalledWith( expectedAction, testPaymentMethods[0].type) );`

skins/laika/tests/unit/components/pages/donation_form/PaymentInterval.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const testIntervals = [
+	{ id: 'one-time', intervalInMonths: 0 },
+	{ id: 'monthly', intervalInMonths: 1 },
+	{ id: 'yearly', intervalInMonths: 12 },
+];
+
+describe( 'PaymentInterval', () => {
+
+	it( 'sends new interval to store when amount is selected', () => {
```
tzhelyazkova: `when it is selected`

skins/laika/tests/unit/components/pages/donation_form/PaymentType.spec.ts
gbirke: I have spent quite some time going back and forth between those versions and a third version, where I extract the data to view_model and use it in the test. There are tradeoffs for both versions:

* The "plain strings and values" version conforms to the principle that tests should state plain and simple what the values are, without complicated assembly or data from external sources. It also demonstrates that the component works independently from the properties that are passed in (I could have passed in 2-month intervals etc). 
* The "composed expected values"/"external data" version has the benefit of using the "real" data that the user sees, with less risk that tests miss something. It can also make the test more expressive, by explicitly stating what, for example, `year` and `12` *mean* (`id` and `intervalInMonths`). 

If you like the "composed expected data" version better, I'll gladly change the tests.

tzhelyazkova: Oh ok, I don't mind, I just thought that it's harder to see that BEZ corresponds to debit card when they are plain strings, because a lot of the language in our domain is not common knowledge.

gbirke:
I've refactored the property data types for intervals and payment types to the numbers and strings that come from the server, to keep server and client side in sync.

PR #1505, opened 2019-05-15
Put all faq questions on one page and add styling

moiikana:
As by the new design decision, all questions in all topics
are shown on the faq page now. They can be opened and closed and
only one question is open at a time.

### DIFF
--- a/skins/laika/src/components/Question.vue
+++ b/skins/laika/src/components/Question.vue
@@ -1,31 +1,32 @@
 <template>
-    <div class="question space-above">
+    <div class="question">
         <div v-bind:class="[ isOpen ? 'border-secondary' : 'border-primary' ]">
             <div class="inline-icon" @click="toggle()"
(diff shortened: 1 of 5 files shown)


skins/laika/src/components/Question.vue
```diff
@@ -1,31 +1,30 @@
 <template>
-    <div class="question space-above">
-        <div v-bind:class="[ isOpen ? 'border-secondary' : 'border-primary' ]">
-            <div class="inline-icon" @click="toggle()"
-                 :data-content-target="!isOpen ? '/page/Häufige Fragen' : ''"
-                 :data-track-content="!isOpen"
-                 :data-content-name="!isOpen ? 'Expand' : ''"
-                 :data-content-piece="!isOpen ? content.question : ''">
```
tzhelyazkova: We use these data attributes to track how often the questions are opened.
Unless this is not a requirement anymore they should be put back in the `<div>`

PR #1507, opened 2019-05-15
Email Component

tzhelyazkova:
:email: :postbox: 
This component is part of the Address Form - [T219172](https://phabricator.wikimedia.org/T219172)
There will be a follow up PR which adds the Newsletter Opt In component to the Address Form.


skins/laika/src/components/pages/donation_form/Email.vue
```diff
@@ -0,0 +1,48 @@
...
+	},
+	methods: {
+		setEmail: function () {
+			this.$store.dispatch( action( NS_ADDRESS, setEmail ), this.$data.emailValue );
+		},
+		validateEmail: function () {
+			if ( this.emailIsValid() ) {
+				this.emailHasError = false;
+				this.setEmail();
+			} else { this.emailHasError = true; }
```
timEulitz: You could avoid the `else` here:

```
this.emailHasError = !this.emailIsValid();
if ( !this.emailHasError ) {
     this.setEmail();
}
```

I don't think it's a super big deal but the method also validates and sets the email, maybe it could also just consolidate the set and validate function into one because I guess this would be more accurate?!

```
setEmail: function () {
    this.emailHasError = !this.emailIsValid();
    if ( !this.emailHasError ) {
        this.$store.dispatch( action( NS_ADDRESS, setEmail ), this.$data.emailValue );
    }
```

tzhelyazkova: I'd like to keep `setEmail` clean a.k.a doing only one thing.
That's why I made `emailIsValid` and `validateEmail` to glue the logic together.

timEulitz: Makes sense!

PR #1508, opened 2019-05-15
Implement Laika address design

timEulitz:
There are still some minor details which do not match completely but I would also like to merge, deploy and test this on my phone / IE11. I think the form scaling also needs some love for mobile, there may not be enough space on smaller devices.

### DIFF
--- a/skins/laika/src/components/pages/donation_form/Name.vue
+++ b/skins/laika/src/components/pages/donation_form/Name.vue
@@ -39,9 +39,10 @@
 			</b-select>
 		</div>
 		<span v-if="showError.salutation" class="help is-danger"> {{ $t( 'form_salutation_error' ) }}</span>
-		<div v-bind:class="[{ invalid: showError.firstName }]">
+		<div v-bind:class="[{ 'is-invalid': showError.firstName }]">
 			<label for="first-name" class="subtitle has-margin-top-36">{{ $t( 'firstname_label' ) }}</label>
-			<b-input type="text"
+			<b-input class="is-medium"
+					type="text"
 					id="first-name"
 					v-model="formData.firstName.value"
 					name="firstName"
@@ -50,7 +51,7 @@
 			</b-input>
 			<span v-if="showError.firstName" class="help is-danger">{{ $t( 'form_firstname_error' ) }}</span>
 		</div>
-		<div v-bind:class="[{ invalid: showError.lastName }]">
+		<div v-bind:class="[{ 'is-invalid': showError.lastName }]">
 			<label for="last-name" class="subtitle has-margin-top-36">{{ $t( 'lastname_label' ) }}</label>
 			<b-input type="text"
 					id="last-name"
@@ -62,7 +63,7 @@
 			<span v-if="showError.lastName" class="help is-danger">{{ $t( 'form_lastname_error' ) }}</span>
 		</div>
 	</div>
-	<div v-else-if="addressType === AddressTypeModel.COMPANY" v-bind:class="[{ invalid: showError.companyName }]">
+	<div v-else-if="addressType === AddressTypeModel.COMPANY" v-bind:class="[{ 'is-invalid': showError.companyName }]">
 		<label for="company-name" class="subtitle has-margin-top-36">{{ $t( 'companyname_label' ) }}</label>
 		<b-input type="text"
 				id="company-name"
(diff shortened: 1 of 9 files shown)


skins/laika/src/components/pages/donation_form/Postal.vue
```diff
@@ -1,48 +1,52 @@
 <template>
 <div>
 	<div v-bind:class="[{ invalid: showError.street }]">
-		<label for="street">{{ $t('street_label') }}</label>
-		<input type="text"
+		<h2 class="subtitle has-margin-top-36">{{ $t( 'street_label' ) }}</h2>
```
gbirke: Please use `<label>` for better accessibility and semantics

skins/laika/src/scss/custom.scss
```diff
@@ -8,6 +8,21 @@
...
+.is-form-input {
+	.select {
+		select {
+			border-radius: 0;
+			width: 300px;
+			font-size: 1rem;
+			line-height: 40px;
+			height: 48px;
+		}
+		// Don't hate the player, hate the game
```
gbirke: Please explain this madness in more detail

skins/laika/src/components/pages/donation_form/Name.vue
```diff
@@ -1,82 +1,76 @@
...
-				id="company-name"
-				name="company"
-				:placeholder="$t( 'companyname_label' )"
-				v-model="formData.companyName.value"
-				@blur="validateInput(formData, 'companyName')">
-		<span v-if="showError.companyName" class="error-text">{{ $t( 'form_companyname_error' )  }}</span>
-	</div>
-	<div v-else-if="addressType === AddressTypeModel.PERSON">
-		<label for="salutation">{{ $t( 'salutation_label' ) }}</label>
+	<div v-if="addressType === AddressTypeModel.PERSON">
```
tzhelyazkova: Why swap the if statement?

skins/laika/src/scss/overrides.scss
```diff
@@ -6,6 +6,14 @@
 	margin-top: 30px;
 }
 
+.input {
+	max-width: 300px;
+}
+
+.is-danger, .is-warning {
+	font-size: 1em;
```
tzhelyazkova: Is there a requirement to have different font size for error messages? :thinking:

skins/laika/src/components/pages/donation_form/Name.vue
timEulitz: My brain kept being confused because the list goes:

Person
Company
Anon

But the code was:

Company
Person
Nothing

So this just seems more logical. Also, since "Person" is selected by default, it saves an ultra small amount of CPU time because previously, the first `if` was always false.

skins/laika/src/scss/overrides.scss
timEulitz: Error messages in have a bigger font size by default, it looks pretty horrible in our design so I have to force it back to 1em (which is the default font-size of 14px).

skins/laika/src/components/pages/donation_form/Name.vue
tzhelyazkova: Ok, I'm fine with leaving it this way, but the Address Type order has nothing to do with this component.
If UX decides we should reorder to Anon Company Person you won't go and reorder the `if-else`s here.

skins/laika/src/scss/overrides.scss
tzhelyazkova: My concern was that we are overriding font size specifically for this. I did a search in `\laika` for `font-size`: 
Generally the font is `14px`, but we have here 1em, and for radio we have `13px`. The text in the input fields is set to font size `1rem`.
Overall we have different font sizes and different unit measurements. It can easily get out of hand. Also just remembered that Volker mentioned pixels are not accessibility friendly

timEulitz: I generally only use em or rem. The only places where you should see pixels on font sizes is the root HTML element, if you see it anywhere else, please do point it out in a review.

The radio button labels still have 14px (or rather 1 em) but I have forced the actual round radio element to be 13 pixels to resolve the "googly eyes" problem for the most part. I would call that somewhat of a "hack" but it was the best I could do without completely rewriting the Buefy radio buttons, it has something to do with how the radio buttons get rendered when their reference font size is 14px.

PR #1512, opened 2019-05-23
Remove remaining unsued state keys

tzhelyazkova:
:bug: :beetle: :ant: Bug: [T222798](https://phabricator.wikimedia.org/T222798)

### DIFF
--- a/skins/cat17/src/app/lib/validation_dispatchers/bankdata.js
+++ b/skins/cat17/src/app/lib/validation_dispatchers/bankdata.js
@@ -17,7 +17,7 @@ var objectAssign = require( 'object-assign' ),
 	 * @return {ValidationDispatcher}
 	 */
 	createBankDataValidationDispatcher = function ( validator, initialValues ) {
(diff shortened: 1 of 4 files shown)


skins/cat17/src/app/lib/validation/bank_data_validator.js
```diff
@@ -26,21 +26,21 @@ export class BankDataValidator {
...
-	 * @param {string} bankCode
+	 * @param {string} accountId
+	 * @param {string} bankId
 	 * @return {Promise}
 	 */
-	validateClassicBankData( accountNumber, bankCode ) {
-		accountNumber = accountNumber.replace( /\s/g, '' );
-		bankCode = bankCode.replace( /\s/g, '' );
-		if ( accountNumber === '' || bankCode === '' ) {
+	validateClassicBankData( accountId, bankId ) {
```
gbirke: I'm not sure if we should rename the fields - yes, it's more consistent with `validateBankData`, but at the same time, the "domain language" of the classic bank data - at least in the translation from German - would be ""account number" and "bank code".

PR #1513, opened 2019-05-23
Simplify by removing uuidgen substitutions

adamwight:


README.md
```diff
@@ -24,7 +24,9 @@ User facing application for the [Wikimedia Deutschland](https://wikimedia.de) fu
 
 ## Installation
 
-For development you need to have Docker and docker-compose installed. You need at least Docker Version >= 17.09.0 and docker-compose version >= 1.17.0. If your OS does not come with the right version, please use the official installation instructions for [Docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/). You don't need to install other dependencies (PHP, Node.js, MySQL) on your machine.
+For development you need to have Docker and docker-compose installed. You need at least Docker Version >= 17.09.0 and docker-compose version >= 1.17.0. If your OS does not come with the right version, please use the official installation instructions for [Docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/). You don't need to install other dependencies (PHP, Node.js, MySQL) on your machine, but you will need:
+
+    sudo apt install uuid-runtime
```
tzhelyazkova: hm, that's specific to Linux. I'm not sure it belongs here.

adamwight: Good point!  The general idea is that mentioning the package name might be useful for some people, since it's not bundled in the default installation and doesn't match the missing script name (`uuidgen`).  I'm not sure how to best document, personally I hate those multi-section instructions like "for Debian .... for Ubuntu ....".

gbirke: @adamwight Thanks for the first patch and trying out things!

Instead of a verbose text we could just have a "System requirements" subsection, mentioning `uuidgen` as a requirement. The distribution-specific package name is then up to the person installing the dev environment.

adamwight: Sorry to shift the goalposts here, but: would you mind explaining why we need the uuidgen in the first place?  It seems to create a new container name every time you run the makefile, which I don't understand.  The `-$@` suffix makes sense because you don't want e.g. a phpunit container colliding with your development container, but that clearly isn't the purpose of the UUID.

gbirke: I really appreciate the questions and questioning. It's good to have a fresh pair of eyes. 

Some time ago I wanted to speed up the `make ci` command, because it does lots of stuff: Checking the code style, running phpUnit tests, running PHPStan, Building the JS assets, running all JavaScript tests. `make` can run tasks in parallel with the [`-j` flag](https://www.gnu.org/software/make/manual/html_node/Parallel.html). While using the parallel execution feature, `docker-compose` ran into trouble because it was trying to reuse the `app` container (don't know the specifics any more, just that there were errors). Giving each `app` container a unique name solved that problem. See #1245 for the commit that introduced it.

I must admit that after implementing this, I've totally forgotten about the `-j` flag and never used it. The PR description mentions travis, but it was not activated there. It would probably ony be a moderate speed up as the Travis environment provides only 2 CPU cores.

I've just measured the speed improvements and found them to be semi-negligible. On my machien the parallel execution shaved off 40 seconds (of a 240 second build). When the build process has an error in the earlier stages, the non-parallel execution is even *faster* since `make` won't execute all following tasks. 

So, what do you think - is the tradeoff between a slightly more complex Makefile and a system dependency on the one side versus a potential, but rarely used speed up of the build process worth it? 

If we decide to remove the parallel execution feature, then please remove this edit from the PR and do a separate PR that reverts #1245

adamwight: I wonder if it would be possible to detect the `-j` flag and only include the extra UUID names when the flag is present.  That doesn't simplify the Makefile, but supports the slight speedup when desired, and protects casual users from the external dependency.

adamwight: I can't find a nice way to test the flags, since `findstring` will match partial words, e.g. on my machine `$(MAKEFLAGS)` always contains `--jobserver-fds=3,4` which causes `$(findstring j, $(MAKEFLAGS))` to evaluate to true.

Here's a nasty solution though,
```
UNIQUE_APP_CONTAINER = $(shell (echo $(MAKEFLAGS) | grep -qE -- '-j[0-9]*\>') && uuidgen)-app
```

Makefile
```diff
@@ -7,7 +7,7 @@ NPM_FLAGS     := --prefer-offline
 DOCKER_FLAGS  := --interactive --tty
 TEST_DIR      :=
 REDUX_LOG     :=
-UNIQUE_APP_CONTAINER := $(shell uuidgen)-app
```
adamwight: tbh, I'm creeped out that simple assignment `:=` doesn't work here and I don't understand why not, but it seems that the MAKEFLAGS aren't available at this point in the script.

So-called recursive assignment is annoying because it causes a new UUID to be generated for each step, but it might be harmless.

gbirke:
Thank you so much for your efforts to make this work! I still think we should rather remove that unneeded feature than making the Makefile more complicated. 

Ping @timEulitz @tzhelyazkova What do you think?

timEulitz:
I am in favor of just getting rid of the whole UUID approach altogether. Unless I am missing something here, I think Adam's change does not really resolve the 'dependency issue' with UUID and makes the Makefile more complicated, so I would rather avoid that and keep it as simple as possible.

adamwight:
I have no opinion about this, personally--I think the makefile is fine with or without the `uuidgen` magic.  But here's the patch in case you decide to go ahead with removing these!

PR #1514, opened 2019-05-24
[WIP] Add content of the supporters list page

moiikana:


skins/laika/templates/partials/back_link.html.twig
```diff
@@ -0,0 +1,5 @@
+<div class="back-block col-xs-12 col-sm-12 col-md-12">
+	<ul class="list-menu link-back list-unstyled">
+		<li><a href="{$ path( '/' ) $}" class="">{$ 'back_to_donation' | trans $}</a></li>
+	</ul>
+</div>
```
gbirke: I think this template might have slipped into this PR by accident. The laika version of `templates/page_layouts/base.html.twig` does not use it.

skins/laika/src/pages/supporters.ts
```diff
@@ -0,0 +1,107 @@
...
+
+	removeTableHeading();
+
+	createNewAccordionElements();
+
+	renameTableToDiv();
+
+	addToggleFunction();
+
+};
```
gbirke: Changing the HTML works and conforms to the design, but for every design change we now have to dig in this code. I appreciate that we're not duplicating the supporters list HTML, but another solution would have been to render the content in a hidden div and then extract the data from the HTML and pass it to a `SupportersList` Vue component. I think that code might even be smaller. 

It's not a blocker, but if we leave the code like that it's technical debt in the original sense of the word.

timEulitz:
Continuing this task in different PR with slightly different approach: #1602

PR #1517, opened 2019-05-27
Fix donation confirmation email templates

timEulitz:
https://phabricator.wikimedia.org/T202860


app/mail_templates/Donation_Confirmation.txt.twig
```diff
@@ -1,5 +1,5 @@
-{%- set formattedAmount = donation.amount|localizedcurrency('Euro', 'de_DE') -%}
```
timEulitz: This will always print the euro sign which is why it has to be replaced with a plain number_format as far as I can tell.

Previously I thought the first parameter of `localizedcurrency` was for the currency text but it actually needs to be 'EUR' according in line with some ISO standard and will then always force the currency to be rendered as `12,34 €`.

PR #1520, opened 2019-05-31
Load donation form pages dynamically

tzhelyazkova:
:heavy_check_mark:  [T224115](https://phabricator.wikimedia.org/T224115)
:page_facing_up: Split the donation form into pages and load them dynamically when `next` or `previous` buttons are clicked. Make `next` button disabled in case of invalid fields.
:memo: Implementing the Overview Component and the Submit functionality will be in separate PRs.

### DIFF
--- a/skins/laika/src/components/pages/DonationForm.vue
+++ b/skins/laika/src/components/pages/DonationForm.vue
@@ -55,29 +55,28 @@ export default Vue.extend( {
 	data: function () {
 		return {
 			currentFormComponent: 'Payment',
-			buttonsVisibility: { 
-				previous: false, 
-				next: true, 
-				submit: false, 
+			buttonsVisibility: {
+				previous: false,
+				next: true,
+				submit: false,
 			},
 		};
 	},
 	computed: {
 		currentProperties: {
-			get() {
-				if ( this.$data.currentFormComponent === 'Payment' ) {
-					return {
-						validateAmountUrl: this.$props.validateAmountUrl,
-						paymentAmounts: this.$props.paymentAmounts,
-						paymentIntervals: this.$props.paymentIntervals,
-						paymentTypes: this.$props.paymentTypes,
-					};
-				} else if ( this.$data.currentFormComponent === 'AddressForm' ) {
+			get(): object {
+				if ( this.$data.currentFormComponent === 'AddressForm' ) {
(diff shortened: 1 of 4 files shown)


gbirke:
Wouldn't it make the code simpler and the components more reusable  if the display and event handling of the "previous/next/submit" buttons was handled by `DonationForm`?

tzhelyazkova:
> Wouldn't it make the code simpler and the components more reusable if the display and event handling of the "previous/next/submit" buttons was handled by `DonationForm`?

I looked at the Figma components again and it seems that for Membership the pages are swapped compared to Donation. So yes, it makes more sense to move out the buttons from `Payment` and `Address` so we can reuse them more easily. Thanks :+1:

skins/laika/src/components/pages/DonationForm.vue
```diff
@@ -39,5 +58,67 @@ export default Vue.extend( {
...
+			buttonsVisibility: {
+				previous: false,
+				next: true,
+				submit: false,
+			},
+		};
+	},
+	computed: {
+		currentProperties: {
+			get(): object {
```
gbirke: Another instance where strong typing would both help (by making sure the types match for the component and inside the computed property) and be a pain.

tzhelyazkova:
> Looks good to me. I am not sure if this is still WIP but since Gabriel asked you to create tests for this, it sounded like this is the agreed behavior:
> 
> I think the "Next" button should not be `disabled` when the payment data is not filled out yet. It makes it hard to see why the button is inactive and we should rather make it clickable and show the errors for the missing form values.

I marked it as ready for review so it's not a WIP. I will address Gabriel's comment.
The behavior of the next button is the one described in the ticket so that's a design decision I assume.

timEulitz:
@tzhelyazkova 

> The behavior of the next button is the one described in the ticket so that's a design decision I assume.

I don't think the respective line in the ticket meant that the button would be disabled but rather that the user is shown the error messages and nothing happens. As far as I remember, the button should always be active and validation error messages should be displayed when the button is clicked:

_If a form section is invalid, respective error messages are displayed. Proceeding to the next page is not allowed in that case._

Maybe double-check with Kai or Jan before you implement any tests, I am fairly sure about this but I may be wrong.

PR #1523, opened 2019-06-05
Update membership confirmation email

timEulitz:
Ticket: https://phabricator.wikimedia.org/T202862

Linting issues that surfaced while working on this have been documented in this ticket:

https://phabricator.wikimedia.org/T225086

### DIFF
--- a/app/mail_templates/Membership_Application_Confirmation.txt.twig
+++ b/app/mail_templates/Membership_Application_Confirmation.txt.twig
@@ -1,4 +1,4 @@
-{%- set formattedAmount = membershipFee|number_format(2, ',', '.') -%}
+{%- set formattedAmount = membershipFee|number_format(2, ',', '.') ~ ' Euro' -%}
 {% if membershipType == 'sustaining' %}
 {$- greeting_generator.createInformalLastnameGreeting( salutation, lastName )|raw $}
 {% else %}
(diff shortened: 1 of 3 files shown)


app/mail_templates/Membership_Application_Confirmation.txt.twig
```diff
@@ -1,4 +1,4 @@
-{%- set formattedAmount = membershipFee|localizedcurrency('EUR', 'de_DE') -%}
+{%- set formattedAmount = membershipFee|number_format(2, ',', '.') -%}
```
gbirke: This change made me go "Huh?" and I had to go to the ticket and read the comments to see why this was needed. In the future, could you write the "why" of the change (or a short description of the change) in the commit messages? It will make it easier for our future selves to track why we made changes.

And I'm not sure if we should move the hard coded currency name from the application to the translation files. `locallizedcurrency` might not have produced the right results, but this patch changes the meaning of the variable from "localized number with currency name" to "localized number". Is this change consistent across all mail template files?

timEulitz: Hey, I already explained this in this PR but forgot to change this file as well:

https://github.com/wmde/FundraisingFrontend/pull/1517

But alright, I see how I should have added this to the commit message because that info is now hard to find. Initially I also wanted to leave a comment in the code but I was afraid it would mess up the email formatting :see_no_evil:.

timEulitz: I have updated the commit message now but I am not sure what other changes you expect.

timEulitz: As discussed: Add " Euro" string to the end of `formattedAmount` and remove it from the email templates to ensure the currency is not mentioned in the templates and only here. Same for donation confirmation email.

timEulitz: @gbirke recheck :robot:

PR #1524, opened 2019-06-05
Separate localization strings for Laika skin

timEulitz:
The Laika skin will now have its own messages file in order to ensure
that only strings which are actually used by the skin are loaded by the
client. Previously, old and unused strings from the previous skins was
also loaded alongside new strings which was somewhat messy. This
approach also means we will not have to do a clean-up of the messages
files once the old skins will be removed, as the new laika_messages.json
will only contain strings which we actually need.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8104,12 +8104,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/wmde/fundraising-frontend-content",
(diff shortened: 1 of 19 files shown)


timEulitz:
I have also fixed a few little things as part of this PR, I have created the correct place holder texts, for example.

timEulitz:
The new messages file can be found here:

https://github.com/wmde/fundraising-frontend-content/blob/test/i18n/de_DE/messages/messages_laika.json

PR #1527, opened 2019-06-13
Add membership benefits to confirmation page

timEulitz:
https://phabricator.wikimedia.org/T223061

### DIFF
--- a/skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
+++ b/skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
@@ -4,17 +4,19 @@
             <h2 class="title is-size-5">{{ $t( 'donation_confirmation_membership_call_to_action_title' ) }}</h2>
             <p>{{ $t( 'donation_confirmation_membership_call_to_action_text' ) }}</p>
             <p>
-                <a href="#">
+                <a :href="membershipApplicationUrl">
                     <b-button type="is-primary is-main has-margin-top-36">{{ $t('donation_confirmation_membership_button') }}</b-button>
                 </a>
             </p>
         </div>
         <div class="column is-half">
-            <p>{{ $t( 'donation_confirmation_bottombox_membership_benefit_1' ) }}</p>
-            <p class="has-margin-top-18">{{ $t( 'donation_confirmation_bottombox_membership_benefit_2' ) }}</p>
-            <p class="has-margin-top-18">{{ $t( 'donation_confirmation_bottombox_membership_benefit_3' ) }}</p>
-            <p class="has-margin-top-18">{{ $t( 'donation_confirmation_bottombox_membership_benefit_4' ) }}</p>
-            <p class="has-margin-top-18"><a href="https://wikimedia.de/de/mitglied-werden">{{ $t( 'donation_confirmation_bottombox_membership_link' ) }}</a></p>
+            <ul class="membership-benefits">
+                <li>{{ $t( 'donation_confirmation_bottombox_membership_benefit_1' ) }}</li>
+                <li class="has-margin-top-18">{{ $t( 'donation_confirmation_bottombox_membership_benefit_2' ) }}</li>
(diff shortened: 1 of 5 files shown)


skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
```diff
@@ -3,9 +3,19 @@
         <div class="column is-half">
             <h2 class="title is-size-5">{{ $t( 'donation_confirmation_membership_call_to_action_title' ) }}</h2>
             <p>{{ $t( 'donation_confirmation_membership_call_to_action_text' ) }}</p>
-            <p><b-button type="is-primary is-main">{{ $t('donation_confirmation_membership_button') }}</b-button></p>
+            <p>
+                <a href="#">
```
timEulitz: @tzhelyazkova This link actually needs to be like on cat17 where it leads to the membership application page with the form filled out but I did not get around doing that.

See cat17 for reference: `skins/cat17/templates/partials/donation_confirmation/membership_advantages.html.twig`, line 17.

gbirke: You can already generate the link target with the available data on the confirmation page, do you want to change the link in a different PR or can you squeeze that in?

skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
```diff
@@ -3,9 +3,19 @@
...
+                    <b-button type="is-primary is-main has-margin-top-36">{{ $t('donation_confirmation_membership_button') }}</b-button>
+                </a>
+            </p>
+        </div>
+        <div class="column is-half">
+            <p>{{ $t( 'donation_confirmation_bottombox_membership_benefit_1' ) }}</p>
```
gbirke: Could you make this a list? The paragraph markup is not very semantic and looks funny when the lien is too long (text of the 2nd line on the same indentation level as the bullet point). With a list you can solve that.

skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
```diff
@@ -3,9 +3,21 @@
...
-            <p><b-button type="is-primary is-main">{{ $t('donation_confirmation_membership_button') }}</b-button></p>
+            <p>
+                <a :href="membershipApplicationUrl">
+                    <b-button type="is-primary is-main has-margin-top-36">{{ $t('donation_confirmation_membership_button') }}</b-button>
+                </a>
+            </p>
+        </div>
+        <div class="column is-half">
+            <ul>
+                <li>{{ $t( 'donation_confirmation_bottombox_membership_benefit_1' ) }}</li>
```
gbirke: The list bullets are still in the messages text and not in the CSS for the list items. Having `<li>` elements is already semantically better, but not from a typography/aestetics stand point. Would you mind making that change?

skins/laika/src/components/pages/donation_confirmation/MembershipInfo.vue
```diff
@@ -15,13 +27,29 @@ import Vue from 'vue';
...
 
 <style lang="scss">
     @import "../../../scss/custom";
-    .membership-benefits-wrapper {
-        border: 1px solid $fun-color-primary-light;
-        border-radius: 2px;
+    .membership-benefits {
+        li {
+            &:before {
+                content: "– ";
```
gbirke: Styling lists in a typographically correct way (bullets outside of the line break area) should be done with `list-style`. Please confer with UX if bullets instead of dashes are ok. If yes, then use `list-style: disc`, otherwise you need [add margin trickery](https://stackoverflow.com/a/16507041/130121) since `list-style: "–"` is not  [yet supported on most browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#Browser_compatibility).

timEulitz: I checked with Jan and he said it's fine to use standard bullet points.

PR #1532, opened 2019-06-24
Add tracking data to donation form

timEulitz:
T226378

### DIFF
--- a/src/Presentation/Presenters/DonationFormPresenter.php
+++ b/src/Presentation/Presenters/DonationFormPresenter.php
@@ -45,9 +45,7 @@ class DonationFormPresenter {
 			],
 			'tracking' => [
 				'bannerImpressionCount' => $trackingInfo->getSingleBannerImpressionCount(),
-				'impressionCount' => $trackingInfo->getTotalImpressionCount(),
-				'trackingSource' => $trackingInfo->getSource(),
-				'tracking' => $trackingInfo->getTracking()
+				'impressionCount' => $trackingInfo->getTotalImpressionCount()
 			],
 			'urls' => $urlEndpoints
 		] );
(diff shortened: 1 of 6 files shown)


src/Presentation/Presenters/DonationFormPresenter.php
```diff
@@ -46,7 +45,9 @@ public function present( Euro $amount, string $paymentType, int $paymentInterval
 			],
 			'tracking' => [
 				'bannerImpressionCount' => $trackingInfo->getSingleBannerImpressionCount(),
-				'impressionCount' => $trackingInfo->getTotalImpressionCount()
+				'impressionCount' => $trackingInfo->getTotalImpressionCount(),
+				'trackingSource' => $trackingInfo->getSource(),
+				'tracking' => $trackingInfo->getTracking()
```
timEulitz: Not entirely sure about this, I figured this is the best way to do this but please check this out in detail. I also had to make those two functions nullable for cases where no value was set.

gbirke: I was baffled that `trackingSource` and `tracking` were not part of any form in the skins (as I assumed when writing the ticket, sorry for that) and had to search hard how that mechanism works. I found in `app/Bootstrap.php` where the values are read from various potential sources (cookie, a field called `tracking` that we probably haven't used for a while and finally Piwik (Matomo) campaign and keyword), set as attributes in the request and stored in a cookie after each request. In essence, this means that the `trackingSource` and `tracking` values are "magically" available, without us having to pass them around in form fields. You can remove them from the component and presenter.

I'll create a ticket for improving this inconsistent tracking situation. IMHO the best approach would be to save every tracking-related information in the session (the `spenden_tracking` cookie is only valid until the end of the session) and remove the impression counts from the forms.

PR #1533, opened 2019-06-25
Create bank account form fields

timEulitz:


gbirke:
I'd modify the code as follows:
* have two actions for the store, e.g. `setBankData( iban: string, bic?: string)` and `setLegacyBankdata(accountNumber: string, bankNumber: string)`, that call the right server API endpoint  (`check-iban` vs `generate-iban`) and call the validity mutators when the server returns a response. If the server returns success, set the values for iban and bic (with the values from the server)
* decide in the bank data component what case you have (see cat17, you can copy the script part almaost 1:1 from there) and call the right store action depending on case, when the fields are filled.
* modify the `paymentDataIsValid` getter and `MARK_EMPTY_FIELDS_INVALID` mutator to include/ignore the bank data fields depending on payment type. (this needs some unit tests). Maybe it's better to split `MARK_EMPTY_FIELDS_INVALID` into two mutators because the bank data is on the address page. 

Later (e.g. in my PR) we need to remember to call the `payment/markEmptyValuesAsInvalid` action when submitting the address form to make sure that if the user left the bank data fields (partially) empty (and thus no server call has happened and the bank data fields are still in `Validity.INCOMPLETE'  state), we force a validity change.

skins/laika/src/store/payment/actions.ts
```diff
@@ -52,12 +49,12 @@ export const actions = {
 		context.commit( SET_TYPE, payload );
 		context.commit( SET_TYPE_VALIDITY );
 	},
-	[ setAccountId ]( context: ActionContext<Payment, any>, payload: TypeData ): void {
-		context.commit( SET_BANK_ACCOUNT_ID, payload );
-		context.commit( SET_BANK_ACCOUNT_ID_VALIDITY );
+	[ setBankData ]( context: ActionContext<Payment, any>, payload: BankAccountData ): void {
```
gbirke: Since you need to pass the URL, you can't use the BankAccountData as type here. Depending on where that interface is used, I'd either add the URL field and rename the interface (e.g. `BankAccountRequest` or `BankAccountValidationParams`) or define the type inline:

    [ setBankData ]( context: ActionContext<Payment, any>, payload: { validationUrl: string, bankdata: BankAccountData }): void {

skins/laika/src/store/payment/actions.ts
```diff
@@ -52,12 +49,12 @@ export const actions = {
...
+		//TODO Perform AJAX request
+		context.commit( SET_BANKDATA, payload );
 	},
-	[ setBankId ]( context: ActionContext<Payment, any>, payload: TypeData ): void {
-		context.commit( SET_BANK_ID, payload );
-		context.commit( SET_BANK_ID_VALIDITY );
+	[ setLegacyBankdata ]( context: ActionContext<Payment, any>, payload: BankAccountData ): void {
```
gbirke: When passing the URL in, the code in both actions will be identical. Maybe let the BankData component decide which URL to use and have only one store action?

skins/laika/src/components/pages/donation_form/PaymentBankData.vue
```diff
@@ -52,12 +52,31 @@ export default Vue.extend( {
...
-			this.$store.dispatch( action( NS_PAYMENT, setAccountId ), this.$data.accountId );
+		validate() {
+			if ( this.isAccountIdEmpty() ) {
+				return;
+			}
+			if ( this.looksLikeIban() ) {
+				this.$store.dispatch( action( NS_PAYMENT, setBankData ), this.$data );
+				return;
+			}
+			this.$store.dispatch( action( NS_PAYMENT, setLegacyBankdata ), this.$data );
```
gbirke: The `dispatch` method returns a promise. You can use that to update the bank data fields when the server returns a success response (especially relevant for the account number -> BIC conversion).

skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
```diff
@@ -91,10 +91,13 @@ export default Vue.extend( {
 		submit() {
 			( this.$refs.address as any ).validateForm().then( () => {
 				if ( this.$store.getters[ NS_ADDRESS + '/requiredFieldsAreValid' ] ) {
+					if ( this.$store.getters[ NS_PAYMENT + '/isDirectDebitPayment' ] &&
+						!this.$store.getters[ NS_BANKDATA + '/bankDataIsValid' ] ) {
+						return;
+					}
```
gbirke: 👍 
Now that I see this code, it makes total sense for me to conditionally check in the component instead of complicating the store code with conditional and cross-namespace checks. Especially since the `NS_PAYMENT` namespace might become `NS_DONATION_PAYMENT` while NS_BANKDATA will be used both for donation and membership.

skins/laika/src/components/pages/donation_form/SubmitValues.vue
```diff
@@ -4,6 +4,8 @@
 		<input type="hidden" name="zahlweise" :value="payment.type">
 		<input type="hidden" name="periode" :value="payment.interval">
 		<input type="hidden" name="betrag" :value="formattedAmount">
+		<input v-if="isBankData" type="hidden" name="iban" :value="bankdata.iban">
```
gbirke: I think you can keep the code simple and forgo the `if` and `isBankData` property. The backend will ignore the data anyways and if a hidden field is empty, it won't submit.

timEulitz: I was thinking more of a weird case where a user entered valid IBAN data, then went back and changed the payment type and then submitted the form. I think in that case the IBAN data would then send along to the server if it were not for that if. Not sure if that check is really necessary though, I can go back and test it if need be.

gbirke:
I've user-tested it and it works fine. Tried German IBAN, Legacy bank data conversion, Austrian IBAN, going back and switching payment type, going back again and switching payment type again.

timEulitz:
Alright, thanks!

timEulitz:
Should be good now, please review and possibly merge before creating other big PRs because this is getting out of hand :see_no_evil:.

However, two things:

On my local environment at home, the bank data validation seems to fail because it seems to connect to the wrong port which then causes the connection to be blocked:

connection

This was not an issue on my notebook, however since I did two relatively complex rebases, I am not sure if the rebases are at fault or if this issue already existed before that on my desktop. Could you user-test this and see if you have the same issue?

________________________

Secondly, sporadically, when running `npm run test:unit`, this error will pop up but it will not cause the tests to fail:

```
TypeError: Cannot read property 'protocol' of undefined

tests/unit/components/pages/donation_form/BankData.spec.ts

This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 10)
```

I am really not sure what causes it and what makes it even more confusing is that it really occurs randomly - the next time executing the tests it may just vanish. Do either of you have an idea what could cause this? @gbirke @tzhelyazkova

skins/laika/src/view_models/BankAccount.ts
```diff
@@ -0,0 +1,29 @@
...
+    accountId: string
+    bankId: string
+}
+
+export interface BankAccountRequest {
+    validationUrl: string
+    requestParams: object
+}
+
+export interface BankAccountResponse {
```
tzhelyazkova: I'm ok with this type of interface but I think `[key: string]: string` is better for responses coming from the server because this is now tight to how the back end responds. In case we add something there this is going to start failing.

skins/laika/tests/unit/components/pages/donation_form/BankData.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+			mocks: {
+				$t: () => {},
+			},
+		} );
+
+		const iban = wrapper.find( '#iban' );
+		const bic = wrapper.find( '#bic' );
+		wrapper.setData( { accountId: 'AT12345605171238489890', bankId: 'ABCDDEFFXXX' } );
+		expect( ( ( <HTMLInputElement> iban.element ).value ) ).toMatch( 'AT12345605171238489890' );
+		expect( ( ( <HTMLInputElement> bic.element ).value ) ).toMatch( 'ABCDDEFFXXX' );
```
tzhelyazkova: Is this because TypeScript complains? :thinking:

composer.lock
```diff
@@ -1296,33 +1296,37 @@
         },
         {
             "name": "guzzlehttp/psr7",
-            "version": "1.5.2",
+            "version": "1.6.1",
```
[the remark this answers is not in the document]
timEulitz: Let's try to keep it in mind and talk about it after the next daily.

skins/laika/tests/unit/components/pages/donation_form/BankData.spec.ts
timEulitz: Of course, I'm not a typecasting masochist :hocho::face_with_head_bandage:!

I have tried a ton of things but it's simply impossible to get the `value` without first casting to `HTMLInputElement`.

skins/laika/src/view_models/BankAccount.ts
timEulitz: I have updated the interface.

skins/laika/README.md
```diff
@@ -15,21 +15,11 @@ npm run serve
 npm run build
```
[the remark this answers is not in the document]
timEulitz: Yeah, I've deleted it now and long-term we can probably document things in the general README.

composer.lock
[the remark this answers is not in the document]
gbirke: The `composer.lock` for the FundraisingFrontend is essential to have stable deployments and avoid accidentally deploying wrong master of any bounded context dependencies. If the frequent updates to the messages cause merge conflicts (which are easily resolved) that bother us too much, then we should rather think of a different way of incorporating the messages into the dev environment.

skins/laika/src/components/pages/donation_form/PaymentBankData.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+	<fieldset class="has-margin-bottom-36">
+		<legend class="title is-size-5">{{ $t( 'donation_form_payment_bankdata_title' ) }}</legend>
+		<span>{{ $t( 'donation_form_payment_bankdata_legend' ) }}</span>
+		<div v-bind:class="[{ 'is-invalid': bankDataIsInvalid }]">
```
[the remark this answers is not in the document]
gbirke: The `b-radio` <-> Vuex interaction was a very inconvenient and hacky developer experience. I hope this was just due to the nature of `<input type=radio>` elements and not a general flaw of Buefy.

composer.lock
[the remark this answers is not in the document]
tzhelyazkova: The thing is we never commit a `composer.lock` with only the content repo version bumped up. It's always a few things, because it's kind of annoying having to run update and then hand pick only the content repo in the lock file. And all of those random composer updates people run locally tend to collide with each other now and then. I didn't say the merge conflicts are hard to resolve but it'd be nice if we don't have them so often.
I admit  I don't understand the argument that we use the lock file to avoid accidentally deploying the wrong master, so let's discuss it in the next daily.

PR #1537, opened 2019-07-01
Create membership type

tzhelyazkova:
- [x] Setup Membership structure (folders, parent vue component, store)
- [x] Create membership type page

Follow up PRs will cover the rest of the AC from the ticket, meaning adding address fields, birthday field and populating the form in case of coming from the donation confirmation page.

### DIFF
--- a/skins/laika/src/components/pages/membership_form/MembershipType.vue
+++ b/skins/laika/src/components/pages/membership_form/MembershipType.vue
@@ -1,0 +1,48 @@
+<template>
+    <fieldset class="has-margin-top-36">
+        <legend class="title is-size-5">{{ $t('membership_type_legend') }}</legend>
+        <div>
+            <b-radio :class="{ 'is-active': selectedType === 'sustaining' }"
+                    type="radio"
+                    id="sustaining"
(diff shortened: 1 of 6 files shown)


gbirke:
When copying from Donation components, please keep in mind that in our domain language we distinguish between _donation amount_ and _membership fee_. Although in some places, "amount" might still make sense.

skins/laika/src/components/pages/membership_form/MembershipType.vue
```diff
@@ -0,0 +1,54 @@
...
+	},
+} );
+</script>
+<style lang="scss" scoped>
+    .b-radio.radio {
+        margin-left: 0;
+        height: 6.5em;
+    }
+    p {
+        color: rgba(0, 0, 0, 0.6);
```
timEulitz: It's better to use the color variables and auto-generated CSS classes for this. You can assign the `has-text-dark-lighter` CSS class to the `<p>` elements and remove this CSS rule, this will apply `$fun-color-dark-lighter` which is the same as yours.

PR #1538, opened 2019-07-02
Membership address form

tzhelyazkova:

- [x] Create Address parent component for membership and add all the needed address subcomponents to it
- [x] Create membership address store module and connect it to membership address form
- [x] Create Date of Birth component
- [x] Create Receipt Opt Out component

It'd be better if I do this last step in a separate PR as it will require a lot of files being moved to a new directory. That will bump the number of `Files changed` by a lot and will make this PR hard to review.
- [ ] Extract the components used both in donation and membership into a new shared folder

### DIFF
--- a/skins/laika/src/components/pages/donation_form/DateOfBirth.vue
+++ b/skins/laika/src/components/pages/donation_form/DateOfBirth.vue
@@ -1,0 +1,17 @@
+<template>
+    <div>
+
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend( {
+
+} );
+</script>
+
+<style scoped>
+
+</style>
--- a/skins/laika/src/components/pages/donation_form/DateOfBirth.vue
+++ b/skins/laika/src/components/pages/donation_form/DateOfBirth.vue
@@ -1,17 +1,50 @@
 <template>
-    <div>
-
-    </div>
+    <fieldset>
+		<label for="birthDate" class="subtitle has-margin-top-36">{{ $t( 'membership_birth_date_label' ) }}</label>
+		<b-input type="text"
+			id="birthDate"
+			:placeholder="$t( 'membership_birth_date_placeholder' )"
+			v-model="date"
+			@blur="validateDate">
+		</b-input>
+		<span v-if="dateHasError" class="help is-danger">{{ $t( 'membership_birth_date_error' ) }}</span>
+    </fieldset>
 </template>
 
 <script lang="ts">
 import Vue from 'vue';
+import { NS_MEMBERSHIP_ADDRESS } from '@/store/namespaces';
+import { action } from '@/store/util';
+import { setDate } from '@/store/membership_address/actionTypes';
 
 export default Vue.extend( {
-
+	name: 'DateOfBirth',
+	data: function () {
+		return {
+			date: '',
+			datePattern: new RegExp( [
+				'^(?:(?:31(\\.)',
+				'(?:0?[13578]|1[02]))\\1|(?:(?:29|30)(\\.)',
+				'(?:0?[13-9]|1[0-2])\\2))(?:(?:1[6-9]|[2-9]\\d)?\\d{2})$|^(?:29(\\.)',
+				'0?2\\3(?:(?:(?:1[6-9]|[2-9]\\d)?(?:0[48]|[2468][048]|[13579][26])|(?:(?:16|[2468][048]|[3579][26])00))))$',
+				'|^(?:0?[1-9]|1\\d|2[0-8])(\\.)(?:(?:0?[1-9])|(?:1[0-2]))\\4(?:(?:1[6-9]|[2-9]\\d)?\\d{2})$',
+			].join( '' ) ),
+			dateHasError: false,
+		};
+	},
+	methods: {
+		setDate: function () {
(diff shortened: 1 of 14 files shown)


skins/laika/src/components/pages/membership_form/Address.vue
```diff
@@ -0,0 +1,150 @@
...
+		validateForm(): Promise<ValidationResult> {
+			return this.$store.dispatch( action( NS_MEMBERSHIP_ADDRESS, validateAddress ), this.$props.validateAddressUrl );
+		},
+		onFieldChange( fieldName: string ): void {
+			this.$store.dispatch( action( NS_MEMBERSHIP_ADDRESS, setAddressField ), this.formData[ fieldName ] );
+		},
+	},
+} );
+</script>
+<style lang="scss" scoped>
```
timEulitz: I think we should open a discussion about scoped SCSS some time. This creates CSS selectors with data attributes which do not perform that well and it's something we should avoid unless really necessary in my opinion. The SCSS rules here may be better placed in a more general location for better reusability and performance.

skins/laika/src/components/pages/donation_form/DateOfBirth.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+    <fieldset>
+		<label for="birthDate" class="subtitle">{{ $t( 'membership_birth_date_label' ) }}</label>
```
timEulitz: This label needs the class `has-margin-top-36` otherwise it's too close to the element above it.

skins/laika/src/components/pages/membership_form/subpages/AddressPage.vue
```diff
@@ -1,17 +1,20 @@
 <template>
     <div class="column is-full">
         <membership-type></membership-type>
+		<address-fields v-bind="$props" ref="address"></address-fields>
```
timEulitz: TIL `v-bind="$props"`

skins/laika/src/components/pages/membership_form/Address.vue
tzhelyazkova: Yeah this should probably be extracted, it's the same rules for the buttons on the donation form.

timEulitz:
> I'm concerned that duplicating the address namespace makes the code less maintainable. The validation criteria for both are identical (even the URL is the same). It's perfectly valid to do it, since we're also doing it on the backend, but there we don't have 5+ files with lots of code to represent an address.

If it's possible to not repeat the code without creating a messy setup, I would definitely be in favor of having just a single source of truth for addresses.

skins/laika/src/components/pages/donation_form/DateOfBirth.vue
```diff
@@ -0,0 +1,50 @@
...
+	name: 'DateOfBirth',
+	data: function () {
+		return {
+			date: '',
+			datePattern: new RegExp( [
+				'^(?:(?:31(\\.)',
+				'(?:0?[13578]|1[02]))\\1|(?:(?:29|30)(\\.)',
+				'(?:0?[13-9]|1[0-2])\\2))(?:(?:1[6-9]|[2-9]\\d)?\\d{2})$|^(?:29(\\.)',
+				'0?2\\3(?:(?:(?:1[6-9]|[2-9]\\d)?(?:0[48]|[2468][048]|[13579][26])|(?:(?:16|[2468][048]|[3579][26])00))))$',
+				'|^(?:0?[1-9]|1\\d|2[0-8])(\\.)(?:(?:0?[1-9])|(?:1[0-2]))\\4(?:(?:1[6-9]|[2-9]\\d)?\\d{2})$',
```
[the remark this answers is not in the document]
tzhelyazkova: The string templating will require quite some time to figure out correctly, I don't want to put so much effort into this regexp. I simply copied this one from SO.
If in the future we have English translations and have to allow English formatting of the date we can get rid of this specific regexp and create a new basic one (that just checks numbers or something).

tzhelyazkova:
> I'm concerned that duplicating the address namespace makes the code less maintainable. The validation criteria for both are identical (even the URL is the same). It's perfectly valid to do it, since we're also doing it on the backend, but there we don't have 5+ files with lots of code to represent an address.
> 
> Other solutions would be to nest namespaces (NS_MEMBERSHIP/NS_ADDRESS, NS_MEMBERSHIP/NS_MEMBER) or to introduce specific namespaces (NS_DONATION, NS_MEMBERSHIP) for the parts that differ. What do you think, @tzhelyazkova , @timEulitz ? Do we gain maintainability from being DRY? Do we get better separation of concerns of the different domains?

In my head separation of concerns is better (clearly having membership modules vs donation modules). I do agree we are copying most of the logic and there are many files, but that's just how we chose to structure our code. We could've put everything (state, getters, mutations, actions ) in one file. Also TS adds some unavoidable files (actionTypes, mutationTypes). I don't immediately see the benefit of reusing one store for both cases. Maybe I'll come to your opinion when I start moving the common vue components and extract some store related logic. 

My other concern is that we can do this for address, but probably not for payment? And that's kind of inconsistent. I remember being confused by the donation/membership store in `cat17`, therefor I'm trying to make the domain in the front end as clear as possible which includes separating contexts clearly. 

I might be wrong about all this :woman_shrugging: but lets merge this in the current form and when I put up a PR with a `shared` folder we can revisit the topic. :ok: :question:

PR #1547, opened 2019-07-08
Add membership fee store

gbirke:
Add membership fee store. It is very similar to `payment`, but with
different defaults (payment type preselected and valid because we don't
have a payment type selection until
https://phabricator.wikimedia.org/T147400 is active) and no
initialization (beause that doesn't make sense in the context of
memberships).
Add setting interval validity because membership has no default
interval.


skins/laika/src/store/membership_fee/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export default function (): Module<Payment, any> {
+	const state: Payment = {
+		validity: {
+			fee: Validity.INCOMPLETE,
+			interval: Validity.INCOMPLETE,
+			type: Validity.VALID,
+		},
+		values: {
+			fee: '', // membership fee in cents
```
tzhelyazkova: how about we name it `feeInCents`?

skins/laika/src/store/membership_fee/index.ts
```diff
@@ -0,0 +1,31 @@
+import { Module } from 'vuex';
+import { Payment } from '@/view_models/Payment';
+import { Validity } from '@/view_models/Validity';
+import { actions } from '@/store/payment/actions';
+import { getters } from '@/store/payment/getters';
+import { mutations } from '@/store/payment/mutations';
+
```
tzhelyazkova: all these paths should be `membership_fee` instead of `payment`

skins/laika/tests/unit/store/membership_fee_store.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+			};
+			const action = actions.markEmptyValuesAsInvalid as any;
+			action( context );
+			expect( context.commit ).toBeCalledWith(
+				'MARK_EMPTY_FIELDS_INVALID'
+			);
+		} );
+	} );
+
+	describe( 'Actions/setInterval', () => {
```
tzhelyazkova: I think the two tests here can be combined into one like in 
https://github.com/wmde/FundraisingFrontend/blob/wmde19/skins/laika/tests/unit/store/bankdata_store.spec.ts#L156

skins/laika/src/store/membership_fee/index.ts
gbirke: Thanks for pointing it out, I also like to use code instead of comments. I won't update it now since I'll be working on the banner, but feel free to change it in subsequent PRs.

PR #1548, opened 2019-07-08
Missing unit tests

tzhelyazkova:
This PR adds some missing unit tests.
There are some more which are missing but since other things seem to have gotten priority let's merge this and continue with the rest (Address component and donation address store) in a different PR

### DIFF
--- a/skins/laika/tests/unit/components/pages/donation_form/AddressType.spec.ts
+++ b/skins/laika/tests/unit/components/pages/donation_form/AddressType.spec.ts
@@ -1,0 +1,24 @@
+import { mount, createLocalVue } from '@vue/test-utils';
+import AddressType from '@/components/pages/donation_form/AddressType.vue';
+import Buefy from 'buefy';
+
+const localVue = createLocalVue();
+localVue.use( Buefy );
+
+describe( 'AddressType.vue', () => {
+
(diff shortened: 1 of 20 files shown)


skins/laika/tests/unit/components/pages/donation_form/AddressType.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+				},
+			} ),
+			event = 'address-type',
+			company = wrapper.find( '#company' );
+		company.trigger( 'change' );
+		const person = wrapper.find( '#personal' );
+		person.trigger( 'change' );
+		const anon = wrapper.find( '#anonymous' );
+		anon.trigger( 'change' );
+		expect( wrapper.emitted( event ) ).toHaveLength( 3 );
```
gbirke: You _could_ check if the events submitted the right type data, but having at least the basic event handling checked is better than nothing.

tzhelyazkova: Good that you mentioned that, because when I added those assertions I realized triggering a change event doesn't work for radio buttons. You have to trigger a click event, even though the component is triggering a change 🤷‍♀

PR #1549, opened 2019-07-08
Add minimumAmount property to AmountSelection

gbirke:
This is for the membership page, where some choices need to be inactive
if the interval*amount is below the yearly minimum amount.

The server does the actual validation, minimumAmount is a property that
only affects the component behavior/display.


gbirke:
> Shouldn't this PR also contain the updated `Adress.vue` components in donation and membership where this property is passed? 🤔

It was a small change I could do while waiting for the review on #1544  
I'll create the membership address component and then cherry-pick this commit and close the PR.

The payment component in donation needs no change because `minimumAmount` has a default value of 0.

PR #1550, opened 2019-07-11
Add missing membership address form specifics

tzhelyazkova:

- [x] Choosing "Firma" disables the membership type option "Aktive Mitgliedschaft".
- [x] Passing the membership type in the URL (type=sustaining) sets the given type and hides the membership type options.
- [x] Populate address fields if the address data is available on render
- [x] Populate email and pre-set address type if available on render

### DIFF
--- a/app/Controllers/ApplyForMembershipController.php
+++ b/app/Controllers/ApplyForMembershipController.php
@@ -36,10 +36,7 @@ class ApplyForMembershipController {
 		$params = [
 			'urls' => Routes::getNamedRouteUrls( $ffFactory->getUrlGenerator() )
 		];
-
-		if ( $httpRequest->query->get( 'type' ) === 'sustaining' ) {
-			$params['showMembershipTypeOption'] = false;
-		}
+		$params['showMembershipTypeOption'] = ( $httpRequest->query->get( 'type' ) === 'sustaining' ? false : true );
 
 		$useCase = $ffFactory->newGetDonationUseCase( $httpRequest->query->get( 'donationAccessToken', '' ) );
 		$responseModel = $useCase->showConfirmation(
--- a/app/Controllers/ApplyForMembershipController.php
+++ b/app/Controllers/ApplyForMembershipController.php
@@ -36,7 +36,7 @@ class ApplyForMembershipController {
 		$params = [
 			'urls' => Routes::getNamedRouteUrls( $ffFactory->getUrlGenerator() )
 		];
-		$params['showMembershipTypeOption'] = ( $httpRequest->query->get( 'type' ) === 'sustaining' ? false : true );
+		$params['showMembershipTypeOption'] = $httpRequest->query->get( 'type' ) !== 'sustaining';
 
 		$useCase = $ffFactory->newGetDonationUseCase( $httpRequest->query->get( 'donationAccessToken', '' ) );
 		$responseModel = $useCase->showConfirmation(
--- a/skins/laika/src/components/pages/membership_form/Address.vue
+++ b/skins/laika/src/components/pages/membership_form/Address.vue
@@ -2,7 +2,7 @@
 	<div id="addressForm" class="column is-full">
 		<div class="has-margin-top-18">
 			<h1 class="title is-size-1">{{ $t( 'donation_form_section_address_title' ) }}</h1>
-			<address-type :initial-value="addressTypeFromName(initialFormValues.addressType)" v-on:address-type="setAddressType( $event )"></address-type>
+			<address-type :initial-value="initialFormValues.addressType | addressTypeFromName" v-on:address-type="setAddressType( $event )"></address-type>
 		</div>
 		<h1 class="has-margin-top-36 title is-size-5">{{ $t( 'donation_form_section_address_title' ) }}</h1>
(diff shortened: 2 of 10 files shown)


gbirke:
A general comment: the data for the last 2 TODOs is available in the template (`applicationVars` in the entry point):
`showMembershipTypeOption` to decide if the membership type should be shown
`initialFormValues` contains the address data.

skins/laika/src/components/pages/membership_form/MembershipType.vue
```diff
@@ -46,6 +49,22 @@ export default Vue.extend( {
...
+			get: function () {
+				return AddressTypeModel;
+			},
+		},
+		...mapGetters( NS_MEMBERSHIP_ADDRESS, [ 'addressType' ] ),
+		isActiveTypeDisabled: {
+			get: function () {
+				if ( this.$store.getters[ NS_MEMBERSHIP_ADDRESS + '/addressType' ] === AddressTypeModel.COMPANY ) {
+					this.$data.selectedType = MembershipTypeModel.SUSTAINING;
+					this.setType();
```
gbirke: Please don't do a computed property with side effects. At least in my understanding of Vue, computed properties should only return values. Also, they are cached, so we can only kind-of rely on the change.

A better (or at least more explicit) approach IMHO would be to create a `mounted` handler where you add a watcher to the store. See https://dev.to/viniciuskneves/watch-for-vuex-state-changes-2mgj#vuex-watch for a code example and explanation

app/Controllers/ApplyForMembershipController.php
```diff
@@ -36,10 +36,7 @@ public function showApplicationForm( FunFunFactory $ffFactory, Request $httpRequ
 		$params = [
 			'urls' => Routes::getNamedRouteUrls( $ffFactory->getUrlGenerator() )
 		];
-
-		if ( $httpRequest->query->get( 'type' ) === 'sustaining' ) {
-			$params['showMembershipTypeOption'] = false;
-		}
+		$params['showMembershipTypeOption'] = ( $httpRequest->query->get( 'type' ) === 'sustaining' ? false : true );
```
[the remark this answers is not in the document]
tzhelyazkova: oops right, that was stupid :sweat_smile:

skins/laika/src/components/pages/membership_form/Address.vue
```diff
@@ -116,10 +117,23 @@ export default Vue.extend( {
...
 		] ),
 	},
+	mounted() {
+		const initialFormValues = this.$props.initialFormValues;
+		if ( initialFormValues !== undefined ) {
+			Object.keys( this.$data.formData ).forEach( ( fieldName: string ) => {
+				this.formData[ fieldName ].value = initialFormValues[ fieldName ];
+			} );
+		}
+	},
```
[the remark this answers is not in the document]
tzhelyazkova: Writing a mounted function was more convenient than going through the entire `formData` object and assigning the appropriate `initialFormValue`. Plus I'd always have to check if the object is undefined, so `formData` becomes even hareder to read. 
Since we need to populate the form only in case of the donor coming from the donation confirmation page I thought this is easier and more isolated.
`AddressType` has only one data attribute so it's not hard to read. Either way is fine there.

skins/laika/src/components/pages/membership_form/MembershipType.vue
tzhelyazkova: Ok I see how having this additional logic in the getter does not look nice, but I don't think computed properties are meant to only return values. We can have a setter as part of a computed property where you'd make some changes based on what you called "side effects" and that will affect the getter. I committed a new version, where the logic is split between a getter and a setter. 
I thought a watcher's main usage is when you have async logic and we don't really have that here. IMHO it seems a bit too complex of a solution for the case that we have. It's *just* making an option disabled based on the value of another option.
So how about the setter solution?

gbirke: The problem with the setter is that it's not called anywhere (e.g. `isActiveTypeDisabled = false` in the template or `this.$props.isActiveTypeDisabled = false` in the code). But it can't be triggered by user interaction in this component, only by store changes, so using a setter here doesn't work.

The functionality we want to have (and test) is "When the address type in the store changes to 'company' and the selected type is `active`, then unselect the selected type." (or alternatively, depending on UX feedback  select "sustaining", although that would subvert the user choice, possibly out of the view of the user and thus not validated). 

> I thought a watcher's main usage is when you have async logic and we don't really have that here.

Watchers (in this case store watchers) are IMHO a good way to tackle the "when a store value changes, change `this.$data?" scenario in an event-based way. The article I linked to is the first result when I google for "vue change data when vuex changes"

>  I don't think computed properties are meant to only return values.

We're not using ESLint rules for Vue files, but among the plugin rules is one that says that computed properties should not have side effects: https://eslint.vuejs.org/rules/no-side-effects-in-computed-properties.html 
I take that as an indicator that computed properties should be pure functions (i.e. functions without side effects).

If we did not use a Buefy radio component, we could do away with the local state altogether and use a computed property and events, but unfortunately that pattern does not work well with the Buefy radio component which insists on having a v-model.

PR #1551, opened 2019-07-15
Fix validation issue with optional form fields

timEulitz:
Currently, if a user proceeds with the address page without having filled in all data, the optional `title` form field is marked as invalid instead of being ignored. Since the error will not go away, it might cause some confusion for users or cause them to be stuck.

I think the change I propose here should fix this issue but I just wanted to separate it out because I was not 100% sure.

### DIFF
--- a/skins/laika/src/store/address/mutations.ts
+++ b/skins/laika/src/store/address/mutations.ts
@@ -19,7 +19,7 @@ import { REQUIRED_FIELDS } from '@/store/address/constants';
 export const mutations: MutationTree<AddressState> = {
 	[ VALIDATE_INPUT ]( state: AddressState, field: InputField ) {
 		if ( field.value === '' && field.optionalField ) {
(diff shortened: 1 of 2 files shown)


skins/laika/src/store/address/mutations.ts
```diff
@@ -19,7 +19,7 @@ import { REQUIRED_FIELDS } from '@/store/address/constants';
 export const mutations: MutationTree<AddressState> = {
 	[ VALIDATE_INPUT ]( state: AddressState, field: InputField ) {
 		if ( field.value === '' && field.optionalField ) {
-			state.validity[ field.name ] = Validity.INCOMPLETE;
+			state.validity[ field.name ] = Validity.VALID;
```
tzhelyazkova: hmmm, this does solve the problem but now I'm thinking our logic is confusing.
There is a constant for required fields used in store modules and there's the property `optionaldField` used in vue components. Not cool.
Anyway that's a topic for another PR.

PR #1552, opened 2019-07-15
Render membership payment form

timEulitz:


skins/laika/src/components/shared/AmountSelection.vue
```diff
@@ -52,6 +53,7 @@ export default Vue.extend( {
 			type: String,
 			default: '',
 		},
+		caption: String,
```
gbirke: This might create an error if you leave caption undefined in the parent component. Maybe add a default value.

skins/laika/src/components/pages/membership_form/subpages/PaymentPage.vue
```diff
@@ -1,14 +1,52 @@
...
+		Payment,
+	},
+	props: {
+		validateAmountUrl: String,
+		paymentAmounts: Array as () => Array<String>,
+		paymentIntervals: Array as () => Array<Number>,
+	},
+	methods: {
+		submit() {
+			( this.$refs.address as any ).validateForm().then( () => {
```
gbirke: wrong ref name and missing `ref` attribute in the template. Also, `Payment` component has no  `validateForm ` method yet.

timEulitz:
Feature is not done but I would like to get the progress merged at this point because it's become somewhat stable.

skins/laika/src/components/Question.vue
```diff
@@ -51,3 +51,25 @@ export default Vue.extend( {
 	},
 } );
 </script>
+
+<style lang="scss">
+@import "../scss/custom";
```
tzhelyazkova: why is this part of this PR?

skins/laika/src/components/pages/membership_form/AddressType.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<fieldset>
-		<legend class="title is-size-5">{{ $t( 'membership_section_address_header_type' ) }}</legend>
```
tzhelyazkova: I named all these without `form` in there on purpose but ok..

skins/laika/src/components/pages/membership_form/Payment.vue
```diff
@@ -0,0 +1,72 @@
...
+			interval: ( state: any ) => state[ NS_MEMBERSHIP_FEE ].values.interval,
+		} ),
+		...mapGetters( NS_MEMBERSHIP_FEE, {
+			feeIsValid: 'feeIsValid',
+			typeIsValid: 'typeIsValid',
+			minimumAmount: 'minimumAmount',
+		} ),
+		getMinimumAmount: {
+			get: function () {
+				return ( this as any ).minimumAmount( this.$store.getters[ NS_MEMBERSHIP_ADDRESS + '/addressType' ] );
```
tzhelyazkova: :middle_finger: for TypeScript

skins/laika/src/components/pages/membership_form/subpages/AddressPage.vue
```diff
@@ -24,5 +34,15 @@ export default Vue.extend( {
 		showMembershipTypeOption: Boolean,
 		initialFormValues: [ Object, String ],
 	},
+	methods: {
+		next() {
+			( this.$refs.address as any ).validateForm().then( () => {
+				if ( this.$store.getters[ NS_MEMBERSHIP_ADDRESS + '/requiredFieldsAreValid' ] ) {
+					this.$emit( 'next-page' );
+					return;
```
tzhelyazkova: you don't need this

skins/laika/src/store/membership_fee/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const MEMBERSHIP_MINIMUM_FEE_PERSONAL = 2400; // In cents
```
tzhelyazkova: It's better to put the comment in the name of the constant.

skins/laika/src/store/membership_fee/getters.ts
```diff
@@ -17,4 +19,23 @@ export const getters: GetterTree<MembershipFee, any> = {
...
+		if ( isNaN( interval ) ) {
+			return 0;
+		}
+		const yearlyIntervalMultiplier = interval / 12;
+		if ( addressType === AddressTypeModel.PERSON ) {
+			return MEMBERSHIP_MINIMUM_FEE_PERSONAL * yearlyIntervalMultiplier;
+		}
+		return MEMBERSHIP_MINIMUM_FEE_COMPANY * yearlyIntervalMultiplier;
+	},
+	yearlyAmount: function ( state: MembershipFee ) {
```
tzhelyazkova: return type is missing

gbirke:
Missing parts:

- [ ] Unit test for Payment passing on the right minimumAmount to amountSelection
- [ ] `setInterval` store action should also trigger fee validation (+ unit testing that behavior). Or alternatively, validate separately (see below).
- [ ] Bank data component

We could make the validation much simpler and less "magic" if we don't do the validation in the background. `setInterval` and `setFee` would be "dumb" setters that don't trigger validation and the validation would be a separate action that is triggered on submit - similar on what we do on the last page of the donation form.

skins/laika/src/components/Question.vue
timEulitz: The CSS was in the wrong place before and I moved it when adding my own. Definitely unrelated, yes :bowing_man:.

skins/laika/src/components/pages/membership_form/AddressType.vue
timEulitz: This way it seemed more consistent with how it was done for the donation_form component strings. I mainly wanted to add consistency here, but I see why your naming makes sense too.

PR #1556, opened 2019-07-17
Add submit values to membership form

gbirke:
This is for https://phabricator.wikimedia.org/T226930


gbirke:
Ah yes, I forgot to validate the payment data on submit. Code is now in and mostly works (if you wait for the bank data to validate before you hit "submit"). I'll create a ticket for the race condition.

PR #1559, opened 2019-07-18
Implement logic between interval and amount in membership form

timEulitz:


skins/laika/src/store/membership_fee/actions.ts
```diff
@@ -22,7 +22,34 @@ import {
 } from '@/store/membership_fee/mutationTypes';
 import { ValidationResponse } from '@/store/ValidationResponse';
 import { Validity } from '@/view_models/Validity';
-import { addressTypeName } from '@/view_models/AddressTypeModel';
+import { AddressTypeModel, addressTypeName } from '@/view_models/AddressTypeModel';
+
+function isNonNumeric( value: string ) {
```
tzhelyazkova: this function is a helper and should be in `utils.ts`

skins/laika/src/store/membership_fee/actions.ts
```diff
@@ -22,7 +22,34 @@ import {
...
+		context.commit( SET_FEE_VALIDITY, validity );
+	} );
+}
+
+function validateFeeDataRemotely( context: ActionContext<MembershipFee, any>, validateFeeUrl: string, feeValue: string, interval: string ) {
+	const feeAmount = ( Number( feeValue ) / 100 ).toFixed( 2 );
+	const paymentInterval = interval;
+	const addressType = context.rootState.membership_address.addressType;
+	postFeeData( context, validateFeeUrl, feeAmount, paymentInterval, addressType );
+}
```
tzhelyazkova: I personally don't like writing "plain" functions in the `actions` file. I understand `setFee` is too long but this goes against Vuex's model. Those two functions are outside the actions constant and you are committing mutations from them so it's kind of iffy.
I guess the right way to go will be to extract this in another module which handles api calls. I know this adds more effort though. @gbirke what do you think?

gbirke: I'm not that allergic to have some unexported utility functions in a Vuex file - I see them like private methods of a class, helping to make the code more readable and differentiating between public and private API.

In this case however, having all axios-related functionality in a different file (I guess that's what you mean by "module") sounds good, as it at least allows a semblance of separation of concerns (although by importing the functions we still haven't gained a proper dependency injection, but that's another debate).

What would be an appropriate name for the new file? `server_communication`? `server_api`? `api_calls`? 

Insstead of the context, `validateFeeDataRemotely` should get the addressType as a parameter. and it should return a promise, so you can move back the commits from `postFeeData` into `setFee` and `setInterval`.

tzhelyazkova: I meant a proper store module dedicated to handling the api calls. 
It's very well described here https://codeburst.io/composing-actions-with-vuex-b63466264a37 in the section `Composing actions`.
As to the name, both `api_calls` or `server_communication` sound good to me.

gbirke: Thank you for the link! Wrapping axios in a module sounds tempting, but I'd rather not do it because

- We'd either create an inconsistency in our code or pile 3-5 more points of effort on an already-underestimated 5-point ticket.
- I'm unsure if your code would become more concise or readable if we want to keep our data typed with TypeScript: The payload for the "api/post" action would have to be very generic: URL and POST params as a key-value object. The calling action would have to build the param object from an existing type, e.g. converting from `AddressState` to the key-value object (which we are already doing in the action). 

If you still want the "no unexported functions in a Vuex file" convention, I'd say we can move it out to its own file.

PR #1561, opened 2019-07-18
Initialize membership address and bank data from donation

gbirke:


gbirke:
Added code for initializing bank data.

PR #1564, opened 2019-07-19
Bank Data in Membership

tzhelyazkova:
Feature: [T226927](https://phabricator.wikimedia.org/T226927)


timEulitz:
@gbirke When user testing, `validateBankDataUrl` and `validateLegacyBankDataUrl` are both undefined on the membership application form, they're only defined on the donation form. They're probably not being passed properly from PHP but I cannot remember the correct place for this from the top of my head.

gbirke:
Are you sure you tested my force-pushed version? This was the only thing that I fixed.

timEulitz:
@gbirke Did `git fetch` before you had pushed and wrongly assumed it was up-to-date. Works fine now, thanks.

PR #1566, opened 2019-07-22
Fix radio button custom styling after type removal

timEulitz:
Follow up to: https://github.com/wmde/FundraisingFrontend/pull/1563

As noted by Tonina, the radio button design changed by removing the type. Turns out Buefy adds the type as a class to one of the elements which was used in one of our CSS selectors. Removing the type disabled the custom CSS rule and this PR fixes that.

### DIFF
--- a/skins/laika/src/scss/overrides.scss
+++ b/skins/laika/src/scss/overrides.scss
@@ -30,7 +30,7 @@
 			border-bottom: 2px solid $fun-color-primary;
 		}
 	}
-	.check.radio {
+	.check {
 		font-size: 12px;
 	}
 }


timEulitz:
This is still an issue in Buefy, supposedly this was resolved in one of the recent releases but it clearly was not. Font rendering differs per browser and operating system and the radio buttons look slightly different on each browser and it even depends on the specific radio buttons on a page. On Firefox, Chrome and Edge on Windows, they look broken with the current design with me and this font size change does not make it perfect but it makes it a bit better. Not sure what we should do here, we can also leave it but then I would completely remove that CSS rule.

PR #1568, opened 2019-07-22
Miscellaneous bug fixes

timEulitz:
https://phabricator.wikimedia.org/T228623

- [x] Prevent donating without an address when using direct-debit payment
- [x] Error message for missing salutation is displayed below title field
- [x] Move e-mail address field into address section
- [x] Bank data error message is displayed too early when using account number and bank code
_Note: Improved some other parts of the logic, especially with invalid user input / clearing previous input_
- [x] Buttons in summary overlap in viewport widths between 768px and 1024px
- [x] Summary should display "5 Euro" instead of "5€"
- [x] No error message when email address is missing
- [x] Remove sentence "Wenn Sie keine Spendenbescheinigung brauchen, ist die Angabe der Adresse optional" below donation receipt out-out
- [x] Clickable labels for newsletter opt-in and donation receipt opt-out checkboxes


skins/laika/src/components/pages/donation_form/AddressType.vue
```diff
@@ -41,13 +42,28 @@ export default Vue.extend( {
...
+		isDirectDebit: Boolean,
+	},
 	computed: {
 		AddressTypeModel: {
 			get: function () {
 				return AddressTypeModel;
 			},
 		},
 	},
+	watch: {
```
timEulitz: Don't know if this was the right approach.

skins/laika/src/components/pages/membership_form/Address.vue
```diff
@@ -125,11 +125,11 @@ export default Vue.extend( {
 	},
 	mounted() {
 		const initialFormValues = this.$props.initialFormValues;
-		if ( initialFormValues !== undefined ) {
-			Object.keys( this.$data.formData ).forEach( ( fieldName: string ) => {
+		Object.keys( this.$data.formData ).forEach( ( fieldName: string ) => {
+			if ( initialFormValues[ fieldName ] !== undefined ) {
```
timEulitz: This was not in the ticket but previously, if no form data was supplied, all form fields were overwritten with "undefined" which caused the default country value to be empty instead of 'DE'. Maybe double-check that this is the best approach here.

gbirke: This is solved in #1561

skins/laika/src/components/pages/donation_form/AddressType.vue
gbirke: It's a bit weird that the relatively self-contained address type now knows about payment types (at least at a semantic level). Could you have a property `disabledTypes` instead, that is an empty array by default. Then the parent (`Address`) component would be responsible to generate the right types (empty array or array with `AddressTypeModel.ANON`). This more generic approach will help us when we tackle the new "EMAIL" address type (https://phabricator.wikimedia.org/T220367)

Now that I think of it, this kind of "business logic" probably belongs in a store getter.

skins/laika/src/components/pages/donation_form/AddressType.vue
```diff
@@ -41,13 +42,29 @@ export default Vue.extend( {
...
+		disabledAddressTypes:
+		{
+			handler: function ( disabledAddressTypes ) {
+				const $this = ( this as any );
```
timEulitz: I don't get why TypeScript randomly starts losing its mind with this just because I refactored this function slightly. This was my workaround to get it to work but no clue if it's any good @gbirke.

skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
```diff
@@ -8,15 +8,15 @@
 				</donation-summary>
 				<submit-values :tracking-data="{}"></submit-values>
 				<div class="level has-margin-top-36">
-					<div class="level-left">
+					<div class="level-item">
```
gbirke: Is the `level` type the still the right one, if we no longer distinguish between right and left. Wouldn't the `column` classes be more appropriate?

skins/laika/src/components/pages/membership_form/Address.vue
```diff
@@ -9,10 +9,7 @@
 		<postal :show-error="fieldErrors" :form-data="formData" :countries="countries" v-on:field-changed="onFieldChange"></postal>
 		<date-of-birth/>
         <receipt-opt-out v-on:opted-out="setReceiptOptedOut( $event )"/>
-        <div class="has-margin-top-36">
-			<h1 class="title is-size-1">{{ $t( 'donation_form_section_email_title' ) }}</h1>
-			<email :initial-value="email" v-on:email="setEmail( $event )"/>
-		</div>
+		<email :initial-value="email" :show-error="fieldErrors" v-on:email="setEmail( $event )"/>
```
gbirke: I'd put it one further up, so the two checkboxes are together.

skins/laika/src/components/shared/Email.vue
```diff
@@ -1,18 +1,19 @@
...
+    <fieldset class="has-margin-top-18">
 		<label for="email" class="subtitle">{{ $t( 'donation_form_email_label' ) }}</label>
 		<b-input type="text"
 			id="email"
 			:placeholder="$t( 'donation_form_email_placeholder' )"
 			v-model="emailValue"
 			@blur="validateEmail">
 		</b-input>
-		<span v-if="emailHasError" class="help is-danger">{{ $t( 'donation_form_email_error' ) }}</span>
+		<span v-if="showError.email || emailHasError" class="help is-danger">{{ $t( 'donation_form_email_error' ) }}</span>
```
gbirke: Hmm, this "I validate the email myself, but my validation can be overwritten from the outside" behavior looks strange to me. Do we need an email field that does standalone validation?

skins/laika/src/components/shared/Email.vue
```diff
@@ -25,6 +26,7 @@ export default Vue.extend( {
 	},
 	props: {
 		initialValue: String,
+		showError: Object as () => AddressValidity,
```
gbirke: This should be a boolean field. Better do `fieldErrors.email` in the Address component.

skins/laika/src/components/pages/donation_form/Address.vue
```diff
@@ -98,6 +95,12 @@ export default Vue.extend( {
 					pattern: '',
 					optionalField: false,
 				},
+				email: {
+					name: 'email',
+					value: '',
+					pattern: '^[^@]+@.+$',
+					optionalField: false,
+				},
```
gbirke: You could take the opportunity and get rid of the `setEmail` action in the store and use the `setAddressField` action instead. Besides more concise code, that has the benefit that the email will be validated with the pattern.

timEulitz:
>User-testing the form (and looking at the code) showed that the newsletter opt-out label is still not clickable

I have taken care of that in this PR so ignore this for the scope of this PR: https://github.com/wmde/FundraisingFrontend/pull/1569

skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
timEulitz: Yeah, makes sense.

skins/laika/src/components/pages/membership_form/Address.vue
timEulitz: Agreed.

skins/laika/src/components/shared/Email.vue
timEulitz: Yeah, I thought the same but I had issues implementing it properly. I will give it another go.

timEulitz:
Due to all the refactoring, I have touched a lot of files, let's try to merge this first before branching off of wmde19 to prevent unnecessary merge conflicts.

PR #1571, opened 2019-07-24
Track async validation

gbirke:

PR #1574, opened 2019-07-25
Show async validation

gbirke:
Show spinners on button while validation is taking place.


gbirke:
This is just a proof-of-concept to show some feedback to the user and get an opinion from @jdittrich 

This is based in #1571, which should be merged first.

PR #1580, opened 2019-07-30
Implement donation / membership cancellation page

timEulitz:
https://phabricator.wikimedia.org/T226958

### DIFF
--- a/skins/laika/src/components/pages/DonationCancellation.vue
+++ b/skins/laika/src/components/pages/DonationCancellation.vue
@@ -1,0 +1,22 @@
+<template>
+	<div class="column is-full">
+		<h1 class="title">{{ $t('donation_cancellation_title') }}</h1>
+		<span v-html="$t( 'donation_cancellation_text', { donation_id: cancellationData.donationId, } )"></span>
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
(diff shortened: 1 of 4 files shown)


skins/laika/src/components/pages/DonationCancellation.vue
```diff
@@ -0,0 +1,22 @@
...
+export default Vue.extend( {
+	name: 'DonationCancellation',
+	props: [
+		'cancellationData',
+	],
+} );
+</script>
+
+<style lang="scss">
+	@import "../../scss/custom";
```
tzhelyazkova: doesn't look like you need any custom css

skins/laika/src/components/pages/MembershipCancellation.vue
```diff
@@ -0,0 +1,22 @@
...
+export default Vue.extend( {
+	name: 'DonationCancellation',
+	props: [
+		'cancellationData',
+	],
+} );
+</script>
+
+<style lang="scss">
+	@import "../../scss/custom";
```
tzhelyazkova: here as well

timEulitz:
> I think we should drop "application" from all membership things.

I see why it's there but yeah, seems like an old carry over from ancient times. I have dropped it from the file names but left it in other places for now because that would be a discussion on its own again I would say.

PR #1582, opened 2019-07-31
Remove all scoped CSS blocks for performance gain

timEulitz:
The uncompressed CSS file size was reduced from 2.6 Megabytes to 450 Kilobytes by removing all scoped slots. Some of the blocks were actually not needed at all and the others could be refactored by simply using CSS selectors with classes from the parent elements.

### DIFF
--- a/skins/laika/src/components/pages/membership_form/Address.vue
+++ b/skins/laika/src/components/pages/membership_form/Address.vue
@@ -152,19 +152,3 @@ export default Vue.extend( {
 	},
 } );
 </script>
-<style lang="scss" scoped>
-    @import "../../../scss/custom";
-
-    button.is-main {
-        height: 54px;
-        font-size: 1em;
-        font-weight: bold;
-        width: 250px;
-        border-radius: 0;
-    }
-    @include until($tablet) {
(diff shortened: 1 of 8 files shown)


skins/laika/src/components/pages/membership_form/Address.vue
```diff
@@ -152,19 +152,3 @@ export default Vue.extend( {
...
 } );
 </script>
-<style lang="scss" scoped>
-    @import "../../../scss/custom";
-
-    button.is-main {
-        height: 54px;
-        font-size: 1em;
-        font-weight: bold;
-        width: 250px;
```
tzhelyazkova: the only difference is that in buefy the width is 270px, I guess it doesn't matter that much.

timEulitz: The CSS selector for main buttons is in `skins/laika/src/scss/overrides.scss` already where it is set to 270px. This scoped rule were not being used anyway because the file does not actually include a button, at least that's how I understand the `scoped` property. I double-checked and the buttons look the same before and after on all resolutions.

PR #1590, opened 2019-08-05
Magical evil css of doom

tzhelyazkova:
This branch has existed for a long time.
Opening a PR for us to decide if we want this change or not.


tzhelyazkova:
Noone protested for 24h so I'm closing this.

PR #1597, opened 2019-08-07
Implement contact form

timEulitz:
https://phabricator.wikimedia.org/T224080

### DIFF
--- a/skins/laika/src/components/pages/Contact.vue
+++ b/skins/laika/src/components/pages/Contact.vue
@@ -3,7 +3,7 @@
 		<h1 class="title">{{ $t( 'contact_form_title' ) }}</h1>
 		<span class="help is-danger has-padding-bottom-18" v-if="contactData.errors">{{ $t('contact_form_error') }}</span>
 		<span class="help is-danger has-padding-bottom-18" v-for="error in contactData.errors">{{ $t( error ) }}</span>
-		<form method="post" action="/contact/get-in-touch" @submit="submit" id="laika-contact">
+		<form method="post" action="/contact/get-in-touch" v-on:submit.prevent="submit" id="laika-contact" ref="form">
 			<fieldset>
 				<div>
 					<label for="firstname" class="subtitle">
@@ -15,10 +15,10 @@
 								id="firstname"
 								name="firstname"
 								:placeholder="$t( 'contact_form_firstname_placeholder' )"
-								v-model="formData.name.value">
+								v-model="formData.firstname.value">
 						</b-input>
 					</b-field>
-					<span v-if="formData.name.validity === Validity.INVALID" class="help is-danger">{{ $t( 'contact_form_firstname_error' ) }}</span>
+					<span v-if="formData.firstname.validity === Validity.INVALID" class="help is-danger">{{ $t( 'contact_form_firstname_error' ) }}</span>
 				</div>
 				<div class="has-margin-top-18">
 					<label for="lastname" class="subtitle">
@@ -30,10 +30,10 @@
 								id="lastname"
 								name="lastname"
 								:placeholder="$t( 'contact_form_lastname_placeholder' )"
-								v-model="formData.surname.value">
+								v-model="formData.lastname.value">
 						</b-input>
 					</b-field>
-					<span v-if="formData.surname.validity === Validity.INVALID" class="help is-danger">{{ $t( 'contact_form_lastname_error' ) }}</span>
+					<span v-if="formData.lastname.validity === Validity.INVALID" class="help is-danger">{{ $t( 'contact_form_lastname_error' ) }}</span>
 				</div>
 				<div class="has-margin-top-18">
 					<label for="donationNumber" class="subtitle">
@@ -70,9 +70,8 @@
 						class="is-form-input"
 						v-model="formData.topic.value"
 						id="category"
-						name="category"
-						:placeholder="$t( 'contact_form_topic_placeholder' )"
-						@input="$emit('field-changed', 'title')">
+						name="category">
+						<option hidden="hidden" disabled="disabled" value="">{{ $t( 'contact_form_topic_placeholder' ) }}</option>
 						<option v-for="option in contactData.contact_categories">{{ option }}</option>
 					</b-select>
 					<span v-if="formData.topic.validity === Validity.INVALID" class="help is-danger has-padding-top-18">{{ $t( 'contact_form_topic_error' ) }}</span>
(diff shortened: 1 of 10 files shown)


skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
+<template>
+	<div class="column is-full contact-form">
+		<h1 class="title">{{ $t( 'contact_form_title' ) }}</h1>
+		<span class="help is-danger has-padding-bottom-18" v-if="contactData.errors">{{ $t('contact_form_error') }}</span>
+		<form method="post" action="/contact/get-in-touch" @submit="submit" id="laika-contact">
+			<fieldset>
+				<div>
```
tzhelyazkova: unnecessary `div`

skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
...
+					</label>
+					<b-field>
+						<b-input type="text"
+								id="firstname"
+								name="firstname"
+								:placeholder="$t( 'contact_form_firstname_placeholder' )"
+								v-model="formData.name.value">
+						</b-input>
+					</b-field>
+					<span v-if="formData.name.validity === Validity.INVALID" class="help is-danger">{{ $t( 'contact_form_firstname_error' ) }}</span>
```
tzhelyazkova: The logic in all the `v-if`s can be transferred into `data` or a computed property. Kinda like what we did in `Address.vue`.

skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
...
+		return {
+			formData: {
+				name: {
+					name: 'name',
+					value: '',
+					pattern: '^.+$',
+					optionalField: true,
+					validity: Validity.VALID,
+				},
+				surname: {
```
tzhelyazkova: in the name of consistency let's have `firstName` and `lastName`

skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
...
+				let field = this.$data.formData[ fieldName ];
+				field.validity = Helper.inputIsValid( field.value, field.pattern, field.optionalField );
+				if ( field.validity !== Validity.VALID ) {
+					isValid = false;
+				}
+			} );
+			if ( isValid ) {
+				return true;
+			}
+			event.preventDefault();
```
tzhelyazkova: There are event modifiers in Vue which you can use directly in the template. 
In this case you can have `<form v-on:submit.prevent>` and remove this line from the ts.

skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
...
+		},
+	},
+	computed: {
```
tzhelyazkova: not a blocker but can you please move `computed` after `props`

skins/laika/src/store/util.ts
```diff
@@ -1,5 +1,6 @@
 import { Validity } from '@/view_models/Validity';
-import { PostData, FormData } from '@/view_models/Address';
+import { FormData, PostData } from '@/view_models/Address';
```
tzhelyazkova: ? :laughing:

skins/laika/src/store/util.ts
```diff
@@ -11,12 +12,21 @@ export const action = buildActionOrMutationName;
...
 export const Helper = {
-	inputIsValid: ( value: string, pattern: string ) => {
+	inputIsValid: ( value: string, pattern: string, isOptional?: boolean ) => {
+		if ( isOptional && Helper.isEmpty( value ) ) {
+			return Validity.VALID;
+		}
 		if ( pattern === null ) {
 			return value !== '' ? Validity.VALID : Validity.INVALID;
 		}
+		if ( Helper.isEmpty( value ) ) {
```
tzhelyazkova: uhm, isn't this just `isEmpty( value )`

skins/laika/src/store/util.ts
```diff
@@ -11,12 +12,21 @@ export const action = buildActionOrMutationName;
...
+			return Validity.INVALID;
+		}
 		return new RegExp( pattern ).test( value ) ? Validity.VALID : Validity.INVALID;
 	},
+	isEmpty( value: string ) {
+		return value === undefined || value === null || value === '';
```
tzhelyazkova: With TypeScript that's technically impossible. We're always passing a string, so all the check should do is compare the value to an empty string.

skins/laika/src/view_models/Contact.ts
```diff
@@ -0,0 +1,17 @@
+import { Validity } from '@/view_models/Validity';
+
+export interface AddressValidity {
+    [key: string]: boolean
+}
+
+export interface InputField {
+    name: string,
+    value: string | null,
```
tzhelyazkova: when is `value` null? According to the data object it's initially an empty string.

skins/laika/src/view_models/Contact.ts
```diff
@@ -0,0 +1,17 @@
+import { Validity } from '@/view_models/Validity';
+
+export interface AddressValidity {
```
tzhelyazkova: I didn't see this used anywhere.

skins/laika/src/components/pages/Contact.vue
```diff
@@ -0,0 +1,202 @@
...
+				</div>
+				<div class="has-margin-top-18">
+					<label for="category" class="subtitle">{{ $t( 'contact_form_topic_placeholder' ) }}</label>
+					<b-select
+						class="is-form-input"
+						v-model="formData.topic.value"
+						id="category"
+						name="category"
+						:placeholder="$t( 'contact_form_topic_placeholder' )"
+						@input="$emit('field-changed', 'title')">
```
tzhelyazkova: probably a copy paste mistake?

skins/laika/src/components/pages/Contact.vue
timEulitz: Cool :snowflake:

skins/laika/src/store/util.ts
timEulitz: I blame PHPStorm for autistically re-organizing that :man_shrugging:

skins/laika/src/view_models/Contact.ts
timEulitz: The drop down in its default state is `null` (empty string and undefined cause it to bug out) so I added `null` here. If we're putting on our fancy pants we can also create a new interface for just the subject dropdown but I figured this was okay and did not want to complicate things.

skins/laika/src/store/util.ts
timEulitz: I think it allows undefined and null to be passed anyway, that's why I had to add it. The subject drop down can be in a null state and I added undefined just to be sure.

skins/laika/src/store/util.ts
timEulitz: Typescript.

tzhelyazkova: then it must be `this.isEmpty`, referencing its own name is just weird :D

skins/laika/src/store/util.ts
tzhelyazkova: We have strict mode set to true in the TS compiler options, so strict null checks is also on -> you can't pass undefined and null to a string.

tzhelyazkova: If you actually want value here to be a string or null, we need a union type.

skins/laika/src/components/pages/Contact.vue
timEulitz: All the other input elements and their labels / errors have this `div` so it's there for consistency's sake I would argue.

tzhelyazkova: It literally  doesn't do anything :D

skins/laika/src/components/pages/Contact.vue
```diff
@@ -3,7 +3,7 @@
 		<h1 class="title">{{ $t( 'contact_form_title' ) }}</h1>
 		<span class="help is-danger has-padding-bottom-18" v-if="contactData.errors">{{ $t('contact_form_error') }}</span>
 		<span class="help is-danger has-padding-bottom-18" v-for="error in contactData.errors">{{ $t( error ) }}</span>
-		<form method="post" action="/contact/get-in-touch" @submit="submit" id="laika-contact">
+		<form method="post" action="/contact/get-in-touch" @submit="submit" v-on:submit.prevent id="laika-contact">
```
tzhelyazkova: `@submit="submit"` is the same as `v-on:submit`

skins/laika/src/components/pages/Contact.vue
tzhelyazkova: this is still here

skins/laika/src/components/pages/Contact.vue
```diff
@@ -176,8 +176,15 @@ export default Vue.extend( {
...
+		Validity: {
+			get() {
+				return Validity;
+			},
+		},
+	},
 	methods: {
-		submit( event: Event ) {
+		submit() {
 			let isValid = true;
```
tzhelyazkova: what does this flag do? :thinking:

skins/laika/src/view_models/Contact.ts
timEulitz: Copy-pasta :spaghetti:

skins/laika/src/components/pages/Contact.vue
timEulitz: Blagodarya!

skins/laika/src/components/pages/Contact.vue
timEulitz: It's needed too only submit the form conditionally. I have updated that part of the code now to not use the event.

PR #1599, opened 2019-08-08
Implement miscellaneous pages which are currently placeholders

timEulitz:


timEulitz:
> So is it decided we are not gonna provide the donation comment functionality?

I have created a ticket here but I am waiting for Jan's design specifications for it:

https://phabricator.wikimedia.org/T230338

However, the comment functionality will run in a pop-up so it will not have a separate page unlike in the previous skins.

gbirke:
Approved as an MVP, but for a better user "experience" (i.e. something goes wrong with a non-js browser), we should implement these three error page templates with additional `<noscript>` sections. I've created another ticket for that: https://phabricator.wikimedia.org/T230404

PR #1600, opened 2019-08-12
Add tracking to bank data component and refactor foreign IBANs

timEulitz:
https://phabricator.wikimedia.org/T230319

### DIFF
--- a/skins/laika/src/components/shared/PaymentBankData.vue
+++ b/skins/laika/src/components/shared/PaymentBankData.vue
@@ -59,13 +59,13 @@ export default Vue.extend( {
 		getTrackingCode(): string {
 			if ( this.looksLikeIban() ) {
 				return 'IBAN';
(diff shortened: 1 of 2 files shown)


tzhelyazkova:
> +2 for the IBAN/BIC related part.
> 
> I'm not sure if the tracking works reliably (might also a problem on the FAQ page), I've asked a question on the Matomo forum: https://forum.matomo.org/t/do-content-tracking-data-attributes-work-with-client-side-generated-html-vue-react/33844

It does work for FAQ, I've personally tested it.
Looking at the `Network` tab in dev tools, `piwik.js` is generated after chunk, vendors, and the entry point.

timEulitz:
I have tested it locally and it appears to be working just fine. You can test it on the "Tracking" Matomo and by selecting site 5 (test-spenden).

PR #1609, opened 2019-08-15
Implement donation comment list

timEulitz:
https://phabricator.wikimedia.org/T223062

Please review & finalize as I am on vacation now.


skins/laika/public/fonts/fontello/config.json
```diff
@@ -29,6 +29,18 @@
...
+      "uid": "d870630ff8f81e6de3958ecaeac532f2",
+      "css": "left-open",
+      "code": 59395,
+      "src": "fontawesome"
+    },
+    {
+      "uid": "399ef63b1e23ab1b761dfbb5591fa4da",
+      "css": "right-open",
+      "code": 59396,
+      "src": "fontawesome"
```
timEulitz: Make sure to run `make ui` otherwise the icons will not be visible.

skins/laika/src/components/pages/CommentList.vue
```diff
@@ -0,0 +1,106 @@
...
+			height: 48px;
+			&.mdi-arrow-left {
+				left: 0;
+			}
+			&.mdi-arrow-right {
+				left: 160px;
+			}
+		}
+		> .control {
+			position: absolute;
```
timEulitz: Unfortunately the Buefy dropdown really requires `absolute` position fuckery like this. It's not a big deal and should work well across browsers but I still don't like this sort of hackery. Alternatively we could have also used the native `Pagination` feature but it does not really match our design. I would not spend too much time on this now, it works and looks alright as far as I can tell.

src/Presentation/Presenters/CommentListJsonPresenter.php
```diff
@@ -8,19 +8,20 @@
...
 	public function present( CommentList $commentList ): array {
+		// TODO Translate keys into English once old skins are phased out
 		return array_map(
 			function( CommentWithAmount $comment ) {
 				return [
 					'betrag' => $comment->getDonationAmount(),
 					'spender' => $comment->getAuthorName(),
 					'kommentar' => $comment->getCommentText(),
 					'datum' => $comment->getDonationTime()->format( 'r' ),
+					'datum_formatted' => $comment->getDonationTime()->format( 'd.m.Y \u\m H:i:s \U\h\r' ),
```
timEulitz: Renders as `15.08.2019 um 16:38:30 Uhr` because formatting that in JavaScript would be SO MUCH more effort than doing it in PHP straight away. `'datum' => $comment->getDonationTime()->format( 'r' ),` is truly a pain to format in JavaScript, let's not waste time on that.

For backwards compatibility reasons I have left the existing `datum` field in place. I assume that you will not like the hard-coded German string (_"01.02.2003 um 12:34:56 Uhr"_) here but I personally don't think it's worth putting effort into creating a better solution for such a low-visibility page. If you still want to make it localizable, go ahead. Again though, I would really say let's not spend too much time on this tiny bit and maybe refactor this after the old skins are gone.

PR #1614, opened 2019-08-26
Activate skin test

gbirke:
Change campaign configuration to start skin test on August 27th 2019


gbirke:
Self-merging for upcoming test

PR #1622, opened 2019-08-28
Add anon disclaimer

gbirke:
This fixes https://phabricator.wikimedia.org/T231314


gbirke:
User-tested by Till & found ok, self-merging for upcoming test

PR #1623, opened 2019-08-28
Remove Placeholder component

tzhelyazkova:
This component is no longer needed because the skin is now active on production and this should not be used anywhere in the UI.


gbirke:
The "Address Data Update" still uses the placeholder, since we haven't implemented it yet. https://phabricator.wikimedia.org/T230334 They are not needed for the skin test, but might be needed during the campaign if we have a "donate as anon, add address on the donation confirmation page" kind of test. If we remove the Placeholder, the address update page will show an error instead of a placeholder.

tzhelyazkova:
> The "Address Data Update" still uses the placeholder, since we haven't implemented it yet. https://phabricator.wikimedia.org/T230334 They are not needed for the skin test, but might be needed during the campaign if we have a "donate as anon, add address on the donation confirmation page" kind of test. If we remove the Placeholder, the address update page will show an error instead of a placeholder.

Ok, that begs the question why do we even serve that code. 
I'll close this for now.

PR #1625, opened 2019-08-29
Fix mobile laika

gbirke:
This is for https://phabricator.wikimedia.org/T230904 and https://phabricator.wikimedia.org/T231647

### DIFF
--- a/skins/laika/src/components/pages/Supporters.vue
+++ b/skins/laika/src/components/pages/Supporters.vue
@@ -1,5 +1,5 @@
 <template>
-	<div class="column is-full">
+	<div class="supporters">
 		<supporter v-for="( supporter, index ) in supporters"
 				:key="index"
 				v-on:supporter-opened="setSupporterId( $event )"
--- a/skins/laika/src/components/pages/Supporters.vue
+++ b/skins/laika/src/components/pages/Supporters.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="supporters">
-		<supporter v-for="( supporter, index ) in supporters"
+		<supporter lang="de" v-for="( supporter, index ) in supporters"
 				:key="index"
 				v-on:supporter-opened="setSupporterId( $event )"
 				v-on:supporter-closed="setSupporterId( null )"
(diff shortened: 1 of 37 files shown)


timEulitz:
Two things I have noticed while checking it out but I am not sure if this just falls under the WIP state of this task:

I think according to the mobile design, there should be a minimum 3 pixels gap on the left and right of the screen, so the white main form does not extend all the way to the sides of the screen on mobile. This means the form is 314px wide in its smallest state rather than 320px.

Also, the changes of this PR affect the desktop design by removing all the padding from there too:

design

skins/laika/src/scss/variables.scss
```diff
@@ -39,6 +39,10 @@ $custom-colors: (
 		"gray-very-light": ($fun-color-gray-very-light),
 );
 
+// 320px is the smallest assumed mobile size, minus 2 x column gap (18px).
+// We don't use calc(100vw-36px) for compat reasons
+$fun-smallest-width: 284px;
```
timEulitz: I think according to the mobile design, there should be a minimum 3 pixels gap on the left and right of the screen, so the white main form does not extend all the way to the sides of the screen on mobile. This means the form is 314px wide in its smallest state rather than 320px, so this value should be 278px.

gbirke: Good catch! I'm thanking my past self for making this a variable.

timEulitz:
The supporters list is cut off partially on 320px width devices.

timEulitz:
> The predefined amounts are now rectangular opposed to before when they were square. Not sure if this is a problem for the design.

It was one of those "silent changes" introduced to the desktop design, it also looks like that on Figma now.

skins/laika/src/components/pages/Supporters.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="supporters">
-		<supporter v-for="( supporter, index ) in supporters"
+		<supporter lang="de" v-for="( supporter, index ) in supporters"
```
tzhelyazkova: that kind of goes in the opposite direction of the whole i18n thing we got set up.

timEulitz: It was mainly for the `hyphens:auto` CSS attribute to ensure that it will hyphenate the German company names and German donor names according to German hyphenation rules. Just a quirky on feature to make it look better in the browsers that support it.

PR #1627, opened 2019-09-02
Fix design of radio button elements

timEulitz:
https://phabricator.wikimedia.org/T230866


skins/laika/src/scss/variables.scss
```diff
@@ -3,7 +3,6 @@ $widescreen-enabled: false;
 $fullhd-enabled: false;
 
 $fun-color-bright: #ffffff;
-$fun-color-off-bright: #b7b7b7;
```
timEulitz: This color was a duplicate so I deleted it and replaced all of its occurrences.

PR #1629, opened 2019-09-06
Display server response message on success

tzhelyazkova:
Feature: [T230338](https://phabricator.wikimedia.org/T230338#5468077)
There are different messages the server responds with when the status is `OK`. Therefor we want to display the message from the server in the popup instead of the comment text field.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8121,12 +8121,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/wmde/fundraising-frontend-content",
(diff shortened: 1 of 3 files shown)


tzhelyazkova:
> Looks good to me, I would just use a different message for the button, the "Cancel" text does not really fit. Maybe create a new string like "Zur Spendenübersicht" ("Back to the donation summary") instead.

Fine :nerd_face: :de:

PR #1630, opened 2019-09-06
Split laika form tracking

tzhelyazkova:
Feature: [T231555](https://phabricator.wikimedia.org/T231555)

Split the form in `laika` skin into two forms - payment and personal data.
This will enable us to set up more thorough tracking in Matomo.

### DIFF
--- a/skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
+++ b/skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
@@ -3,7 +3,7 @@
 			id="laika-donation-personal-data"
 			class="address-page column is-full"
 			ref="personal"
-			action="#"
+			action="/donation/add"
 			method="post">
 		<h1 v-if="!paymentWasInitialized" class="title is-size-1">{{ $t( 'donation_form_section_headline' ) }}</h1>
 		<div class="has-margin-left-18 has-margin-right-18">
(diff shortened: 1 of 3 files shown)


skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
```diff
@@ -134,13 +139,18 @@ export default Vue.extend( {
 							document.getElementsByClassName( 'help is-danger' )[ 0 ].scrollIntoView( { behavior: 'smooth', block: 'center', inline: 'nearest' } );
 							return;
 						}
-						this.$emit( 'submit-donation' );
+						( this as any ).submitDonationForm();
```
gbirke: 😭 😞 
Nothing you can do here with our current setup, I'm really looking forward to Vue 3

skins/laika/src/components/pages/donation_form/subpages/PaymentPage.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-	<div class="payment-page">
+	<form name="laika-donation-payment-data"
+		id="laika-donation-payment-data"
+		class="payment-page column is-full"
```
gbirke: I'd lose the "data" suffix here, because it's kind of implied. "Personal data" is a compund word, but "payment" would suffice imho. But that's just my personal taste and me trying to avoid any hints of AbstractDataFactoryModelManagerContainerHelper.

skins/laika/src/components/pages/donation_form/subpages/AddressPage.vue
tzhelyazkova: Yeah, I was surprised the function is not recognized by TS.
I know we have other cases like this and there's no TS complaint there, idk :woman_shrugging:

PR #1632, opened 2019-09-06
Implement various fixes to responsive Laika design

timEulitz:
I have gone through a bunch of pages and tried to make them look like in Figma on desktop an mobile. The biggest "broken" thing right now is the donation comment popup on mobile - scrolling to the bottom is not possible on smaller viewport heights which makes the comment form submission impossible.

### DIFF
--- a/skins/laika/package.json
+++ b/skins/laika/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "axios": "^0.18.1",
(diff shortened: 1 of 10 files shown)


timEulitz:
I have fixed the comment pop-up now, there was a typo in the Vue property and the `overflow-y` had to be set to auto.

PR #1638, opened 2019-09-17
Composer maintenance update

timEulitz:


timEulitz:
@tzhelyazkova I ran update-js and Laika was not updated, I was also surprised and I cannot imagine there are no updated libraries but that should have covered Laika as well?!

tzhelyazkova:
> @tzhelyazkova I ran update-js and Laika was not updated, I was also surprised and I cannot imagine there are no updated libraries but that should have covered Laika as well?!

Nah, it covers cat17 and 10h16. `install-js` covers all.

timEulitz:
Ah alright, thanks, I will re-run that tomorrow.

PR #1643, opened 2019-10-02
Add Bucket testing for Laika

gbirke:
Introduce bucket testing features for Laika, add first test

### DIFF
--- a/skins/laika/src/scss/overrides.scss
+++ b/skins/laika/src/scss/overrides.scss
@@ -68,6 +68,28 @@
 	}
 }
 
+.campaigns {
+	&--anon-form-display {
+		&--two-steps {
+			.radio-container {
+				display: flex;
+			}
+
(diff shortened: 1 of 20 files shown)


skins/laika/src/scss/overrides.scss
```diff
@@ -68,6 +68,25 @@
...
+	}
+}
+
+.campaigns--anon-form-display--address-type {
+	/* Remove margin added by buefy, since each button is on its own line */
+	.b-radio.radio + .radio {
+		margin-left: 0;
+	}
+}
+
```
timEulitz: I would suggest to group this with nesting so it's easier to see what belongs to this test:

```
.campaigns {
    &--anon-form-display {
         &--two-steps {
           // ...
         }
         &--address-type {
           // ...
         }
    }
}
```

PR #1648, opened 2019-10-08
Change style of payment summary

tzhelyazkova:
Feature: [T234425](https://phabricator.wikimedia.org/T234425)


tzhelyazkova:
> I really like the flexbox solution!
> 
> At the moment, the flex properties make the child elements evenly spaced, which breaks the left text into 2 lines for some payment types. I think the spirit of the ticket is to make the payment summary to be as slim as possible, at least for desktop and tablet viewports. Maybe you can change the flex properties of the children to give the text more room than the button.
> 
> Bildschirmfoto 2019-10-08 um 18 26 42

I asked in the ticket about the end goal. I also thought is to make everything in one row, but at the end of the day that depends on the very long German string on the left right :)
I will apply what you suggested and see if it works for our current strings.

PR #1652, opened 2019-10-22
Remove cat17 from FundraisingFrontend

timEulitz:
What a day


composer.json
```diff
@@ -4,7 +4,7 @@
 	"homepage": "https://github.com/wmde/FundraisingFrontend",
 	"license": "GPL-2.0+",
 	"require": {
-		"php": ">=7.1",
+		"php": ">=7.3",
```
timEulitz: I am making use of the `object` return type in one of the tests which required PHP 7.2. The servers are running 7.3 sot the Composer file should reflect that. I actually could not find a reference to this for the Docker settings but Docker seems to be working just fine.

tzhelyazkova:
YAY !

composer.json
gbirke: The PHP version for our custom Docker images is in https://github.com/wmde/fundraising-frontend-docker/blob/master/generate_dockerfiles.php

PR #1656, opened 2019-10-30
Make skin changeable until end of January 2020

gbirke:


gbirke:
Self-Merging to enable testing quickly

PR #1657, opened 2019-11-03
Compact design

timEulitz:
https://phabricator.wikimedia.org/T236667


gbirke:
The design is deployed to production to be able to start the test. Please check if there are any issues with the code.

One last issue is missing space on small screen sizes, I could fix that in a followup ticket.
Image Pasted at 2019-11-5 13-01

gbirke:
Self-merging now since the code is on prod anyway, but has slight variances from master in the campaign config. Please still look at it and review it, @moiikana @timEulitz

PR #1662, opened 2019-11-14
Implement address update modal for anon donations

timEulitz:
https://phabricator.wikimedia.org/T237272

### DIFF
--- a/skins/laika/src/components/pages/donation_confirmation/AddressModal.vue
+++ b/skins/laika/src/components/pages/donation_confirmation/AddressModal.vue
@@ -1,0 +1,203 @@
+<template>
+	<form id="address-update-form" name="address-update-form" v-on:submit.prevent="submit" method="post" ref="form" class="modal-card">
+		<address-type v-on:address-type="setAddressType( $event )" :disabled-anonymous-type="true"></address-type>
+		<name :show-error="fieldErrors" :form-data="formData" :address-type="addressType" v-on:field-changed="onFieldChange"></name>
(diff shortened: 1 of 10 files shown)


skins/laika/src/components/pages/donation_confirmation/AddressModal.vue
```diff
@@ -0,0 +1,203 @@
...
+					axios.post(
+						this.$props.updateDonorUrl,
+						jsonForm,
+						{ headers: { 'Content-Type': 'multipart/form-data' } }
+					).then( ( validationResult: AxiosResponse<any> ) => {
+						if( validationResult.data.status === 'OK' ) {
+							this.$emit( 'address-updated' );
+						} else {
+							this.$emit( 'address-update-failed' );
+						}
```
gbirke: With the current event handler setup in `DonationConfirmation`, the modal will close when the server returns an error. This seems like a sensible behavior to me (because "retrying" it in the model probably won't help), but maybe we should pass the error message key in the event and display it in `DonationConfirmation`.

timEulitz: I am not sure if you miswrote or if the behavior broke in the meantime but the modal will / should not close when an error is received. I think Jan mentioned this during a UX meeting and his reasoning was sort of like "the user should have time to read the text an move on on their own terms" so the modal stays up.

gbirke: I did not user-test that, I deducted the behavior from the code. Good to know, I'll user-test it then when the PR is finished.

timEulitz:
@gbirke Still WIP, I will pick it up in the morning.

PR #1665, opened 2019-11-15
Improve donation confirmation modal

gbirke:


gbirke:
Self-merging for upcoming test, of you see something you don't like, @timEulitz , please do a followup PR

PR #1669, opened 2019-11-19
Fix close button on mobile address change modal

timEulitz:


skins/laika/src/scss/overrides.scss
```diff
@@ -140,15 +140,28 @@ button.is-low {
...
+			padding: 0;
+		}
+	}
+}
+.address-modal {
+	.modal {
+		&-close {
+			&:after, &:before {
+				@include until($tablet) {
+					background-color: $fun-color-dark;
```
gbirke: 🤣

timEulitz: Yeah, I hate it too :see_no_evil:.

PR #1678, opened 2019-11-26
Prevent named comments on anonymous donations

timEulitz:
https://phabricator.wikimedia.org/T238009


skins/laika/src/components/DonationCommentPopUp.vue
```diff
@@ -17,7 +17,7 @@
 				<b-input id="comment" name="comment" type="textarea"></b-input>
 				<p v-if="commentErrored" class="help is-danger"> {{ $t( 'donation_comment_popup_error' ) }}</p>
 			</div>
-			<div class="field has-margin-bottom-18">
+			<div class="field has-margin-bottom-18" v-if="!isAnonymous">
```
gbirke: I found it a bit confusing to have a field with the same name as a computed property but its own value.  Maybe call the property `showNonAnonymous` or `showPublishAuthor`?

PR #1679, opened 2019-11-27
Add amount too low clientside validation message

moiikana:
split up the amount's invalid state into two distinctive states (-> too high, -> too low)
so the Payment-component can check which error message it should display

https://phabricator.wikimedia.org/T239202

### DIFF
--- a/skins/laika/src/store/payment/getters.ts
+++ b/skins/laika/src/store/payment/getters.ts
@@ -1,5 +1,5 @@
 import { GetterTree } from 'vuex';
-import { Payment } from '@/view_models/Payment';
+import { AmountValidity, Payment } from '@/view_models/Payment';
 import { Validity } from '@/view_models/Validity';
 
 export const getters: GetterTree<Payment, any> = {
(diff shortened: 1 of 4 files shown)


skins/laika/src/store/payment/getters.ts
```diff
@@ -9,6 +9,19 @@ export const getters: GetterTree<Payment, any> = {
 	typeIsValid: function ( state: Payment ): boolean {
 		return state.validity.type !== Validity.INVALID;
 	},
+	amountValidity: function ( state: Payment ): AmountValidity {
+		/* TODO reuse configuration amounts from config files:
+		    "donation-minimum-amount": 1,
+		    "donation-maximum-amount": 100000,
+		    */
```
gbirke: I've created the ticket you can reference here: https://phabricator.wikimedia.org/T239349

skins/laika/src/store/payment/getters.ts
```diff
@@ -9,6 +9,19 @@ export const getters: GetterTree<Payment, any> = {
 	typeIsValid: function ( state: Payment ): boolean {
 		return state.validity.type !== Validity.INVALID;
 	},
+	amountValidity: function ( state: Payment ): AmountValidity {
```
gbirke: Please write a unit test for the new getter.

PR #1680, opened 2019-11-27
Show warnings when form placeholders are used

timEulitz:
https://phabricator.wikimedia.org/T238535

### DIFF
--- a/src/Infrastructure/InternetDomainNameValidator.php
+++ b/src/Infrastructure/InternetDomainNameValidator.php
@@ -7,12 +7,23 @@ namespace WMDE\Fundraising\Frontend\Infrastructure;
 use WMDE\FunValidators\DomainNameValidator;
 
 /**
- * @licence GNU GPL v2+
- * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @license GNU GPL v2+
  */
 class InternetDomainNameValidator implements DomainNameValidator {
 
+	private const BANNED_DOMAINS = [
+		'example.com'
+	];
+
 	public function isValid( string $domain ): bool {
+		return !$this->isDomainBanned( $domain ) && $this->domainHasDnsEntry( $domain );
+	}
+
+	private function isDomainBanned( string $domain ): bool {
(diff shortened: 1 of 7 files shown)


src/Infrastructure/InternetDomainNameValidator.php
```diff
@@ -12,8 +12,17 @@
...
+	];
+
 	public function isValid( string $domain ): bool {
-		return checkdnsrr( $domain, 'MX' ) || checkdnsrr( $domain, 'A' ) || checkdnsrr( $domain, 'SOA' );
+		return !$this->isDomainBanned( $domain ) &&
+			( checkdnsrr( $domain, 'MX' ) || checkdnsrr( $domain, 'A' ) || checkdnsrr( $domain, 'SOA' ) );
+	}
+
+	private function isDomainBanned( string $domain ): bool {
+		return in_array( trim( $domain ), self::BANNED_DOMAINS );
```
gbirke: I don't understand why the `trim` is needed here, at this point of the application the domain name (coming from the email) should already be trimmed. But it doesn't hurt.

Maybe also add `strtolower` so people don't get away with `exAmple.cOM`

src/Infrastructure/InternetDomainNameValidator.php
```diff
@@ -12,8 +12,17 @@
...
 class InternetDomainNameValidator implements DomainNameValidator {
 
+	private const BANNED_DOMAINS = [
+		'example.com'
+	];
+
 	public function isValid( string $domain ): bool {
-		return checkdnsrr( $domain, 'MX' ) || checkdnsrr( $domain, 'A' ) || checkdnsrr( $domain, 'SOA' );
+		return !$this->isDomainBanned( $domain ) &&
+			( checkdnsrr( $domain, 'MX' ) || checkdnsrr( $domain, 'A' ) || checkdnsrr( $domain, 'SOA' ) );
```
gbirke: For consistencies sake you could also put this check into a `domainHasDnsEntry` method, so the `isValid` reads more nicely.

timEulitz: Yeah, that makes the code much nicer for sure, I have made that change!

PR #1681, opened 2019-11-28
Handle ENTER key event on payment page

moiikana:
https://phabricator.wikimedia.org/T239395

### DIFF
--- a/skins/laika/src/components/pages/donation_form/subpages/PaymentPage.vue
+++ b/skins/laika/src/components/pages/donation_form/subpages/PaymentPage.vue
@@ -4,7 +4,8 @@
 		class="payment-page"
 		ref="payment"
 		action="/donation/add"
-		method="post">
+		method="post"
+		@keydown.enter.prevent="next()">
 		<payment v-bind="$props"></payment>
 		<div class="level has-margin-top-36">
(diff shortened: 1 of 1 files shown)


skins/laika/src/components/pages/donation_form/subpages/PaymentPage.vue
```diff
@@ -4,7 +4,8 @@
 		class="payment-page"
 		ref="payment"
 		action="/donation/add"
-		method="post">
+		method="post"
+		@keydown.enter.prevent="next()">
```
gbirke: This will call the `next()` method too early, leading to error messages and not being able to "escape" fields via "enter" (enter acting more like "Tab"). I'd prefer a behavior where hitting enter blurs the fields (and thus triggers the blur behavior). Please consult with Jan for the behavior.

gbirke: After checking some more, I saw that I ran into Firefox-on-Mac bugs/features. See https://bugzilla.mozilla.org/show_bug.cgi?id=756028 and https://bugzilla.mozilla.org/show_bug.cgi?id=1535742

PR #1684, opened 2019-12-02
Improve donation confirmation modal form tracking

timEulitz:
https://phabricator.wikimedia.org/T239115


skins/laika/src/components/DonationCommentPopUp.vue
```diff
@@ -69,6 +69,9 @@ export default Vue.extend( {
 			},
 		},
 	},
+	created: function () {
```
gbirke: IMHO this  should be `mounted` instead of `created`, because the DOM nodes don't exist at this point. See https://alligator.io/vuejs/component-lifecycle/

timEulitz:
@gbirke I am a bit stuck on resolving the Travis issue, I have tried to globally declare `_paq` but all of that is not working, maybe I am completely misunderstanding the issue here. Could you have a look? Maybe this needs to be approached completely differently after all?

gbirke:
I think this has nothing to do with this commit, but was probably introduced in #1678 
The `DonationCommentPopUp.spec.js` creates a new `DonationCommentPopUp` which uses tracking functions that expect `_paq` to be available. My guess is the test environment has a different set of "global" variables for each test. My approach would be to mock the tracking functions themselves instead of `_paq`, that's why we've encapsulated the functionality. See
 https://jestjs.io/docs/en/es6-class-mocks for how to mock ES6 module imports with Jest.

PR #1690, opened 2019-12-18
Disable payment type or interval on choosing sub or 1 

moiikana:
https://phabricator.wikimedia.org/T240983

### DIFF
--- a/skins/laika/src/components/pages/donation_form/Payment.vue
+++ b/skins/laika/src/components/pages/donation_form/Payment.vue
@@ -63,10 +63,12 @@ export default Vue.extend( {
 				}
 				return disabledTypes;
 			},
-			disabledPaymentIntervals: ( state: any ) => {
+			disabledPaymentIntervals: function ( state: any ) {
 				var disabledIntervals : String[] = [];
 				if ( state[ NS_PAYMENT ].values.type === 'SUB' ) {
-					disabledIntervals.push( '1', '3', '6', '12' );
+					disabledIntervals = this.$props.paymentIntervals
+						.filter( ( interval:Number ) => Number( interval ) > 0 )
+						.map( ( interval:Number ) => String( interval ) );
 				}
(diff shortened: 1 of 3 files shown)


skins/laika/src/components/pages/donation_form/Payment.vue
```diff
@@ -52,7 +53,23 @@ export default Vue.extend( {
...
+				}
+				if ( state[ NS_PAYMENT ].values.interval !== '0' ) {
+					disabledTypes.push( 'SUB' );
+				}
+				return disabledTypes;
+			},
+			disabledPaymentIntervals: ( state: any ) => {
+				var disabledIntervals : String[] = [];
+				if ( state[ NS_PAYMENT ].values.type === 'SUB' ) {
+					disabledIntervals.push( '1', '3', '6', '12' );
```
gbirke: It's highly unlikely that we'll have different intervals in the future, but in the interest of keeping the amount of places where we list the intervals low, you could do 

    disabledIntervals = this.$props.paymentIntervals.filter( interval => Number(interval) > 0 ).map( interval => String(interval));

or wrap that in a function called `getListOfRecurringIntervals`

skins/laika/src/components/shared/PaymentInterval.vue
```diff
@@ -7,9 +7,13 @@
 						:id="'interval-' + interval"
 						name="interval"
 						v-model="selectedInterval"
-						:native-value="interval.toString()"
+						:native-value="interval.toString()" :disabled="disabledPaymentIntervals.indexOf( interval.toString() ) > -1"
```
gbirke: Linebreak missing here

skins/laika/src/components/pages/donation_form/Payment.vue
moiikana: obviously the better approach, thanks

PR #1691, opened 2019-12-30
Update links for membership info

gbirke:
See https://phabricator.wikimedia.org/T241080#5766808


gbirke:
Self-merging because it's just a markup change

PR #1705, opened 2020-03-13
Delete 10h16

gbirke:
https://phabricator.wikimedia.org/T242435


Abban:
@gbirke Can you look at line 39 in the Makefile? It contains a reference to 10h16 but I'm not sure what the command does.

Abban:
CI tests are passing locally..

gbirke:
The `install-js` and `update-js` lines need to be adapted: Instead of relying on the postinstall and postupdate hooks and the `npm_skins` script to distribute the load, run `npm install` and `npm update` directly in the laika directory (specifying a workdir with the `-w` parameter), while still mounting the main application directiory in `/data` inside the docker container. Example for update:

    docker run --rm $(DOCKER_FLAGS) --user $(current_user):$(current_group) -v $(BUILD_DIR):/data:delegated -w /data/skins/laika -v $(TMPDIR)/home:/var/nodehome:delegated -v $(TMPDIR)/passwd:/etc/passwd $(NODE_IMAGE) npm update $(NPM_FLAGS)

package.json
```diff
@@ -3,12 +3,10 @@
 	"version": "2.0.0-dev",
 	"description": "WMDE fundraising application (end user facing part)",
 	"scripts": {
-		"postinstall": "node npm_skins install",
-		"build-assets": "node npm_skins run build-assets",
-		"copy-assets": "find skins/ -maxdepth 2 -name 'web' -type d -exec sh -c 'dirname \"$1\" | xargs -I{} cp -r \"./{}/web/.\" \"./web/{}\"' -- {} \\;",
-		"test": "node npm_skins run test",
-		"cs": "node npm_skins run cs",
-		"ci": "node npm_skins run ci",
+		"postinstall": "cd skins/laika && npm install",
```
gbirke: We need a similar thing for `postupdate`

gbirke:
> The `install-js` and `update-js` lines need to be adapted: Instead of relying on the postinstall and postupdate hooks and the `npm_skins` script to distribute the load, run `npm install` and `npm update` directly in the laika directory (specifying a workdir with the `-w` parameter)

Now that I saw the `package.json`I changed my mind - it's probably better to just run the central package.json and rely on postinstall/postupdate hooks. You can remove the `-w` parameter.

PR #1708, opened 2020-03-19
Invert address change relation

gbirke:


src/Infrastructure/DoctrinePostPersistSubscriberCreateAddressChange.php
```diff
@@ -0,0 +1,127 @@
...
+					$address->getPostalCode(),
+					$address->getCity(),
+					$address->getCountryCode()
+				);
+			default:
+				return null;
+		}
+	}
+
+	private function getAddressType( object $name ): int {
```
gbirke: In cases like these where you have a [Union Typehint](https://stackoverflow.com/questions/46775489/is-it-possible-to-type-hint-more-than-one-type) with types that have the same "interface" (meaning they share the same method, not that they share the same interface) it's better to have no PHP typehint and specify the union type in a doc comment. PHP 8 will fix this.

Also, I'd rather have `AddressChangedBuilder` passed in here and called directly instead of "indirecting" with constants. You'll decrease the overall cyclomatic complexity of this class.

tests/Fixtures/FakeEntityManager.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\Fixtures;
+
+use Doctrine\ORM\EntityManager;
+
+class FakeEntityManager extends EntityManager {
```
gbirke: In our [nomenclature of test doubles](https://www.entropywins.wtf/blog/2016/05/13/5-ways-to-write-better-mocks/) this would be an `EntityManagerSpy`, because it records calls.

.gitignore
```diff
@@ -1,8 +1,11 @@
 *~
 !.*
 
+.DS_Store
```
gbirke: IMHO this Mac-specific file type should rather go into your [`~/.gitignore_global`](http://egorsmirnov.me/2015/05/04/global-gitignore-file.html) file

Abban: I agree, but I also like having redundancy. Those things have a tendency to sneak into a repo when you're not looking. I'll remove it from here anyway.

gbirke:
I just realized that `DoctrinePostPersistSubscriberCreateAddressChange` checks for the wrong instances of Donation and Membership :( 

Until the [epic database refactoring](https://phabricator.wikimedia.org/T203679) is done, we have our domain objects in the namespace `WMDE\Fundraising\SomeContext\Domain\Model` that handle state changes and business logic. These domain objects get translated into Doctrine entities (currently in the `WMDE\Fundraising\Entities` namespace, soon in the `WMDE\Fundraising\SomeContext\DataAccess\Entities` namespace) in the Repository classes. The Doctrine event handling only knows Doctrine entities, so you'll have to change the code to use the entities.

Abban:
@gbirke Can you take another look through the implementation of the event handler?

In swapping the entities over I discovered that setting the `adresstyp` data array item is not consistent between the Donation and MembershipApplication entities.

Also when Donations are created in the ListCommentsHtmlRouteTest and ListCommentsJsonRouteTest edge to edge tests, the `adresstyp` is not being created. Rather than adjust the tests I made the event handler default to a private address if it's not set in the Donation.

Abban:
**Note:** I had to update the php version to 7.4. This will require the server to be updated to it on the next deployment.

PR #1711, opened 2020-03-24
Add Vuex feature toggle plugin

gbirke:
Add new hypothetical action for initializing address data. At the
moment we're using it only as a trigger to change the default data for
one of the Vuex plugins that reacts to a feature toggle.


gbirke:
This is how I imagine the Vuex plugin. Let's take this PR as a basis for discussion, not something to be merged immediately.

PR #1718, opened 2020-04-01
Add path to server migration readme

moiikana:


deployment/README.md
```diff
@@ -62,7 +62,9 @@ You can add the parameters `-l fundraising_frontend_test` or `-l fundraising_fro
 The contents of the maintenance mode files are stored at https://github.com/wmde/fundraising-maintenance
 
 ### 3. Run migration scripts
-Login to the server of the fundraising application (test or production), change to the new release directory (the timestamped directory, *not* the `html`).
+Login to the server of the fundraising application (test server or production server, depending on your use case).
+In /usr/share/ngix/www/ you can find the folder of the server you want to migrate. The folder contains the release directories.
```
gbirke, line 66:
```suggestion
In `/usr/share/ngix/www/` you can find the directory namend after the domain of the server you want to migrate. The directory contains the the last 5 releases, each in its own directory.
```

Someone told me once that the word "folder" is only used by Mac users and that the word to use is "directory". Since then I've always used "directory" in the documentation (even when calling them folders in everyday language). I don't mind gradually converting to "folder" (if we call agree that we call those things "folders" all the time), but we should keep it consistent on a per-document basis.

deployment/README.md
```diff
@@ -62,7 +62,9 @@ You can add the parameters `-l fundraising_frontend_test` or `-l fundraising_fro
 The contents of the maintenance mode files are stored at https://github.com/wmde/fundraising-maintenance
 
 ### 3. Run migration scripts
-Login to the server of the fundraising application (test or production), change to the new release directory (the timestamped directory, *not* the `html`).
+Login to the server of the fundraising application (test server or production server, depending on your use case).
```
gbirke, line 65:
```suggestion
Login to the server of the fundraising application (test server or production server, depending on which version you want to deploy).
```

PR #1721, opened 2020-04-03
Add explanation message when coming from banner with BEZ

moiikana:
https://phabricator.wikimedia.org/T248819

### DIFF
--- a/skins/laika/src/components/pages/donation_form/Address.vue
+++ b/skins/laika/src/components/pages/donation_form/Address.vue
@@ -9,6 +9,7 @@
 					slot="campaigns.address_type.no_preselection"
 					v-on:address-type="setAddressType( $event )"
 					:disabledAddressTypes="disabledAddressTypes"
+					:is-direct-debit-from-banner="isDirectDebitFromBanner"
 					:initial-address-type="null">
 			</address-type>
 
@@ -16,6 +17,7 @@
 					slot="campaigns.address_type.preselection"
 					v-on:address-type="setAddressType( $event )"
 					:disabledAddressTypes="disabledAddressTypes"
+					:is-direct-debit-from-banner="isDirectDebitFromBanner"
 					initial-address-type="person">
 			</address-type>
 			<div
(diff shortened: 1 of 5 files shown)


skins/laika/src/components/pages/donation_form/Address.vue
```diff
@@ -50,7 +52,7 @@ import { mapGetters } from 'vuex';
 import { AddressValidity, AddressFormData, ValidationResult } from '@/view_models/Address';
 import { AddressTypeModel } from '@/view_models/AddressTypeModel';
 import { Validity } from '@/view_models/Validity';
-import { NS_ADDRESS } from '@/store/namespaces';
+import { NS_ADDRESS, NS_PAYMENT } from '@/store/namespaces';
```
gbirke, line 55: This feels kind of wrong for a component named `Address`

moiikana: @gbirke I moved it up to AddressPage.
I'm just not sure how to test the new property.. 
I added a test in AddressType.spec.ts to check whether setting the property changes the html and shows the message, but this is probably not what exactly should be tested. 
I actually wanted to test in AddressType.spec.ts whether the state being "initialized" and the getter isDirectDebit being a certain way, the property value should be true/false, but couldn't implement how to get that "initialized" value. Should I keep trying to implement that?

gbirke: The test in `AddressType.spec.ts` is about the visibility of the element, depending on the parameter. So you'd instantiate the wrapper with different values for `isDirectDebitFromBanner` in `propsData`, find the element and call its `isVisible` method in an assertion.

There are no tests for `AdressPage` yet, so that's more tricky. One test I can think of at the moment would be to take 2 snapshots of the rendered component (doing a mount instead of shallowmount), with different (mock)store states and asserting that the snapshots are different. 
Another (maybe easier) way to do this would be to [stub](https://vue-test-utils.vuejs.org/api/options.html#stubs) the Address component in the `AddressPage` test with a custom component that renders the value of the propery `isDirectDebitFromBanner` and then use the `find` method of the wrapper to check the output of the stub component.

gbirke, line 55: Please remove the unused import of `NS_PAYMENT`

PR #1723, opened 2020-04-07
Actiavte address preselection A/B test

gbirke:


gbirke:
I've manually adapted the deployed campaign config on the producation server, this PR is just to keep the campaign active for the next deploys.

PR #1724, opened 2020-04-07
Fix use-of-funds page

gbirke:
Move the code for creation of the use-of-funds renderer back into
FunFunFactory because the skin a/b test no longer exists and the
ChoiceFactory was throwing errors.

This is a followup for #1705


gbirke:
Deployed & fixed

PR #1725, opened 2020-04-07
Show SUB info only if SUB is chosen

moiikana:
... and show the proper info message for BEZ payment type.

https://phabricator.wikimedia.org/T248820

### DIFF
--- a/skins/laika/src/components/pages/donation_form/PaymentType.vue
+++ b/skins/laika/src/components/pages/donation_form/PaymentType.vue
@@ -11,8 +11,11 @@
 						:disabled="disabledPaymentTypes.indexOf( paymentType ) > -1"
 						@change.native="setType">
 					{{ $t( paymentType ) }}
-					<div v-show=" disabledPaymentTypes.indexOf( paymentType ) > -1" class="has-text-dark-lighter has-margin-top-18">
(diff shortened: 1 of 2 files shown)


skins/laika/src/components/pages/donation_form/PaymentType.vue
```diff
@@ -11,9 +11,12 @@
 						:disabled="disabledPaymentTypes.indexOf( paymentType ) > -1"
 						@change.native="setType">
 					{{ $t( paymentType ) }}
-					<div v-show=" disabledPaymentTypes.indexOf( paymentType ) > -1" class="has-text-dark-lighter has-margin-top-18">
+					<div v-show="disabledPaymentTypes.indexOf( paymentType ) > -1 && paymentType==='SUB'" class="has-text-dark-lighter has-margin-top-18">
```
Abban, line 14: It might be better to move this logic to a component method to make these elements cleaner.

Your markup would then be something like: `v-show="isActivePaymentType('SUB')"`

moiikana: I need to check for the specific paymenttype value right there in the template because it should show 2 different messages in each case, so I could not move this to a method. But I did move it at least for the array check.

PR #1730, opened 2020-04-28
Update laika

Abban:
https://phabricator.wikimedia.org/T249560


skins/laika/templates/Base_Layout.html.twig
```diff
@@ -39,7 +39,10 @@
 {% endblock %}
 
 {% block scripts %}
-    {% if not assets_path %}
+    {% if assets_path %}
+        <script src="{% if assets_path %}{$ assets_path $}{% else %}{$ basepath $}/skins/laika{% endif %}/js/chunk-vendors.js"></script>
```
gbirke, line 43: why check again for assets_path?

Also, in the else branch, we need to add chunk-common

gbirke:
I jsut checked the compiled version myself and it looks good. Consider this approved after you've cleaned up the duplicate "if"s.

Interestingly, the build does not generate `chunk-commons.js` that ssems to be a dev thing.

skins/laika/templates/Base_Layout.html.twig
```diff
@@ -39,7 +39,10 @@
 {% endblock %}
 
 {% block scripts %}
-    {% if not assets_path %}
+    {% if assets_path %}
+        <script src="{$ assets_path $}/js/chunk-vendors.js"></script>
+        <script src="{$ assets_path $}/skins/laika/js/chunk-common.js"></script>
```
gbirke, line 44: Path must be 
```suggestion
        <script src="{$ assets_path $}/js/chunk-common.js"></script>
```

PR #1737, opened 2020-05-14
Show bank code field only when necessary

moiikana:
https://phabricator.wikimedia.org/T251733

### DIFF
--- a/skins/laika/src/components/shared/PaymentBankData.vue
+++ b/skins/laika/src/components/shared/PaymentBankData.vue
@@ -18,7 +18,7 @@
 				</b-input>
 			</b-field>
 		</div>
-		<div v-bind:class="['form-input', { 'is-invalid': bankDataIsInvalid }]">
+		<div v-show="isBankFieldEnabled" v-bind:class="['form-input', { 'is-invalid': bankDataIsInvalid }]">
 			<label for="bic" class="subtitle">{{ $t( labels.bic ) }}</label>
 			<b-field>
 				<b-input class="is-medium"
@@ -71,6 +71,9 @@ export default Vue.extend( {
 		isBankIdDisabled(): boolean {
 			return this.looksLikeIban();
 		},
+		isBankFieldEnabled(): boolean {
+			return this.looksLikeBankAccountNumber() || ( this.looksLikeIban() && this.bankDataIsValid );
+		},
 		bankIdentifier: {
 			get: function (): string {
 				if ( this.looksLikeGermanIban() ) {
@@ -87,7 +90,7 @@ export default Vue.extend( {
 				return {
 					iban: 'donation_form_payment_bankdata_account_iban_label',
 					bic: 'donation_form_payment_bankdata_bank_bic_label',
-					bicPlaceholder: 'donation_form_payment_bankdata_bank_bic_placeholder',
+					bicPlaceholder: 'donation_form_payment_bankdata_bank_bic_placeholder2',
 				};
 			} else if ( this.looksLikeBankAccountNumber() ) {
 				return {
@@ -104,6 +107,7 @@ export default Vue.extend( {
 		},
 		...mapGetters( NS_BANKDATA, [
 			'bankDataIsInvalid',
+			'bankDataIsValid',
 			'getBankName',
 		] ),
 	},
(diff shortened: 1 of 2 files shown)


skins/laika/src/components/shared/PaymentBankData.vue
```diff
@@ -71,6 +71,9 @@ export default Vue.extend( {
 		isBankIdDisabled(): boolean {
 			return this.looksLikeIban();
 		},
+		isBankFieldEnabled(): boolean {
+			return this.looksLikeBankAccountNumber() || ( this.looksLikeIban() && this.bankDataIsValid );
```
gbirke, line 75: :+1:

skins/laika/src/components/shared/PaymentBankData.vue
```diff
@@ -18,7 +18,7 @@
 				</b-input>
 			</b-field>
 		</div>
-		<div v-bind:class="['form-input', { 'is-invalid': bankDataIsInvalid }]">
+		<div v-show="isBankFieldEnabled" v-bind:class="['form-input', { 'is-invalid': bankDataIsInvalid }]">
```
Abban, line 21: You could add a unit test for this that checks if the element is visible or not

skins/laika/tests/unit/components/shared/BankData.spec.ts
```diff
@@ -215,6 +216,52 @@ describe( 'BankData', () => {
...
+		let bankInfoInput = wrapper.find( 'input#bic' );
+		expect( bankInfoInput.isVisible() ).toBe( false );
+
+		wrapper.setData( { accountId: 'DE89370400440532013000' } );
+
+		// this should be done differently...
+		await wrapper.vm.$nextTick();
+		await store.dispatch( NS_BANKDATA, setBankData );
+		await wrapper.vm.$nextTick();
+		// ...
```
moiikana, line 259: do you have an idea how to mock/await the validation of the iban here? @gbirke @Abban 
the desired result should be the store getter `bankdataIsValid` to be `true`. Then the bank field input should be visible.
 (alternatively: how to mock a computed value?)

Abban: If you fil the #iban field with the desired data then trigger a blur you should be able to then see if the bic field is visible on the page or not.

skins/laika/tests/unit/components/shared/BankData.spec.ts
```diff
@@ -215,6 +216,52 @@ describe( 'BankData', () => {
...
+		expect( bankInfoInput.isVisible() ).toBe( false );
+
+		wrapper.setData( { accountId: '123' } );
+		await wrapper.vm.$nextTick();
+
+		bankInfoInput = wrapper.find( 'input#bic' );
+		expect( bankInfoInput.isVisible() ).toBe( true );
+	} );
+
+	it( 'renders bank code / BIC field when valid IBAN was entered and validated', async () => {
```
gbirke, line 238: The way this test is written, it almost looks like an integration/browser test, simulating user input, server response and then desired state. I'm not sure we should go this route, because it will make the test brittle and we have a lot to set up. This test and other test can probably be simplified by setting up different store states and checking what's displayed. And by interacting with the form and check that the component dispatches the right actions to the store.

See https://vue-test-utils.vuejs.org/guides/using-with-vuex.html#mocking-with-modules (and the previous sections) for ideas on how to create a local store dummy for testing instead of using the `createStore` function.

skins/laika/tests/unit/components/shared/BankData.spec.ts
```diff
@@ -215,6 +217,83 @@ describe( 'BankData', () => {
...
+				$t: () => {},
+			},
+		} );
+		wrapper.setData( { accountId: 'DE89370400440532013000' } );
+
+		let bankInfoInput = wrapper.find( 'input#bic' );
+		expect( bankInfoInput.isVisible() ).toBe( true );
+	} );
+
+	it( 'does not render bank code / BIC field if IBAN not validated', async () => {
```
gbirke, line 268: Since this test is more about the *state* after the validation and not about the *action* of validating, I'd call this test `does not render bank code / BIC field if IBAN is invalid`. Who do you say as a native-speaker, @Abban ?

Abban: 👍

PR #1741, opened 2020-05-20
Add persistent form storage

Abban:
https://phabricator.wikimedia.org/T252379
https://phabricator.wikimedia.org/T252380


skins/laika/src/store/LocalStorageRepository.ts
```diff
@@ -0,0 +1,83 @@
...
+		if ( bytes.length % 3 === 2 ) {
+			base64String = base64String.substring( 0, base64String.length - 1 ) + '=';
+		} else if ( bytes.length % 3 === 1 ) {
+			base64String =
+				base64String.substring( 0, base64String.length - 2 ) + '==';
+		}
+
+		return base64String;
+	}
+}
```
gbirke, line 83: Wow, having a whole bas64 implementation in here seems a bit excessive. have you tried using the `btoa` and `atob` functions, together with  functions for [converting strings into array buffers](https://developers.google.com/web/updates/2012/06/How-to-convert-ArrayBuffer-to-and-from-String)?

skins/laika/src/store/create_data_persister.ts
```diff
@@ -0,0 +1,111 @@
...
+		} catch ( e ) {
+			this.repository.removeItem( `${this.keyNamespace}/${key}` );
+			return null;
+		}
+	}
+
+	getPlugin( items: DataPersistenceItem[] ) {
+		return ( store: Store<any> ) => {
+			store.subscribe( ( mutation: MutationPayload, state: any ) => {
+				const item = items.find( item => item.mutation === mutation.type );
```
gbirke, line 44: This re-use of `item` both as const and function param had me confused for a moment. Maybe change one of the names?

skins/laika/src/store/create_data_persister.ts
```diff
@@ -0,0 +1,111 @@
...
+						case DataPersistenceMutationType.KEY_VALUE_PAIR:
+							if ( item.fields.includes( mutation.payload.name ) ) {
+								this.save( mutation.payload.name, JSON.stringify( mutation.payload.value ) );
+							}
+							break;
+					}
+				}
+			} );
+		};
+	}
```
gbirke, line 59: These 6 levels of nesting are dizzying, but I don't blame you, I know it's because of the sub-optimal way we're storing our data. I wish there was a better way.

skins/laika/src/store/LocalStorageRepository.ts
Abban: I couldn't get that to work. If you check out where I [ripped the code from](https://github.com/localForage/localForage/blob/master/src/utils/serializer.js) they also talk about serialising ArrayBufers. I'll give it another look.

skins/laika/src/store/create_data_persister.ts
Abban: I'll flip the if statement and reduce one level at least

skins/laika/src/pages/membership_application.ts
```diff
@@ -27,67 +37,82 @@ interface MembershipAmountModel {
...
-			store.dispatch(
-				action( NS_BANKDATA, initializeBankData ),
-				{
-					accountId: pageData.applicationVars.initialFormValues.iban,
-					bankId: pageData.applicationVars.initialFormValues.bic || '',
-					bankName: pageData.applicationVars.initialFormValues.bankname || '',
-				}
-			)
-		);
+	if ( pageData.applicationVars.initialFormValues !== undefined ) {
```
gbirke, line 66: You lost the check for array here (see comment in the original code). It does not matter for the functionality, since your implementation in the dataInitalizer can handle arrays (Kudos for that, I liked that very much). But from a type(script) perspective, we're sloppy here.

I'd recommend keeping the comment that explains the situation.

app/Controllers/NewDonationController.php
```diff
@@ -43,7 +43,7 @@ public function handle( FunFunFactory $ffFactory, Request $request ): Response {
 				$interval,
 				$validationResult->isSuccessful(),
 				$trackingInfo,
-				$request->get( 'addressType', 'person' ),
+				$request->get( 'addressType', $ffFactory->getChoiceFactory()->getAddressType() ),
```
gbirke, line 46: Please keep the ChoiceFactory private and add a method to FunFunFactory instead. This will help removing the test because we won't have to search for ChoiceFactory usages in the code base.

PR #1748, opened 2020-06-04
Track form validation errors (client side) in web analytics

moiikana:
A new vuex store plugin was added.
It it used in the donation form (address section), membership form and the update address form.
The plugin calls a tracking function whenever the user triggers "invalid" validation state
on a form field.

https://phabricator.wikimedia.org/T252374

### DIFF
--- a/skins/laika/src/pages/donation_form.ts
+++ b/skins/laika/src/pages/donation_form.ts
@@ -18,6 +18,7 @@ import { createInitialDonationAddressValues, createInitialDonationPaymentValues 
 import LocalStorageRepository from '@/store/LocalStorageRepository';
 import { initializeAddress } from '@/store/address/actionTypes';
 import { Country } from '@/view_models/Country';
+import { createTrackFormErrorsPlugin } from '@/store/track_form_errors_plugin';
 
 const PAGE_IDENTIFIER = 'donation-form';
 const FORM_NAMESPACE = 'donation_form';
@@ -44,6 +45,7 @@ const dataPersister = createDataPersister(
 );
 const store = createStore( [
 	dataPersister.getPlugin( persistenceItems ),
+	createTrackFormErrorsPlugin( FORM_NAMESPACE ),
 ] );
 
 const i18n = new VueI18n( {
--- a/skins/laika/src/store/track_form_errors_plugin.ts
+++ b/skins/laika/src/store/track_form_errors_plugin.ts
@@ -1,0 +1,40 @@
+import { MutationPayload, Store } from 'vuex';
+import { mutation as mutationName } from '@/store/util';
+import { NS_ADDRESS, NS_MEMBERSHIP_ADDRESS } from '@/store/namespaces';
+import { trackFormValidationErrors } from '@/tracking';
+import { VALIDATE_INPUT } from '@/store/address/mutationTypes';
+import { Validity } from '@/view_models/Validity';
+
+export const trackFormErrorPlugin = ( store: Store<any>, formName: string ) => {
+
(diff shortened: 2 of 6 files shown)


skins/laika/src/store/track_form_errors_plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+		'donation_form': 'address',
+		'update_address': 'address',
+		'membership_application': 'membership_address',
+	};
+	let storeName = formNameToStoreNameLookUp[ formName ];
+
+	let prevState = { ... store.state[ storeName ].validity };
+
+	store.subscribe( ( mutation: MutationPayload, mutatedState: any ) => {
+		console.log( mutation.type );
```
Abban, line 20: ❗

moiikana: :facepalm:

skins/laika/src/store/track_form_errors_plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+import { mutation as mutationName } from '@/store/util';
+import { NS_ADDRESS, NS_MEMBERSHIP_ADDRESS } from '@/store/namespaces';
+import { trackFormValidationErrors } from '@/tracking';
+import { VALIDATE_INPUT } from '@/store/address/mutationTypes';
+import { Validity } from '@/view_models/Validity';
+
+export const trackFormErrorPlugin = ( store: Store<any>, formName: string ) => {
+
+	let formNameToStoreNameLookUp: { [key: string]: string} = {
+		'donation_form': 'address',
```
gbirke, line 11: These strings now accidentally correspond with the from name constants you provided in the entry points. To kake the meaning/semantics more clear, I think you should move those form names as consteants here, import them in the entry points and use them in the lookup table like this:

```
	const formNameToStoreNameLookUp: { [key: string]: string} = {
		[FORM_DONATION]: 'address',
                [FORM_ADDRESS_UPDATE]: 'address',
                [FORM_MEMBERSHIP]: 'membership_address'
};
```

skins/laika/src/store/track_form_errors_plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+		'update_address': 'address',
+		'membership_application': 'membership_address',
+	};
+	let storeName = formNameToStoreNameLookUp[ formName ];
```
gbirke, line 15: please use `const` instead of `let` for all values that you don't intend to change and use `let` as sparingly as possible. This will lead to to code that is more functional and immutable.

skins/laika/src/pages/donation_form.ts
```diff
@@ -32,7 +34,10 @@ interface DonationFormModel {
 }
 
 const pageData = new PageDataInitializer<DonationFormModel>( '#app' );
-const store = createStore( [ createFeatureTogglePlugin( pageData.selectedBuckets ) ] );
+const store = createStore( [
+	createFeatureTogglePlugin( pageData.selectedBuckets ),
```
gbirke, line 38: Keep in mind that #1741 removes the feature toggle Vuex plugin. If you merge this PR quickly, someone else has to deal with the merge conflicts 😈

Abban: I know you'd never do that to me!

skins/laika/src/store/track_form_errors_plugin.ts
moiikana: in his PR Abban is using those form names in the entry points as well, so I shouldn't really move them..?
https://github.com/wmde/FundraisingFrontend/pull/1741/files

PR #1751, opened 2020-06-09
Remove exception message from production error page

Abban:
https://phabricator.wikimedia.org/T249106

### DIFF
--- a/app/Bootstrap.php
+++ b/app/Bootstrap.php
@@ -83,7 +83,7 @@ class Bootstrap {
 
 		$app->error( function ( AccessDeniedException $e ) use ( $ffFactory ) {
 			return new Response(
-				$ffFactory->newAccessDeniedHtmlPresenter()->present( $e ),
+				$ffFactory->newAccessDeniedHtmlPresenter()->present( $e->getMessage() ),
 				403,
 				[ 'X-Status-Code' => 403 ]
 			);
@@ -130,7 +130,7 @@ class Bootstrap {
 			}
 
 			return new Response(
-				$ffFactory->newInternalErrorHtmlPresenter()->present( $e ),
+				$ffFactory->getInternalErrorHtmlPresenter()->present( $e ),
 				$code
 			);
 		} );
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -145,6 +145,7 @@ use WMDE\Fundraising\Frontend\Presentation\Presenters\DonorUpdateHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\ErrorPageHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\GetInTouchHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\IbanPresenter;
+use WMDE\Fundraising\Frontend\Presentation\Presenters\ExceptionHtmlPresenterInterface;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\InternalErrorHtmlPresenter;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\MembershipApplicationConfirmationHtmlPresenter;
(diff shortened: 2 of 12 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -957,12 +957,18 @@ public function newErrorPageHtmlPresenter(): ErrorPageHtmlPresenter {
 		return new ErrorPageHtmlPresenter( $this->getLayoutTemplate( 'Error_Page.html.twig' ) );
 	}
 
-	public function newInternalErrorHtmlPresenter(): InternalErrorHtmlPresenter {
-		return new InternalErrorHtmlPresenter( $this->getLayoutTemplate( 'Error_Page.html.twig' ) );
+	public function getInternalErrorHtmlPresenter(): ExceptionHtmlPresenterInterface {
+		$presenter = $this->pimple['internal_error_html_presenter'];
```
gbirke, line 961: Please use the `sharedObject` property and the `createSharedObject` function, pimple is on its way out. That way, you can also have a default presenter and leave the `setErrorPageHtmlPresenter` method empyt for the prod  and test environments.

src/Presentation/Presenters/InternalErrorHtmlPresenter.php
```diff
@@ -12,16 +12,15 @@
...
+	private TwigTemplate $template;
 
-	public function __construct( TwigTemplate $template ) {
+	public function setTemplate( TwigTemplate $template ): void {
 		$this->template = $template;
 	}
 
 	public function present( \Exception $exception ): string {
-		return $this->template->render( [ 'message' => $exception->getMessage() ] );
+		return $this->template->render( [ 'message' => false ] );
```
gbirke, line 24: You could just render without `message`, IMHO that's better than a `string|boolean` type.

src/Presentation/Presenters/ExceptionHtmlPresenterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Presentation\Presenters;
+
+use WMDE\Fundraising\Frontend\Presentation\TwigTemplate;
+
+interface ExceptionHtmlPresenterInterface {
+	public function setTemplate( TwigTemplate $template ): void;
```
gbirke, line 10: Why do you need the template injection? It carries the risk of creating invalid presenters if you forget to call it (and I think you did). `getLayoutTemplate` of FundraisingForntend is public and can be called from the environment factories.

IMO, constructor injection is always preferrable to setter injection.

src/Presentation/Presenters/DevelopmentInternalErrorHtmlPresenter.php
```diff
@@ -0,0 +1,20 @@
...
+class DevelopmentInternalErrorHtmlPresenter implements ExceptionHtmlPresenterInterface {
+
+	private TwigTemplate $template;
+
+	public function setTemplate( TwigTemplate $template ): void {
+		$this->template = $template;
+	}
+
+	public function present( \Exception $exception ): string {
+		return $this->template->render( [ 'message' => $exception->getMessage() ] );
```
gbirke, line 18: For even more developer convenience, you could also assign the stack trace ([`$exception->getTrace`](https://www.php.net/manual/de/exception.gettrace.php)) and render it.

app/Bootstrap.php
```diff
@@ -83,7 +83,7 @@ function ( Request $request, Application $app ) {
 
 		$app->error( function ( AccessDeniedException $e ) use ( $ffFactory ) {
 			return new Response(
-				$ffFactory->newAccessDeniedHtmlPresenter()->present( $e ),
+				$ffFactory->newAccessDeniedHtmlPresenter()->present( $e->getMessage() ),
```
gbirke, line 86: The access  denied presenter is still an InternalErrorHmlPresenter, which expects and exception for `present`.

PR #1754, opened 2020-06-12
Update form validation

Abban:
https://phabricator.wikimedia.org/T251736


Abban:
When you test this can you also take a look at the donation confirmation and make sure I didn't mess up any current UX?

skins/laika/src/pages/contact_form.ts
```diff
@@ -3,20 +3,21 @@ import VueI18n from 'vue-i18n';
...
 
-interface ErrorModel {
-	message: string
+interface ContactFormModel {
+	message: string,
+	contactFormValidationPatterns: ContactFormValidation
 }
 
-const pageData = new PageDataInitializer<ErrorModel>( '#app' );
+const pageData = new PageDataInitializer<ContactFormModel>( '#app' );
```
gbirke, line 20: Good catch of a copy-paste error!

When we developed laika, we started out with the error page and took its page as the template for all other pages.

src/Factories/FunFunFactory.php
```diff
@@ -1428,6 +1434,13 @@ public function getCountries(): string {
 		) )->readAndValidateJson();
 	}
 
+	public function getValidationRules(): string {
+		return ( new JsonStringReader(
+			$this->getI18nDirectory() . '/data/validation.json',
+			new SimpleFileFetcher()
+		) )->readAndValidateJson();
+	}
+
```
gbirke, line 1443: The JsonStringReader is on its way out (see https://phabricator.wikimedia.org/T254880) and its primary function is to provide *unparsed* JSON strings (instead of data structured) to the templating engine. Please change move the `json_decode` call to this method. You can still use `JsonStringReader` to make sure the JSON is valid, but you can also assume it is.

PR #1760, opened 2020-06-18
Fix "unsupported currency" paypal error

moiikana:
Checks now on the paypal payment type before verifying.

https://phabricator.wikimedia.org/T210161


app/Routes.php
```diff
@@ -406,7 +406,7 @@ function ( Request $request ) use ( $ffFactory ) {
 		$app->post(
 			'handle-paypal-payment-notification',
 			function ( Request $request ) use ( $ffFactory ) {
-				return ( new PayPalNotificationHandler( $ffFactory ) )->handle( $request );
+				return ( new PayPalNotificationController( $ffFactory ) )->handle( $request );
```
gbirke, line 409: You can clean these routes even more by making `FunFunFactory` a parameter of the `handle` methods of the controllers and giving the controller as a string here. See `AddDonationController` and `donation/add` in `Routes.php` for reference. The dependency injection framework will use the types of the `handle` parameters and automatically call it with the right instances in the right order. This will be a further step away from Silex towards Symfony.

PR #1762, opened 2020-06-19
Improve paypal error logging

moiikana:
https://phabricator.wikimedia.org/T255296

### DIFF
--- a/app/Controllers/PaypalNotificationController.php
+++ b/app/Controllers/PaypalNotificationController.php
@@ -17,6 +17,7 @@ use WMDE\Fundraising\PaymentContext\RequestModel\PayPalPaymentNotificationReques
 class PaypalNotificationController {
 
 	private const MSG_NOT_HANDLED = 'PayPal request not handled';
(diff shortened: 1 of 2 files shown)


app/Controllers/PaypalNotificationController.php
```diff
@@ -29,8 +29,10 @@ public function handle( FunFunFactory $ffFactory, Request $request ): Response {
 		try {
 			$ffFactory->getPayPalPaymentNotificationVerifier()->verify( $post->all() );
 		} catch ( PayPalPaymentNotificationVerifierException $e ) {
+			$parametersToLog = $post->all();
+			unset( $parametersToLog['payer_email'], $parametersToLog['payer_id'] );
```
Abban, line 33: Maybe move the parameter keys that need to be removed to a class constant: `const BLOCK_LIST = [ 'payer_email', 'payer_id' ]`. Then loop them and unset here. That way we can modify it easier.
```suggestion
			foreach( BLOCK_LIST as $remove ) {
				unset( $parametersToLog[ $remove ] );
			}
```

PR #1765, opened 2020-06-26
Fix form persist for IE11/Edge

Abban:
https://phabricator.wikimedia.org/T256385


skins/laika/src/view_models/DataPersistence.ts
```diff
@@ -0,0 +1,19 @@
...
+	KEY_VALUE_PAIR,
+}
+
+export interface DataPersister {
+	save( key: string, data: string ): void;
+	load( key: string ): Promise<null | string>;
+	getPlugin( items: DataPersistenceItem[] ): any;
+	decryptInitialValues( items: DataPersistenceItem[] ): Promise<void>;
+	getInitialValue( key: string ): string|null;
+}
```
gbirke, line 12: Now that I see the pure interface instead of the whole class, it looks a bit "unfocused" to me - what's the difference between `load` and `getInitialValue`?  The  `decryptInitialValues` method hints at decryption, but why is that not an implementation detail? There are several ways we can go forwards here:

* Add docblocks that explain what each method does (and roughly where in the code)
* Rename some methods to give a better hint at what they are doing
* Move some methods back into the implementations

skins/laika/src/view_models/DataPersistence.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface DataPersistenceItem {
```
gbirke, line 14: I guess this interface could also get a docblock description that explains what it is for and what its properties mean.

skins/laika/src/view_models/DataPersistence.ts
Abban: `save` and `load` are responsible for saving and loading from the repository and are only used internally (Apart from some tests). `decryptInitialValues` loads all the stored values into an internal array for use later, and they are accessed from `getInitialValue`. Why don't I refactor the naming to:

* `save` => `saveToRepository`
* `load` => `loadFromRepository`
* `decryptInitialValues` => `initialize`
* `getInitialValue` => `getValue`

gbirke: 👍

PR #1767, opened 2020-07-01
Add new Coding Style

gbirke:
This is for https://phabricator.wikimedia.org/T249489


Abban:
Should we fix the [Scrutinizer issue](https://github.com/wmde/FundraisingFrontend/pull/1767/files#diff-66c10547821bb592f8309a7df3fa6287R39)?

gbirke:
I'll do that in a separate PR, removing the class which is dead an unused anyway.

PR #1769, opened 2020-07-07
Fix visibility of country in donation summary

moiikana:
https://phabricator.wikimedia.org/T256406

### DIFF
--- a/skins/laika/src/components/DonationSummary.vue
+++ b/skins/laika/src/components/DonationSummary.vue
@@ -59,6 +59,7 @@ export default Vue.extend( {
 		'address',
 		'addressType',
 		'payment',
+		'countries',
 	],
 	methods: {
 		getSummary: function () {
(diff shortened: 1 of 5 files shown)


skins/laika/src/components/DonationSummary.vue
```diff
@@ -67,9 +68,10 @@ export default Vue.extend( {
 			const formattedAmount = this.payment.amount.toFixed( 2 ).replace( '.', ',' );
 			const paymentType = this.$t( this.payment.paymentType );
 			const personType = this.$t( addressTypeRenderer.getPersonTypeMessageKey() );
+			const countryObject = this.countries.find( c => ( c.countryCode === this.address.countryCode || c.countryCode === this.address.country ) );
```
gbirke, line 71: I was baffled by this and had to look up where the different names came from. To prevent the same reaction in our future selves, I suggest at least the following comment, to clarify:

```suggestion
                        // address.country is from store.address, address.countryCode is from donation confirmation
			const countryObject = this.countries.find( c => ( c.countryCode === this.address.countryCode || c.countryCode === this.address.country ) );
```

If you want to be really awesome & fix this, you could rename the `country` property in the address store to countryCode. But I can understand if you deem this too risky, because there are 37 occurrences of "country" in the skin code base and you'd have to determine for each of them if you need to rename it (when accessing the store) or not (for the form field name).

PR #1774, opened 2020-08-11
[WIP] Add database layer to Bucket Test Logger

Abban:
https://phabricator.wikimedia.org/T202285


src/BucketTesting/Domain/Model/BucketLog.php
```diff
@@ -0,0 +1,67 @@
...
+	private array $buckets;
+	private DateTimeInterface $date;
+
+	public function __construct( int $external_id, string $eventName, array $buckets, ?DateTime $date = null ) {
+		$this->external_id = $external_id;
+		$this->eventName = $eventName;
+		$this->date = $date ?? new DateTime();
+
+		$this->buckets = [];
+		foreach ($buckets as $bucket) {
```
gbirke, line 24: Please move creation of the `BucketLogBucket` instances to `DatabaseBucketLogger::getBucketMap` and provide the finished array to the constructor. This avoids array-key-to-object mismatches and we stay type-safe the whole way.

src/BucketTesting/Domain/Model/BucketLog.php
```diff
@@ -0,0 +1,67 @@
...
+
+class BucketLog {
+
+	private int $id;
+	private int $external_id;
+	private string $eventName;
+	private array $buckets;
+	private DateTimeInterface $date;
+
+	public function __construct( int $external_id, string $eventName, array $buckets, ?DateTime $date = null ) {
```
gbirke, line 18: If we didn't have the optional date param, a better way to type the buckets would be with the spread operator. But in this case, please provide a docblock with type information for `$buckets`.

tests/Unit/BucketTesting/Domain/Model/BucketLogTest.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Tests\Unit\BucketTesting\Domain\Model;
+
+use PHPUnit\Framework\TestCase;
+
+class BucketLogTest extends TestCase {
+
+	public function testWhenBucketLogIsCreatedAddsRelatedBuckets() {
```
gbirke, line 11: Should move into the test for `DatabaseBucketLogger`

src/BucketTesting/Domain/Model/BucketLog.php
```diff
@@ -0,0 +1,67 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\BucketTesting\Domain\Model;
+
+use DateTime;
+use DateTimeInterface;
+
+class BucketLog {
```
gbirke, line 10: Let's talk about naming (of this class) and namespace hierarchies. I like that you introduced the `Domain\Model` namespace, in one of the last commits, you should also move `Campaign`, `Bucket` and `CampaignDate` there. But if we do that, then our future selves will ask "What's the difference between a `Bucket` and a `BucketLog` and how are they related to bucket events?" 

I'd be ok with leaving class name and the logging namespace like this, but maybe we can find a better hierarchy.

src/BucketTesting/Domain/Model/BucketLog.php
```diff
@@ -0,0 +1,67 @@
...
+
+	private int $id;
+	private int $external_id;
+	private string $eventName;
+	private array $buckets;
```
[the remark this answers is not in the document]
Abban: Aha! I was wondering how typed arrays work in strict php

src/BucketTesting/Domain/Model/BucketLog.php
Abban: Hm, there's a circular dependency then as each entity will have the other in their constructors

src/BucketTesting/Domain/Model/BucketLog.php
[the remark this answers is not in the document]
gbirke: The best way to do typed arrays is the spread operator. The 2nd best are type annotations, but only an aggressively-configured PHPStan will catch those and we can't configure it aggressivley at the moment.

src/BucketTesting/Domain/Model/BucketLog.php
```diff
@@ -0,0 +1,54 @@
...
+	public function getBuckets(): array {
+		return $this->buckets;
+	}
+
+	/**
+	 * @param BucketLogBucket[] $buckets
+	 */
+	public function setBuckets( array $buckets ): void {
+		$this->buckets = $buckets;
+	}
```
gbirke, line 53: Internally, the Buckets should be an `ArrayCollection`, see https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/association-mapping.html and the setter should pass the object(s)/array on to the collection. Or use an `addBucket` method instead of a setter.

PR #1823, opened 2020-08-27
Tweak deployment

Abban:
For tasks on: https://phabricator.wikimedia.org/T259498

* Output testing pool domain when deploying to testing pool
* Add Travis CI check to make sure all group vars and config files are encrypted (file starts with $ANSIBLE_VAULT;1.1;AES256)


gbirke (approved):
I ade a followup for the encryption check: #1829

gbirke:
I'm closing this - the code will be moved to the infrastructure repo, where we won't need the Makefile checks and will have the deploy output.

PR #1878, opened 2020-09-24
A/B test: provisional address choice + email option

moiikana:
- create new campaign for a/b test
- add 'email' addresstype to the address type model
- implements the feature toggle switch for the a/b test

https://phabricator.wikimedia.org/T260023


app/config/campaigns.yml
```diff
@@ -74,3 +74,16 @@ address_type:
...
+  reference: "https://phabricator.wikimedia.org/T260023"
+  start: "2020-10-01"
+  end: "2020-12-31"
+  buckets:
+    - "provisional_address_options"
+    - "old_address_type_options"
+  default_bucket: "provisional_address_options"
+  url_key: provadd
+  active: true
+  param_only: false
```
gbirke, line 89: This should be `true` because we want to test this specifically with the banners in desktop test07. All other banner tests should get the "old style" address chooser.

skins/laika/src/components/pages/donation_form/Address.vue
```diff
@@ -81,143 +77,31 @@ export default Vue.extend( {
...
-		},
-		onFieldChange( fieldName: string ): void {
-			this.$store.dispatch( action( NS_ADDRESS, setAddressField ), this.$data.formData[ fieldName ] );
-		},
-		onAutofill( autofilledFields: { [key: string]: string; } ) {
-			Object.keys( autofilledFields ).forEach( key => {
-				const fieldName = camelizeName( key );
-				if ( this.$data.formData[ fieldName ] ) {
-					this.$store.dispatch( action( NS_ADDRESS, setAddressField ), this.$data.formData[ fieldName ] );
+	// TODO properly type 'props'
```
gbirke, line 89: Typed props would be really great here (and in `useAddressFunctions`). My first attempt to solve this would be to use `defineComponent` from the composition API + a props interface instead of using `Vue.extend`. See https://stackoverflow.com/questions/59966387/strongly-typing-props-of-vue-components-using-composition-api-and-typescript-typ

The store can stay untyped for now, that's not additional tech debt but one whe had for some time.

gbirke (commented):
There is a bug while selecting addresses: Select "Email" and then "Full address" - the email field will stay there, while it should be hidden until a user selects "person" or "company". Maybe the address type needs to be reset to "Unset" when a user selected the full address type? 

Apart from that - great work, it's coming together!

moiikana:
@Abban 
From what I can see so far, everything should be implemented for this. 
What is missing though: When wanting to donate with email-option the validation for Salutation+Name.. does not work.
To fix that I tried to add a case for `email` in `ValidateAddressController.php` but either that's not it or it's not enough.
Maybe you can find out how to make the validation work :pray: 

Gabriel made a fix in `fundraising-donations` to actually make a (valid) e-mail donation work.
To test the application incl. fix you can add
`"wmde/fundraising-donations": "dev-fix-email-only-donation",`
into the composer.json here.

gbirke:
@Abban When everything works & you deploy this to prod, please merge my donation PR, release a patch version of donations, set `composer.json` back to the major version and do `make update-php`.

skins/laika/src/store/address/constants.ts
```diff
@@ -8,6 +8,7 @@ export interface AddressRequirements {
 export const REQUIRED_FIELDS: AddressRequirements = {
 	[ AddressTypeModel.PERSON ]: [ 'salutation', 'firstName', 'lastName', 'street', 'postcode', 'city', 'country', 'email', 'addressType' ],
 	[ AddressTypeModel.COMPANY ]: [ 'companyName', 'street', 'postcode', 'city', 'country', 'email', 'addressType' ],
+	[ AddressTypeModel.EMAIL ]: [ 'email' ],
```
Abban, line 11: Here's where the validation was failing

PR #1892, opened 2020-09-29
Prepare HTTP Kernel event handling for Symfony

gbirke:
Convert all calls to `before`, `after` and `error` into HTTP event subscribers.

This is for https://phabricator.wikimedia.org/T263434


app/EventHandlers/AddIndicatorAttributeForJsonRequests.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Add indicator attribute `request_stack.is_json` to request for JSON/JSONP requests,
+ * based on the HTTP `Accept` (accepted content type) header.
+ *
+ * @todo When https://phabricator.wikimedia.org/T263436 is done, use Symfony's "_format" parameter instead,
+ *       see https://symfony.com/doc/current/routing.html#special-parameters
+ */
+class AddIndicatorAttributeForJsonRequests implements EventSubscriberInterface {
+
+	private const PRIORITY = 64;
```
Abban, line 21: Where do these priority numbers come from?

Abban (approved):

app/EventHandlers/AddIndicatorAttributeForJsonRequests.php
gbirke: Out of thin air 😬 
Silex used 512 and -512 for its methods, but such huge numbers felt too big for me, so I chose lower exponents of 2 whenever I needed a priority.

PR #1904, opened 2020-10-02
Route controllers

Abban:
https://phabricator.wikimedia.org/T263433

I will handle the redirects and home page as part of the replace Silex ticket (https://phabricator.wikimedia.org/T263436)

### DIFF
--- a/app/Controllers/AddCommentController.php
+++ b/app/Controllers/AddCommentController.php
@@ -1,0 +1,80 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\Controllers;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\DonationContext\UseCases\AddComment\AddCommentRequest;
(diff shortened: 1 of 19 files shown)


gbirke (commented):
Part of this ticket is to remove `Application` as a dependency of the controller methods, because that's coming from Silex:

* Instead of calling `$app->json` you can do `JsonResponse::create`.
* For the session access in `UpdateDonorController` and `AddDonationController` we'll have to find a different, Symfony-Compatible way to inject the session into the controller. It might be enough to add `SessionInterface $session` to the parameters of the controller methods, in the hope that the `SessionValueResolver` of the Symfony HTTP Kernel (also used by Silex) will auto-inject it. But maybe we need to add our own resolver in Bootstrap.php

gbirke (commented):
I'd like to take the opportunity to think again about the naming scheme for our controllers and their methods. At the moment, it's inconsistent (for historial reasons). Most projects choose a `NounController::action` schema, with the action being averb most of the time with the exception being the default action, which is often `index` (instead of `handle`, which is named that way because previously our controllers were called "RouteHandlers"). What are your opinions and suggestions for this, @Abban @moiikana ?

moiikana:
I'd prefer a unified switch from `handle` to `<action>[Object]`, like `::viewComment` or maybe sth like `::show` in some cases

PR #1921, opened 2020-10-07
Catch unknown addressType in initialisation

Abban:
https://phabricator.wikimedia.org/T263984

Is this is the 'correct' place to put this? I'm not sure why it was throwing errors there in the first place.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -7744,16 +7744,16 @@
         },
         {
             "name": "sebastian/type",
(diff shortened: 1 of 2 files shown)


moiikana:
@gbirke this looks fine to me, but maybe you want to have another look at if this still is ok at doing what it was supposed to, too

moiikana (approved):

PR #1934, opened 2020-10-13
Add ADR for Controller naming

gbirke:


gbirke:
This draft might be biased - while writing this, I became more and more fond of the `ActionSentenceController::action` scheme, with my preferred solution using `index` as the action name now and replacing it later with `__invoke` when we have migrated to Symfony. Feel free to add more pros and cons, fix typos and voice your opions here.

PR #1936, opened 2020-10-14
[DNM] Add cookie notice

Abban:
https://phabricator.wikimedia.org/T265255

It will currently only show the cookie notice on the donation form in non production. This will be hooked up properly in https://phabricator.wikimedia.org/T265256


skins/laika/src/components/CookieNotice.vue
```diff
@@ -0,0 +1,143 @@
...
+			textOpen,
+			toggleTextOpen,
+			onOptionalToggle,
+			onCheckButtonClick,
+			onBackButtonClick,
+			removeNotice,
+		};
+	},
+} );
+</script>
```
gbirke, line 143: This component is doing three things at once, making it a bit hard to follow:

- Act as a container, with height adjustment
- A paging mechanism (page 1 and 2, with the ability to go back and forth)
- Keeping the state of the optional buttons.

Maybe you can split this into several components, e.g. container, "paging" and 2 pages, maybe reusing the button components on both pages. This will make the template easier to read and distribute the state a bit.

gbirke (changes requested):

gbirke (approved):
I have user-tested this and saw that on wide displays, the "more" links that should expand the containers do nothing. Maybe you can only hide them if there is more content to display. See https://stackoverflow.com/questions/9333379/check-if-an-elements-content-is-overflowing for examples and hints.

gbirke (approved):

gbirke (changes requested):
Good work integrating the Matomo consent into the consent banner.

However, the consent does not work for allowing Matomo for subsequent requests:

1) Go to page as new user and give consent
2) Click on Membership or FAQ and watch the network requests in the dev tools

Expected Result: 
A request to `piwik.php` on our tracking server

Actual result: 
No requests go out

gbirke (approved):

PR #1937, opened 2020-10-14
Remove Pimple from FunFunFactory

gbirke:
This PR is for becoming more independent from the Silex framework, which uses
the Pimple Dependency Injection Container internally.

- Move shared object creation from Pimple to FunFunFactory internals
- Remove Pimple from FunFunFactory

This is for https://phabricator.wikimedia.org/T263462

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -16,8 +16,6 @@ use FileFetcher\ErrorLoggingFileFetcher;
 use FileFetcher\SimpleFileFetcher;
 use GuzzleHttp\Client;
 use NumberFormatter;
-use Pimple\Container;
-use Pimple\ServiceProviderInterface;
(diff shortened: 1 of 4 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -862,7 +710,16 @@ private function newHonorificValidator(): AllowedValuesValidator {
 	}
 
 	private function getHonorifics(): Honorifics {
-		return $this->pimple['honorifics'];
+		return $this->createSharedObject( Honorifics::class,
```
Abban, line 713: Should we move this into the content repo?

Abban (approved):

src/Factories/FunFunFactory.php
gbirke: Yep, that would improve things. Maybe not in this PR, but I'll work on this.

PR #1940, opened 2020-10-15
Update to Twig 3.0

gbirke:
* Remove dev dependencies that require an incompatible version of
  symfony/twig-bridge. We were not using them anyway.
* Require newer version of symfony/twig-bridge that is compatible with
  Twig 3.0
* Remove abandoned twig/extensions. We are not using it anymore.
* Remove TwigServiceProvider from bootstrap - we're not using the Twig
  instance initialized by Silex.

This is for https://phabricator.wikimedia.org/T265471


Abban (approved):

composer.json
```diff
@@ -11,7 +11,7 @@
 		"ext-sodium": "*",
 
 		"silex/silex": "~2.0",
-		"twig/twig": "~1.23",
+		"twig/twig": "~2.0",
```
moiikana, line 14: I bet I missed/forgot about the discussion about this in the daily or so, but why is this 2.0 and not 3.0 here?

Abban: Oops, in my review I checked all the namespace and Object updates and missed this. Good catch!

gbirke: Brilliant catch, thanks!

I guess I changed it at some point to be compatible with fundraising-content-provider and forgot to bump it.

Twig 3.0 brings some internal changes, so I'm setting this to [DNM]

moiikana (approved):

PR #1945, opened 2020-10-16
Improve web tests

gbirke:
- Make WebRouteTestCase compatible with versions >= 4.x of HttpKernel.
- Remove createSilexApplication method because it returns a
  Silex\Application object.
- Always use NullDomainNameValidator in tests. This simplifies dependency
  injection and avoids network requests in tests that validate emails.


Abban (approved):
All good, just sent a follow up PR (https://github.com/wmde/FundraisingFrontend/pull/1965) to remove some unused namespaces I spotted

PR #1958, opened 2020-10-20
[DNM] Cookie Notice Preferences Functionality

Abban:
https://phabricator.wikimedia.org/T265256


tests/Unit/Infrastructure/EventHandling/DomainEventHandler/BucketLoggingHandlerTest.php
```diff
@@ -86,4 +121,12 @@ public function getBuckets(): array {
 		$campaign->addBucket( new Bucket( 'test1', $campaign, true ) );
 		return $campaign->getBuckets();
 	}
+
+	private function createBucketLoggingHandler( EventDispatcher $eventDispatcher, BucketLogger $logger ): BucketLoggingHandler {
```
gbirke, line 125: I like that you did not include the consent state in the params - while that adds a line to the test setup, it makes it really explicit what the consent is.

gbirke (changes requested):
Works well & thanks for adding tests. I'll do a followup PR for increasing readability/understandability.

There is one thing you need to change, though: The cookie banner needs to appear on every page, not only on the donation form. that probably means adding the `cookieNoticeVisible` property to every entry point 😢

PR #1974, opened 2020-10-28
Rename variables in test

gbirke:
Use `expected` prefix to make it clear that this variable contains an
expected value and not a setup value


Abban (approved):
Merging this even with the failing tests as the fail comes from an issue in the main PR

PR #1998, opened 2020-11-03
Fix import of wrong Request class

gbirke:
Controllers should always get an instance of Symfony\Component\HttpFoundation\Request

### DIFF
--- a/app/Controllers/StaticContent/ShowUseOfFundsController.php
+++ b/app/Controllers/StaticContent/ShowUseOfFundsController.php
@@ -4,8 +4,8 @@ declare( strict_types = 1 );
 
 namespace WMDE\Fundraising\Frontend\App\Controllers\StaticContent;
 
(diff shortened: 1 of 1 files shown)


app/Controllers/StaticContent/ShowUseOfFundsController.php
```diff
@@ -4,8 +4,8 @@
 
 namespace WMDE\Fundraising\Frontend\App\Controllers\StaticContent;
 
+use Symfony\Component\HttpFoundation\Request;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
-use WMDE\Fundraising\PaymentContext\DataAccess\Sofort\Transfer\Request;
```
Abban, line 8: I bet this came from a PhpStorm autofill

Abban (approved):

PR #2006, opened 2020-11-11
Prevent validation issue when loading from local file

moiikana:
https://phabricator.wikimedia.org/T267007

### DIFF
--- a/skins/laika/templates/Base_Layout.html.twig
+++ b/skins/laika/templates/Base_Layout.html.twig
@@ -28,13 +28,16 @@
 </head>
 <body>
 {% block main %}
-    <div id="app" data-application-vars="{$ _context|json_encode|e('html_attr') $}" 
-    data-application-messages="{$ translations()|e('html_attr') $}" 
-    data-assets-path="{% if assets_path %}{$ assets_path $}{% else %}{$ basepath $}/skins/laika{% endif %}">
+    <div id="app">
         <noscript>
-                <h2>{$ site_metadata.no_javascript_header $}</h2>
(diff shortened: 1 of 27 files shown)


skins/laika/templates/Base_Layout.html.twig
```diff
@@ -28,7 +28,8 @@
 </head>
 <body>
 {% block main %}
-    <div id="app" data-application-vars="{$ _context|json_encode|e('html_attr') $}" 
+    <div id="app" />
```
Abban, line 31: This element needs a closing tag, otherwise the browser thinks the appdata one is a child and it gets removed when the app gets mounted.

skins/laika/templates/Base_Layout.html.twig
```diff
@@ -28,7 +28,8 @@
...
+    <div id="appdata" data-application-vars="{$ _context|json_encode|e('html_attr') $}"
```
Abban, line 32: This needs to be set to `display: none;` or the flex css rules will make it a column

skins/laika/templates/Base_Layout.html.twig
```diff
@@ -28,7 +28,8 @@
...
     data-application-messages="{$ translations()|e('html_attr') $}" 
     data-assets-path="{% if assets_path %}{$ assets_path $}{% else %}{$ basepath $}/skins/laika{% endif %}">
         <noscript>
```
Abban, line 35: The noscript tag should be inside `#app`

gbirke (approved):

PR #2013, opened 2020-11-27
Fix PSR-4 autoloading of migrations

gbirke:
PHPStan failed on the Doctrine migration files which were not included
correctly in the composer autoload information.

Ran `composer update` to create new version hash for `composer.lock`


Abban (approved):
Wonder if this wasn't noticed before because we manually ran them?

gbirke:
> Wonder if this wasn't noticed before because we manually ran them?

I guess so,  but the cause might also be some tighter checking by PHPStan or the new composer version.

PR #2014, opened 2020-11-27
Change SubmissionRateLimit storage to Session

gbirke:
Store the submission rate limit in Session instead of cookie, to avoid
having to parse time formats from cookie and to avoid sending extraneous
data to user.

Added session manipulation methods to WebRouteTestCase.
Adding session access to WebRouteTestCase means the Silex Application
object will be persisted, leading to increased memory usage during unit
tests. To alleviate that, we increase the PHP memory limit in the
Makefile.

Previously, AddDonationRouteTest and ApplyForMembershipRouteTest checked
the rate limit cookie for the correct cookie parameters. I've removed
that check for ApplyForMembershipRouteTest and added a generalized check
to AddDonationRouteTest. This test is skipped for now, because it will
fail until we use CookieBuilder everywhere. This change is already on
the horizon with https://github.com/wmde/FundraisingFrontend/pull/1936
When that that PR is merged, we should remove the `markTestAsSkipped`
instruction and try again.

In general AddDonationRouteTest is a bad place to ensure all cookies
have the right parameters, we only (mis)use that test because we're
fairly sure that it is the route that sets the most cookies.

This is for https://phabricator.wikimedia.org/T268766

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,11 +68,11 @@ covers:
 	docker-compose run --rm --no-deps app ./vendor/bin/covers-validator
 
 phpunit:
-	docker-compose run --rm app ./vendor/bin/phpunit $(TEST_DIR)
+	docker-compose run --rm app php -d memory_limit=1G vendor/bin/phpunit $(TEST_DIR)
 
 phpunit-with-coverage:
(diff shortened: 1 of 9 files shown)


Abban (approved):

gbirke:
It's a pity you didn't squash this, I'd like to keep fix commits out of the history if they aren't standalone PRs

PR #2017, opened 2020-12-07
Fix bug in payment type selection

Abban:
https://phabricator.wikimedia.org/T269363

This was introduced in the b side of the provisional_address_options test.
Commit: https://github.com/wmde/FundraisingFrontend/commit/aa85a47ba5db9a898ebcfd4dbcd4a0ad27593d7e
Ticket: https://phabricator.wikimedia.org/T260023#6507727

### DIFF
--- a/skins/laika/src/components/pages/donation_form/Payment.vue
+++ b/skins/laika/src/components/pages/donation_form/Payment.vue
@@ -32,7 +32,7 @@
 					:payment-types="paymentTypes"
 					:error="typeIsValid ? '' : $t('donation_form_payment_type_error')"
 					:title="$t('donation_form_payment_type_title')"
-					:disabled-payment-types="[]"
+					:disabled-payment-types="emailOnlyDisabledPaymentTypes"
 					slot="campaigns.address_provision_options.provisional_address_options"
 					v-on:payment-type-selected="sendTypeToStore"
 			></payment-type>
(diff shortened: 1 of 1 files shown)


gbirke (approved):

skins/laika/src/components/pages/donation_form/Payment.vue
```diff
@@ -32,7 +32,7 @@
 					:payment-types="paymentTypes"
 					:error="typeIsValid ? '' : $t('donation_form_payment_type_error')"
 					:title="$t('donation_form_payment_type_title')"
-					:disabled-payment-types="[]"
+					:disabled-payment-types="emailOnlyDisabledPaymentTypes"
```
moiikana, line 35: why not use `disabledPaymentTypes` here?

Abban: It was to do with this from the linked commit:

>when navigating back to payment page (in email bucket): don't disable any payment option based on address type selection (implemented with feature toggle)

Basically in the B version we don't want the addressType selection to make paymentTypes disabled. A side effect of that was it was also not making Sofort disabled when a non 0 interval was selected. We can't use the computed `disabledPaymentTypes` as that also disables based on addressType, so a new one that only disables based on the interval selected was needed.

PR #2018, opened 2020-12-07
Drop local tables on make setup

Abban:
This fixes errors thrown when the local tables already exist


gbirke (approved):
I'm worried that at some point we might have container in production and someone will accidentally run `make setup`. I know it's far-fetched, but for a command that is run not often, maybe rather document the task?

Abban:
Closing this, I'm going to add another command specifically for rebuilding the local environment instead.

PR #2019, opened 2020-12-15
Add incentive checkbox

Abban:
https://phabricator.wikimedia.org/T270068

Adds an incentive checkbox to the membership application. To see it you need to append `?incentive=1` to the url


src/Factories/FunFunFactory.php
```diff
@@ -1202,9 +1203,17 @@ public function getMembershipApplicationFormTemplate(): TwigTemplate {
 			'countries' => $this->getCountries(),
 			'addressValidationPatterns' => $validation->address,
 			'dateOfBirthValidationPattern' => $validation->dateOfBirth,
+			'incentives' => $this->getChoiceFactory()->getDefaultIncentives()
```
gbirke, line 1206: Do we still need this top-level value here?  I don't see any access from applicationdata, IFAIK it's only in `initialformValues`.

gbirke (approved):

src/Factories/FunFunFactory.php
Abban: Oh, no we don't I left it in by accident

gbirke (approved):
Errors and CI are fixed, thanks

PR #2020, opened 2020-12-15
Add New Donation Confirmation Layout

Abban:
https://phabricator.wikimedia.org/T268649


moiikana:
not so sure what the exact state of this ticket is, 
but the DonationSummary component seems to be a bit broken, there used to be a wrapping text
Bildschirmfoto vom 2021-01-08 13-19-02

moiikana (approved):

PR #2023, opened 2020-12-21
Add correct autocomplete value to city field

gbirke:
Previously the autocomplete attribute just had `on`, which prevented
proper autocompletion.

`address-level1` is probably the correct value for cities in Germany,
but pending https://stackoverflow.com/q/65395168/130121 .


gbirke:
Closing, because
- Vue files no longer in this repo
- Nobody complained about autocompletion
- SO question was never answered properly

PR #2028, opened 2021-01-14
Silex to symfony

gbirke:


gbirke:
This is now ready for review. 
You can start at the "Improve Readme" commit, those commits happened after our shared code review.

To deploy, this needs the changes in https://github.com/wmde/fundraising-infrastructure/pull/324

web/index.php
```diff
@@ -2,12 +2,16 @@
...
 
+use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
 use Symfony\Component\HttpFoundation\Request;
 use WMDE\Fundraising\Frontend\App\Kernel;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
+// We set the env from the .env file instead of web server settings because we want to be independent from the hosting provider
+( new Dotenv() )->bootEnv( __DIR__ . '/../.env' );
```
Abban, line 13: Thought you didn't want to use .env files?

web/index.php
```diff
@@ -15,7 +15,7 @@
 $env = $_SERVER['APP_ENV'] ?? 'dev';
 $debug = $_SERVER['APP_DEBUG'] ? (bool)$_SERVER['APP_DEBUG'] : false;
 
-if ( $debug || 1 == 1 ) {
```
Abban: This was me 😬

Abban (approved):

web/index.php
gbirke: env files are not the best way to keep secrets, but with our environment (web server config controlled by hoster, no containers where you could pass them in) and the requirement that we want to be able to change them without involving the hoster, I think they're the best solution we have at the moment. If you have suggestions on different approaches, I'd like to hear them, but for the time being (i.e. this PR) we go with env files.

I'll have another look at the deploy playbook to set the permissions as minimal as possible (only readable by deploy uder and www-data group).

moiikana:
Requests with common valid input values at 
https://testing05.wikimedia.customers.manitu.net/contact/get-in-touch
https://testing05.wikimedia.customers.manitu.net/apply-for-membership
return a 500 after submit for me.

The cookie banner does not get hidden after clicking on its buttons.

PR #2053, opened 2021-01-18
Change test setup

gbirke:
Change how the environment is bootstrapped and set up in WebRouteTestCase, to prepare for the Symfony update.

### DIFF
--- a/tests/EdgeToEdge/Routes/UpdateAddressRouteTest.php
+++ b/tests/EdgeToEdge/Routes/UpdateAddressRouteTest.php
@@ -21,8 +21,9 @@ class UpdateAddressRouteTest extends WebRouteTestCase {
 	private const DUMMY_DONATION_ID = 0;
 
 	public function testWhenInvalidDataIsSent_serverThrowsAnError(): void {
+		$this->modifyConfiguration( [ 'skin' => 'laika' ] );
 		$this->createEnvironment(
-			[ 'skin' => 'laika' ],
+			[],
 			function ( Client $client, FunFunFactory $factory ): void {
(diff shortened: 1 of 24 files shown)


tests/EdgeToEdge/Routes/UpdateAddressRouteTest.php
```diff
@@ -21,8 +21,8 @@ class UpdateAddressRouteTest extends WebRouteTestCase {
 	private const DUMMY_DONATION_ID = 0;
 
 	public function testWhenInvalidDataIsSent_serverThrowsAnError(): void {
+		$this->modifyConfiguration( [ 'skin' => 'laika' ] );
```
Abban, line 24: Why is the skin is required for these?

moiikana (approved):

tests/EdgeToEdge/Routes/UpdateAddressRouteTest.php
gbirke: The test environment and configuration use the skin `test` by default. When we had server-side rendering, the templates of the `test` skin made it easy to ensure that the controllers and presenter classes assigned the right key and values to the templates, without HTML getting in the way. The "new way" of doing things came with the `laika` skin where we could call `$this->getDataApplicationVars()` to get the data directly from the JSON data attribute. That's why all tests that were written after introducing the laika skin now set laika as default. 

There is a plan to get rid of the `test` skin all together and migrate the existing tests, ditching the call to `modifyConfiguration` and make `laika` the new standard skin in route tests: https://phabricator.wikimedia.org/T163452 



Because the `test` skin that's used by default (and sadly still used by many tests) is on its way out. Previously the progress

PR #2068, opened 2021-01-25
Remove code for legacy parameter handling

moiikana:
https://phabricator.wikimedia.org/T272840


gbirke (approved):

gbirke (approved):
I have user-tested it and the start page looks like before.

PR #2069, opened 2021-01-25
Conditional Travis build

gbirke:


gbirke:
This approach doesn't work because Travis has no `master` branch and does a shallow clone instead. An alternative option could be to run the file change counters as a git hook and save them as a [git note](https://stackoverflow.com/questions/2683248/can-i-add-metadata-to-git-commits-or-can-i-hide-some-tags-in-gitk). But since there is no "quick win" here, I'll be closing this PR soon.

gbirke:
Closing because we will use user means of improving the build

PR #2085, opened 2021-02-04
Make PHPUnit fail faster

gbirke:
This change will halt a PHPUnit run if it encounters errors or failed
tests. This will speed up our CI and local testing.

In my day-today work I rarely need the information which tests fail and
I'm rather interested in fast feedback.


gbirke:
@Abban @moiikana This is a suggestion that came from the experiences with PHPUnit, please object if you find the previous behavior - running all tests regardless of failure/error and then showing a count of all failed/errored tests - more useful.

moiikana:
how about we add a 
`fast-test: covers phpunit --stop-on-failure --stop-on-error`
or something to the makefile instead? 
I feel like seeing the big whole picture could be the better way to go as a default.

Abban:
I agree with @moiikana, for consistency I'd rather keep `make ci` running the way it already does as that's how it is on the other repositories. Fail on first error could be another command or an argument.

PR #2094, opened 2021-02-10
Allow html in cookie banner

Abban:
Allow html in the cookie banner content

https://phabricator.wikimedia.org/T274097


moiikana (changes requested):
this change leads to the visibilityToggle element not being displayed anymore

moiikana (approved):

PR #2145, opened 2021-03-10
Remove incentive AB test related code

moiikana:
- archives the membership application incentive campaign
- makes incentive option the new default

https://phabricator.wikimedia.org/T276216

### DIFF
--- a/src/Factories/ChoiceFactory.php
+++ b/src/Factories/ChoiceFactory.php
@@ -63,14 +63,4 @@ class ChoiceFactory {
 		}
 		throw new UnknownChoiceDefinition( 'Address type configuration failure.' );
 	}
-
-	public function getDefaultIncentives(): array {
-		if ( $this->featureToggle->featureIsActive( 'campaigns.membership_incentive.incentive' ) ) {
-			// TODO: This is currently hardcoded, we should make the list of incentives and defaults configurable
-			return [ 'tote_bag' ];
-		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.membership_incentive.no_incentive' ) ) {
-			return [];
-		}
-		throw new UnknownChoiceDefinition( 'Address type configuration failure.' );
-	}
 }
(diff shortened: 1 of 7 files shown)


src/Factories/ChoiceFactory.php
```diff
@@ -65,12 +65,7 @@ public function getAddressType(): ?string {
...
 	public function getDefaultIncentives(): array {
-		if ( $this->featureToggle->featureIsActive( 'campaigns.membership_incentive.incentive' ) ) {
-			// TODO: This is currently hardcoded, we should make the list of incentives and defaults configurable
-			return [ 'tote_bag' ];
-		} elseif ( $this->featureToggle->featureIsActive( 'campaigns.membership_incentive.no_incentive' ) ) {
-			return [];
-		}
-		throw new UnknownChoiceDefinition( 'Address type configuration failure.' );
+		// TODO: This is currently hardcoded, we should make the list of incentives and defaults configurable
+		return [ 'tote_bag' ];
```
gbirke, line 69: Since the responsibility of `ChoiceFactory` is to create campaign-related instances/values, the workflow for removing campaigns is to replace all calls to the  `ChoiceFactory` method with with the static value/initialization and remove the  `ChoiceFactory` method.

gbirke (changes requested):

src/Factories/ChoiceFactory.php
moiikana: ok. I removed it, I hope I got you right

src/Factories/FunFunFactory.php
```diff
@@ -1206,14 +1206,14 @@ private function getMembershipApplicationFormTemplate(): TwigTemplate {
 			'countries' => $this->getCountries(),
 			'addressValidationPatterns' => $validation->address,
 			'dateOfBirthValidationPattern' => $validation->dateOfBirth,
-			'incentives' => $this->getChoiceFactory()->getDefaultIncentives()
+			'incentives' => [ 'tote_bag' ]
```
gbirke, line 1209: It's slightly dangerous to have the incentives as an array in 2 places, you could wrap them in a constant or method (and maybe even use it in the test). But that's just a nitpick, no blocker.

gbirke (approved):

PR #2158, opened 2021-04-06
Show asset indicator in dev mode

gbirke:
Include an additional CSS file in dev mode that shows how assets are
included (dev server or builtin)


moiikana (approved):
thank you very much :pray:

PR #2166, opened 2021-05-19
Adapt to changes in memberships V5 release

gbirke:
The memberships bounded context introduced a number of
backward-incompatible changes which get addresses in this commit.


gbirke:
This needs a 5.1 release with the changes from https://github.com/wmde/fundraising-memberships/pull/71, then change the `composer.json` version and run `make update-php`

PR #2192, opened 2021-06-01
Update Docker database initialisation

Abban:
This is an alternate MySQL image/database initialisation process. It allows us to easier import the third party Geodata.

https://phabricator.wikimedia.org/T280955


.gitignore
```diff
@@ -29,3 +29,7 @@ web/skins
 deployment/deployment.retry
 
 tmp/
+
+/build/database/*
+!/build/database/01.Database Schema.sql
+!/build/database/02.Default Data.sql
```
gbirke, line 35: Please use underscores instead of spaces. This will avoid the need for escaping when referring to these files on the command line.

doc/HOWTO_Import_Geodata.md
```diff
@@ -0,0 +1,8 @@
+# How to Import the Third Party Geodata
+
+The Geodata zip contains 2 sql files, a small one with community names, and a large one which includes streets.
+
+1. Copy the smaller sql file into /build/database.
+2. Rename it to `00.[FILENAME].sql`. This is needed because Docker runs the files in order, and the Geodata import needs to run before the Schema.
```
gbirke, line 6:
```suggestion
2. Rename it to `00.Geodata.sql`. This is needed because Docker runs the files in order, and the Geodata import needs to run before the Schema.
```
I'm all for indicating that a file name is irrelevant, but using a placeholder here might trip up inexperienced developers who might name that file literally `00.[FILENAME].sql`

doc/HOWTO_Import_Geodata.md
```diff
@@ -0,0 +1,8 @@
...
+3. Remove the database volume `$ docker-compose volume rm fundraising-app_db-storage`.
```
gbirke, line 7:
```suggestion
3. Remove the database volume with the command  
    `docker volume rm APP_DIRECTORY_db-storage`. 
    APP_DIRECTORY is the directory name where you have checked out the `fundraising-application` code. You can check for the volume name by listing all docker volumes with the command  
    `docker volume ls`
```

gbirke (changes requested):
I like the simplicity and explicitness of the file-based approach.

Keeping the DB schema in a checked-in SQL file has two drawbacks:

1) It can get out of sync when we change the schema
2) It creates changes in PRs that don't need to be reviewed (because it's just a generated file)

The ideal solution would be to generate the file on the fly with `./vendor/bin/doctrine orm:schema-tool:create`, but that might not be possible if that commend needs a running database (at which point the DB has already read the directory). So I'm willing to go with the checked-in file. That also has the benefit of making the schema changes clear for the more SQL inclined.

To avoid the file becoming stale, please add a Makefile task that runs `./vendor/bin/doctrine orm:schema-tool:create`, writes the result to a temporary file (e.g. `./var/tmp/db-schema.sql`) and compares it to `01. Database Schema.sql`. The task should fail if `diff` has a non-zero return code. Then add that task to CI.

We then also need some documentation on how to update the SQL file when the schema changes (running `./vendor/bin/doctrine orm:schema-tool:create` with the right output destination). You might want to turn the documentation into a general "How the development environment creates the schema and database fixtures" description, mentioning the automatic creation feature of the docker image, our directory name and add sections for how to regenerate when the CI fails due to schema changes.

doc/HOWTO_Import_Geodata.md
```diff
@@ -0,0 +1,12 @@
+# How to Import the Third Party Geodata
+
+The Geodata zip contains 2 sql files, a small one with community names, and a large one which includes streets.
+
+1. Copy the smaller sql file into /build/database.
+2. Rename it to `00.GEodata.sql`. This is needed because Docker runs the files in order, and the Geodata import needs to run before the Schema.
```
gbirke, line 6:
```suggestion
2. Rename it to `00.Geodata.sql`. This is needed because Docker runs the files in order, and the Geodata import needs to run before the Schema.
```

doc/HOWTO_Import_Geodata.md
```diff
@@ -0,0 +1,12 @@
...
+3. Remove the database volume with the command:
+   
+   ```$ docker volume rm APP_DIRECTORY_db-storage```
+   
+   APP_DIRECTORY is the directory name where you have checked out the `fundraising-application` code. You can check for the volume name by listing all docker volumes with the command `$ docker volume ls`.
+4. Run `$ docker-compose up`. When importing the Geotata this might take a little while as it imports.
```
gbirke, line 12:
```suggestion
4. Run `$ docker-compose up`. When importing the Geodata, this might take several minutes.
```

gbirke (approved):

Abban:
I could also add a line in the docs about running `docker-compose down -v` which will remove all volumes, so it should only be run if you don't care about them all being rebuilt.

Abban:
@dependabot rebase

PR #2210, opened 2021-06-09
[WIP] Impression data repository

Abban:
https://phabricator.wikimedia.org/T284430


gbirke (changes requested):
Please move these files to the backend repo, they don't make sense in the fundraising Application.

PR #2214, opened 2021-06-14
Add PayPal IPN Queue ADR

Abban:
Thia is the ADR from the earlier discussion

https://phabricator.wikimedia.org/T284893

### DIFF
--- a/doc/adr/020_PayPal_IPN_Queue.md
+++ b/doc/adr/020_PayPal_IPN_Queue.md
@@ -1,0 +1,58 @@
+# PayPal IPN Queue
+
+Date: 2021-06-14
+
+Deciders: Kai Nissen, Gabriel Birke, Corinna Hillebrandt, Abban Dunne, Conny Kawohl
+
+Technical Story: https://phabricator.wikimedia.org/T284893
+
+## Status
+
+Declined
+
+## Context and Problem Statement
+
+The PayPal IPNs on Fundraising Frontend started to fail after a deployment 
+and was only noticed some days after when the Project Manager needed to 
+export the data. Upon investigation, it was discovered that:
+
+* Error logging was inactive on the application. (Now fixed)
+* We can’t debug using the responses our system returned to PayPal as 
+  we don’t have access to the IPN log.
+
+This led to a situation where we couldn't get the information required 
+to debug the error. It was suggested we queue all incoming requests from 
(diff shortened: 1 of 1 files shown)


doc/adr/020_PayPal_IPN_Queue.md
```diff
@@ -0,0 +1,53 @@
...
+## Decision Drivers
+
+* **Transparency**: If our system fails we would have a stored queue to 
+  use for debugging.
+* **Automation**: The IPNs wouldn't need to be fired again once an error 
+  becomes fixed as our system would resume processing the queue.
+
+## Pros and Cons
+
+* Good, because debugging would be easier.
```
gbirke, line 36:
```suggestion
* Good, because debugging would be easier - we could use existing data and run it against our code without involvement from PayPal.
```

doc/adr/020_PayPal_IPN_Queue.md
```diff
@@ -0,0 +1,53 @@
...
+* Good, because our system would resume processing automatically.
+* Good, because PayPal stops sending all IPNs when it starts to receive
```
gbirke, line 38:
```suggestion
* Good, because these is less logic that can go wrong, which means PayPal
  will get fewer server errors. When PayPal receives too many server errors, 
  it will stop sending IPNs. 
```

doc/adr/020_PayPal_IPN_Queue.md
```diff
@@ -0,0 +1,53 @@
...
+* **Automation**: The IPNs wouldn't need to be fired again once an error 
+  becomes fixed as our system would resume processing the queue.
+
+## Pros and Cons
+
+* Good, because debugging would be easier.
+* Good, because our system would resume processing automatically.
+* Good, because PayPal stops sending all IPNs when it starts to receive 
+  html 500 errors.
+* Bad, another system to support.
```
gbirke, line 40:
```suggestion
* Bad, another system to support.
* Bad, because we send back each IPN to PayPal to verify that it's genuine.
  It's unclear how PayPal would behave if there is a time delay or 
  a duplicate verification (when re-processing old items).
```

gbirke (approved):
Thank you for writing this!

I've added some clarifications and a "deal breaker" bad thing (we verify received IPNs with Paypal)

PR #2218, opened 2021-06-18
Improve incentives code

gbirke:
Change how the controller gets incentive data
Add regression test for https://phabricator.wikimedia.org/T285101


gbirke:
Regression test will pass when we use a new version of the bounded context. We could use this PR to update it.

PR #2224, opened 2021-06-23
Add Symfony Cache

Abban:
Doctrine v2.9 has removed caching so we need to
use the Symfony library instead.

Also:
* Remove unused cache clearing route and code.
* Update CampaignConfigurationLoader with new caching


src/Factories/EnvironmentSetup/ProductionEnvironmentSetup.php
```diff
@@ -9,23 +9,30 @@
...
+		$this->setCampaignCache( $factory );
 		$this->setPaypalLogger( $factory );
 		$this->setSofortLogger( $factory );
 		$this->setCreditCardLogger( $factory );
 		$this->setDoctrineConfiguration( $factory );
 	}
 
+	private function setCampaignCache( FunFunFactory $factory ) {
+		$factory->setCampaignCache(
+			new Psr16Cache( new FilesystemAdapter( 'campaigns', 3600, $factory->getCachePath() . '/campaigns' ) )
```
gbirke, line 32: Since we need to deploy for new campaign files anyway, you can set the campaign cache to 24 hours or even 1 week

src/Factories/EnvironmentSetup/ProductionEnvironmentSetup.php
```diff
@@ -9,23 +9,30 @@
...
 class ProductionEnvironmentSetup implements EnvironmentSetup {
 
 	public function setEnvironmentDependentInstances( FunFunFactory $factory ) {
 		$this->initializeLoggers( $factory );
-
-		$factory->enableCaching();
 	}
 
 	private function initializeLoggers( FunFunFactory $factory ) {
+		$this->setCampaignCache( $factory );
```
gbirke, line 23: This is not a logger, please move the method call to `setEnvironmentDependentInstances`

gbirke (approved):

PR #2227, opened 2021-06-28
Add Locale Switching

Abban:
This adds backend functionality for swapping locales and a new soft dependency in the FunFunFactory

Ticket: https://phabricator.wikimedia.org/T256612

### DIFF
--- a/app/EventHandlers/StoreLocale.php
+++ b/app/EventHandlers/StoreLocale.php
@@ -1,0 +1,51 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\EventHandlers;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use WMDE\Fundraising\Frontend\App\CookieNames;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+
+class StoreLocale implements EventSubscriberInterface {
+
+	private const PRIORITY = 512;
+
+	private FunFunFactory $factory;
+	private array $allowedLocales;
+
+	public function __construct( FunFunFactory $factory, array $allowedLocales ) {
+		$this->factory = $factory;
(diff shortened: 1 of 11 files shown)


app/EventHandlers/StoreLocale.php
```diff
@@ -0,0 +1,50 @@
...
+		$request->setLocale( $locale );
+		$this->factory->setLocale( $locale );
+	}
+
+	private function findLocale( Request $request ): string {
+		if ( $request->cookies->has( 'locale' ) ) {
+			return $request->cookies->get( 'locale' );
+		}
+
+		return $request->getPreferredLanguage();
```
gbirke, line 48:
```suggestion
		return $request->getPreferredLanguage() ?? '';
```

gbirke (changes requested):
While we saw that `StoreLocale` was implicitly tested in our E2E test, please create a separate unit test that tests the fallback mechanism.

src/Factories/FunFunFactory.php
```diff
@@ -504,7 +502,7 @@ private function getDefaultTwigVariables(): array {
 		return [
 			'honorifics' => $this->getHonorifics()->getList(),
 			'piwik' => $this->config['piwik'],
-			'locale' => $this->config['locale'],
+			'locale' => $this->getLocale(),
```
gbirke, line 505: You can drop this, locale can now be determined on the client side.

src/Factories/FunFunFactory.php
```diff
@@ -772,7 +770,7 @@ public function getInternalErrorHtmlPresenter(): ExceptionHtmlPresenterInterface
 			'Error_Page.html.twig',
 			[
 				'piwik' => $this->config['piwik'],
-				'locale' => $this->config['locale'],
+				'locale' => $this->getLocale(),
```
gbirke, line 773: You can drop this, locale can now be determined on the client side.

gbirke (approved):

gbirke:
For a successful deploy, please make a PR in fundraising-infrastructure that removes the `locale` key from the configs

Abban:
> For a successful deploy, please make a PR in fundraising-infrastructure that removes the `locale` key from the configs

Locale is defaulted by the dist config and not overridden by the ones in infrastructure AFAICT

PR #2229, opened 2021-06-30
[DNM] Pass max donation amount to front end

Abban:
This is to allow the frontend to validate amounts coming from banners without having to ping the server.

Ticket: https://phabricator.wikimedia.org/T262925


Abban:
This won't be needed if I do https://phabricator.wikimedia.org/T239349

PR #2235, opened 2021-07-01
Remove validate amount route

Abban:
Amount validation has been moved to the front end so we not longer need this.

Ticket: https://phabricator.wikimedia.org/T239349


gbirke (approved):
Yay, 90 lines deleted!

PR #2246, opened 2021-07-08
[DNM] Localise Currencies

Abban:
The frontend needs to know the current locale
tag in order for VueI18n to localise currencies

Ticket: https://phabricator.wikimedia.org/T284918


gbirke:
This PR is unneccessary:

In https://github.com/wmde/fundraising-application/pull/2240/commits/a716c31048b3bcce5171f4ce4a36628302704b58 I removed the locale, because the frontend already knows it through the cookie

The correct (locale-based) contents for the data are already read by `FunFunFactory::getI18nDirectory()`

PR #2247, opened 2021-07-09
Localise numbers in emails

Abban:
This adds Symfony and Twig intl classes to allow locale based currency formatting in mail

Ticket: https://phabricator.wikimedia.org/T284918


src/Factories/TwigFactory.php
```diff
@@ -94,7 +95,9 @@ protected function getFunctions(): array {
 	 * @return AbstractExtension[]
 	 */
 	protected function getExtensions(): array {
-		return [];
+		return [
```
gbirke, line 98: Please move this to the MailerTemplatingFactory, we don't format the numbers in the web Twig.

gbirke (changes requested):

gbirke:
I haven't user-tested this because we don't have the translation files in content yet. We'll test everything once they are available

gbirke (approved):

PR #2286, opened 2021-08-10
Improve address change errors

gbirke:
Show a translated message to the user if they try to change their
address again.

Log the token of failed attempts to be able to track the number of
attempts.

Deployed to https://testing05.wikimedia.customers.manitu.net

Not a proper fix for https://phabricator.wikimedia.org/T288200 but at least we have friendlier error messages.

### DIFF
--- a/app/Controllers/AddressChange/ShowUpdateAddressController.php
+++ b/app/Controllers/AddressChange/ShowUpdateAddressController.php
@@ -15,18 +15,17 @@ use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
  */
 class ShowUpdateAddressController {
 
-	public const ADDRESS_CHANGE_SESSION_KEY = 'address_changed';
-
 	public function index( Request $request, FunFunFactory $ffFactory ): Response {
 		$addressToken = $request->get( 'addressToken', '' );
 		if ( $addressToken === '' ) {
-			throw new AccessDeniedException( 'No address change token in request' );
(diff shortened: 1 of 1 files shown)


app/Controllers/AddressChange/ShowUpdateAddressController.php
```diff
@@ -15,18 +15,17 @@
  */
 class ShowUpdateAddressController {
 
-	public const ADDRESS_CHANGE_SESSION_KEY = 'address_changed';
-
 	public function index( Request $request, FunFunFactory $ffFactory ): Response {
 		$addressToken = $request->get( 'addressToken', '' );
 		if ( $addressToken === '' ) {
-			throw new AccessDeniedException( 'No address change token in request' );
+			throw new AccessDeniedException( 'address_change_no_token_in_request' );
```
moiikana, line 21: does the actual translation of that key happens automatically here?

gbirke: Yes, the [error handler passes the message key to the frontend](https://github.com/wmde/fundraising-application/blob/main/app/EventHandlers/HandleExceptions.php#L54)  which [translates the message](https://gitlab.com/fun-tech/fundraising-app-frontend/-/blob/main/src/components/pages/Error.vue)
 I've already added the key to the frontend content repo

Abban (approved):

PR #2289, opened 2021-08-11
Improve DatabaseBucketLogger

gbirke:
A suggested improvement

### DIFF
--- a/src/BucketTesting/Logging/DatabaseBucketLogger.php
+++ b/src/BucketTesting/Logging/DatabaseBucketLogger.php
@@ -5,7 +5,7 @@ namespace WMDE\Fundraising\Frontend\BucketTesting\Logging;
 use WMDE\Fundraising\Frontend\BucketTesting\Domain\BucketLoggingRepository;
 use WMDE\Fundraising\Frontend\BucketTesting\Domain\Model\Bucket;
 use WMDE\Fundraising\Frontend\BucketTesting\Domain\Model\BucketLog;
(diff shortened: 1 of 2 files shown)


moiikana (changes requested):

src/BucketTesting/Logging/DatabaseBucketLogger.php
```diff
@@ -40,4 +40,10 @@ private function addBucketLogBuckets( BucketLog $bucketLog, array $buckets ): vo
 			);
 		}
 	}
+
+	private function campaignIsNotRunning( Campaign $campaign ): bool
+	{
+		$now = new CampaignDate();
+		return $campaign->isActive() && !$campaign->isExpired( $now );
```
moiikana, line 47:
```suggestion
		return !$campaign->isActive() || $campaign->isExpired( $now );
```
not sure if it was even correct before, but this here might be easier to understand?
the campaign being not active (->true) leads to campaign not running
or
the campaign being expired (->) true leads to campaign not running

moiikana (changes requested):

src/BucketTesting/Logging/DatabaseBucketLogger.php
gbirke: You're right. Originally, I named this `campaignIsActive` and inverted your condition. But then I had the negation on the call, so i renamed it campaignIsNotActive and forgot to invert the condition again. Luckily, your unit test caught this!

PR #2313, opened 2021-09-17
Remove unused Piwik class and code

Abban:
There was some legacy tracking code that is
now removed.

Ticket: https://phabricator.wikimedia.org/T290561


gbirke (approved):
I guess the `composer.lock` update was accidental? I'm not against running `make update-php` instead of `make install-php` when starting a new branch (it will reduce dependabot PRs), but it should be a separate commit to express the intent of a PR better - if the composer change is part of a refactoring or feature commit it suggests that the update was necessary for that commit.

Abban:
Oops, yeah that was a mistake.

PR #2320, opened 2021-09-21
Add matomo goal ids to config

Abban:
We use different site and goal ids in test
and production so this adds them to the
environment configuration.

Ticket: https://phabricator.wikimedia.org/T290561

### DIFF
--- a/app/config/schema.json
+++ b/app/config/schema.json
@@ -626,20 +626,32 @@
       "type": "object",
       "title": "Piwik settings",
       "properties": {
-          "baseUrl": {
-            "type": "string",
-            "title": "Tracker URL",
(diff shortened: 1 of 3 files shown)


app/config/schema.json
```diff
@@ -626,20 +626,32 @@
...
+          "type": "integer",
+          "title": "Donation Confirmation Goal ID",
+          "description": "Unique id (configured in Piwik)",
+          "default": 1
+        },
+        "membershipApplicationConfirmationGoalId": {
+          "type": "integer",
+          "title": "Membership Application Confirmation Goal ID",
+          "description": "Unique id (configured in Piwik)",
+          "default": 1
```
gbirke, line 653:
```suggestion
          "default": 2
```

gbirke (approved):
Thanks!

PR #2329, opened 2021-09-29
Add salutations from content

Abban:
This adds a salutations object that initialises
using a new json file in fundraising content

Tickets:
* https://phabricator.wikimedia.org/T291356
* https://phabricator.wikimedia.org/T291357


src/Presentation/Salutations.php
```diff
@@ -0,0 +1,62 @@
...
+
+class Salutations {
+
+	/**
+	 * @var array
+	 */
+	private array $salutations;
+
+	public function __construct( array $salutations ) {
+		if ( !$this->isValid( $salutations ) ) {
```
gbirke, line 19: This will validate the salutations on almost every request. I think in the case of salutations, where the value does not change often, but is heavily used, this is a bad tradeoff between security and performance, as negligible as the performance cost of this validation might be. Could you move the validity check to the frontend-content-repo and instead describe the array shape of the constructor parameter? 

 I like your use of Symfony Validators

src/Factories/FunFunFactory.php
```diff
@@ -684,6 +686,15 @@ function () {
 			} );
 	}
 
+	private function getSalutations(): Salutations {
+		return $this->createSharedObject( Salutations::class,
+			function () {
+				$json = ( new SimpleFileFetcher() )->fetchFile( $this->getI18nDirectory() . '/data/salutations.json' );
+				$data = json_decode( $json, true, 16, JSON_THROW_ON_ERROR );
+				return new Salutations( $data[ 'salutations' ] );
```
gbirke, line 694: The data in `salutations.json` could also be an array, without the need for a wrapper object. Or did you have any future extensions in mind that would benefit form an object?

src/Presentation/Salutations.php
```diff
@@ -0,0 +1,62 @@
...
+
+		$this->salutations = $salutations;
+	}
+
+	public function getList(): array {
+		return $this->salutations;
+	}
+
+	public function getSalutation( string $value ): ?array {
+		$data = array_filter( $this->salutations, fn( $salutation ) => $salutation['value'] == $value );
```
gbirke, line 31: Will we use `getSalutation` in our PHP code? If we foresee using it, we could optimize this lookup by using a key-value object in for salutations, where we drop `value` and it as key for the rest of the data instead.

gbirke (changes requested):

src/Factories/FunFunFactory.php
Abban: I was trying to get a little type safety

src/Presentation/Salutations.php
Abban: Yeah, we need it for the email greetings

src/Infrastructure/Translation/GreetingGenerator.php
```diff
@@ -4,87 +4,68 @@
...
+
+		return $this->translator->trans( $salutationConfig['greetings']['lastNameInformal'], [
+			'%spacedTitle%' => self::getSpacedTitle( $title ),
+			'%firstName%' => $firstName,
+			'%lastName%' => $lastName,
+		] );
+	}
+
+	private function getSalutationConfig( ?string $salutation ): ?array {
+		return $salutation ? $this->salutations->getSalutation( $salutation ) : null;
```
gbirke, line 69: I always try to avoid `null` values. The following method will also guarantee a result so you can get rid of the `|| $salutationConfig` checks and have calls like  `$this->translator->trans( $this->getSalutationTranslationKey( $salutation, 'formal' )`

```suggestion
	private function getSalutationTranslationKey( ?string $salutation, string $greetingType ): string {
	        if (!$salutation ) { return $this->genericGreeting; }
	        return $this->salutations->getSalutation( $salutation )['greetings'][$greetingType];
```

It also has the benefit of being more explicit what the return value is (a translation key) and masks the nested array structure.

composer.json
```diff
@@ -133,7 +133,7 @@
 
 		"wmde/psr-log-test-doubles": "~2.1",
 		"wmde/fundraising-phpcs": "~3.0",
-		"wmde/fundraising-frontend-content": "dev-test",
+		"wmde/fundraising-frontend-content": "dev-add-salutations",
```
gbirke, line 136: Don't forget to put this back before merging

gbirke (approved):

PR #2333, opened 2021-10-04
Upgrade PHPStan to level 2

gbirke:
Add a baseline file for a Symfony-related error. We could fix that with
the Symfony plugin, but that might introduce other errors.


gbirke:
Superseded by #2335

PR #2336, opened 2021-10-11
Add locale setting through URL parameter

Abban:
We need to allow locale to be set from banners
so allow it to come from a url if the cookie
doesn't already exist.

Ticket: https://phabricator.wikimedia.org/T292735

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -133,7 +133,7 @@
 
 		"wmde/psr-log-test-doubles": "~2.1",
 		"wmde/fundraising-phpcs": "~4.0",
-		"wmde/fundraising-frontend-content": "dev-test",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -133,7 +133,7 @@
 
 		"wmde/psr-log-test-doubles": "~2.1",
 		"wmde/fundraising-phpcs": "~4.0",
-		"wmde/fundraising-frontend-content": "dev-test",
+		"wmde/fundraising-frontend-content": "dev-english",
```
moiikana, line 136: set back to dev-test before merging

manicki (approved):

PR #2346, opened 2021-11-05
Move Rate Limit Cookie to Confirmation Page

Abban:
This is to ensure that the cookie is not set
on redirect to an external payment provider,
because if the user returns from one without
completing the payment they become rate limited.

Ticket: https://phabricator.wikimedia.org/T292710

### DIFF
--- a/tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
+++ b/tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
@@ -5,6 +5,7 @@ declare( strict_types = 1 );
 namespace WMDE\Fundraising\Frontend\Tests\EdgeToEdge\Routes;
 
 use Symfony\Component\BrowserKit\AbstractBrowser as Client;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use WMDE\Fundraising\DonationContext\Domain\Model\Donation;
 use WMDE\Fundraising\DonationContext\Tests\Data\ValidDonation;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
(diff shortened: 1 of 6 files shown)


tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
```diff
@@ -148,4 +149,26 @@ public function testWhenNoDonation_accessIsDenied(): void {
...
+				'id' => $this->donation->getId(),
+				'accessToken' => self::CORRECT_ACCESS_TOKEN
+			]
+		);
+
+		/** @var SessionInterface $session */
+		$session = static::getContainer()->get( 'session' );
+		$donationTimestamp = $session->get( FunFunFactory::DONATION_RATE_LIMIT_SESSION_KEY );
+		$this->assertNotNull( $donationTimestamp );
+		$this->assertEqualsWithDelta( time(), $donationTimestamp->getTimestamp(), 5.0, 'Timestamp should be not more than 5 seconds old' );
```
gbirke, line 171: If you really want to decouple the code from external services, you could use [wmde/Clock](https://github.com/wmde/Clock) as a dependency of `DonationSubmissionRateLimiter` and swap the `SystemClock` with a `StubClock` in this test. But that's not a must (because it would also make the test setup more complicated and add new setters and getters to FunFunFactory).

gbirke (approved):
Please remove `SessionInterface` from AddDonationController (as suggested by Scrutinizer) , then this is ready to go.

tests/EdgeToEdge/Routes/ShowDonationConfirmationRouteTest.php
Abban: I think it'll be okay as it is, it's a very basic class

PR #2348, opened 2021-11-11
Add Address Change API routes

Abban:
Ticket: https://phabricator.wikimedia.org/T288200

### DIFF
--- a/app/Controllers/API/Donation/AddressChangeController.php
+++ b/app/Controllers/API/Donation/AddressChangeController.php
@@ -1,0 +1,94 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\Controllers\API\Donation;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\AddressChangeContext\UseCases\ChangeAddress\ChangeAddressRequest;
+use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
+
+class AddressChangeController extends AbstractApiController {
+
+	public const ADDRESS_TYPE_PERSON = 'person';
+	public const ADDRESS_TYPE_COMPANY = 'company';
+
+	private const MESSAGE_NO_IDENTIFIER = 'address_change_no_identifier_in_request';
+	private const MESSAGE_TOKEN_NOT_FOUND = 'address_change_token_not_found';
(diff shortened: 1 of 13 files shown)


moiikana (approved):

app/Controllers/API/Donation/AddressChangeController.php
```diff
@@ -0,0 +1,94 @@
...
+	public const ADDRESS_TYPE_PERSON = 'person';
+	public const ADDRESS_TYPE_COMPANY = 'company';
+
+	private const MESSAGE_NO_IDENTIFIER = 'address_change_no_identifier_in_request';
+	private const MESSAGE_TOKEN_NOT_FOUND = 'address_change_token_not_found';
+	private const MESSAGE_EMPTY_BODY = 'address_change_empty_request_body';
+	private const MESSAGE_FAILED = 'address_change_failed';
+
+	public function show( Request $request, FunFunFactory $ffFactory, string $identifier, ?string $previousIdentifier ): Response {
+		$readAddressChangeUseCase = $ffFactory->newReadAddressChangeUseCase();
```
gbirke, line 25: This method is missing

app/Controllers/API/Donation/AddressChangeController.php
```diff
@@ -0,0 +1,94 @@
...
+
+	public const ADDRESS_TYPE_PERSON = 'person';
+	public const ADDRESS_TYPE_COMPANY = 'company';
+
+	private const MESSAGE_NO_IDENTIFIER = 'address_change_no_identifier_in_request';
+	private const MESSAGE_TOKEN_NOT_FOUND = 'address_change_token_not_found';
+	private const MESSAGE_EMPTY_BODY = 'address_change_empty_request_body';
+	private const MESSAGE_FAILED = 'address_change_failed';
+
+	public function show( Request $request, FunFunFactory $ffFactory, string $identifier, ?string $previousIdentifier ): Response {
```
gbirke, line 24: Putting in the FuncFunFactory is kind of "oversharing" and dangerously close to the service locator antipattern.
It's convenient, because we don't need to add every use case to the `services.yaml`, so we can leave this as-is for now.
In the future, I'd like to have a [CompilerPass](https://symfony.com/doc/current/service_container/compiler_passes.html) for our container that iterates over all methods of FunFunFactory and adds all the use cases created by its public methods to the container.

gbirke (changes requested):

app/Controllers/API/Donation/AddressChangeController.php
Abban: What do you mean?

gbirke: When I checked, the method was not implemented in `FunFunFactory`. I think I implemented it in one of my additional PRs. Now it's in.

gbirke (approved):

PR #2349, opened 2021-11-13
Improve PayPalPaymentNotificationVerifier

gbirke:
When PayPal sends an invalid verification response, include the response
in the exception message. Adapt the unit test.

Add docblocks
Add types
Refactor if-statements to make the exceptional case more explicit.

### DIFF
--- a/src/Infrastructure/Payment/PayPalPaymentNotificationVerifier.php
+++ b/src/Infrastructure/Payment/PayPalPaymentNotificationVerifier.php
@@ -8,6 +8,11 @@ use GuzzleHttp\Client;
 use GuzzleHttp\RequestOptions;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Infrastructure/Payment/PayPalPaymentNotificationVerifier.php
```diff
@@ -69,19 +82,21 @@ public function verify( array $request ): void {
 		}
 
 		$responseBody = trim( $result->getBody()->getContents() );
+		if ( $responseBody === 'VERIFIED' ) {
+			return;
+		}
+
```
Abban, line 88: You could set the error message string here based on the $responseBody and only define the PayPalPaymentNotificationVerifierException throw once.

Abban (approved):

PR #2350, opened 2021-11-17
Add configuration for C21_WMDE_Test_12

Abban:
This test hides the cookie banner for
some users. In order to do this we need
to force the storage of the AB test
cookies for them.

Ticket: https://phabricator.wikimedia.org/T295869


gbirke (commented):
Cookies are set as expected and the code works well.

We'll have to check with our PM if we're allowed to store the the buckets of the other campaigns when we use the override. It's not a big deal since it's "dead" data that's logged nowhere.

However, there is another hitch with this ticket: When users haven't consented, the Matomo keyword and campaign won't be written into the donation data blob (key `tracking`) which will screw up the analysis. Let's talk about this during the daily. Maybe Kai has more insights or requirements.

moiikana (approved):

PR #2351, opened 2021-11-18
Add Symfony cache warmer

Abban:
This command will be run on deployment to
compile the Twig templates rather than them
being lazy compiled at runtime.

https://phabricator.wikimedia.org/T295617

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -89,6 +89,11 @@ services:
         class: WMDE\Fundraising\Frontend\Infrastructure\CookieBuilder
         factory: ['@WMDE\Fundraising\Frontend\Factories\FunFunFactory', 'getCookieBuilder']
 
+    WMDE\Fundraising\Frontend\Infrastructure\SkinCacheWarmer:
+        class: WMDE\Fundraising\Frontend\Infrastructure\SkinCacheWarmer
+        tags:
+            - { name: kernel.cache_warmer }
+
     app.errbit_logger:
         class: Airbrake\MonologHandler
         factory: ['WMDE\Fundraising\Frontend\Factories\ErrbitLoggerFactory', 'createErrbitHandler']
--- a/src/Infrastructure/SkinCacheWarmer.php
+++ b/src/Infrastructure/SkinCacheWarmer.php
@@ -1,0 +1,55 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\Infrastructure;
+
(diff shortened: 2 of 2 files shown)


cli/CacheWarmingCommand.php
```diff
@@ -0,0 +1,76 @@
...
+			}
+		}
+		catch ( Error $e ) {
+			return Command::FAILURE;
+		}
+
+		return Command::SUCCESS;
+	}
+
+	private function getTemplates(): \ArrayIterator {
```
gbirke, line 44:
```suggestion
       /**
        * @return ArrayIterator<string>
        */
	private function getTemplates(): \ArrayIterator {
```

For future-proof support of higher PHPStan levels

cli/CacheWarmingCommand.php
```diff
@@ -0,0 +1,76 @@
...
+		foreach ( $loader->getNamespaces() as $namespace ) {
+			foreach ( $loader->getPaths( $namespace ) as $path ) {
+				$templates = array_merge( $templates, $this->findTemplatesInDirectory( $path, $namespace ) );
+			}
+		}
+
+		return new \ArrayIterator( array_unique( $templates ) );
+	}
+
+	private function findTemplatesInDirectory( string $dir, string $namespace ): array {
```
gbirke, line 60:
```suggestion
       /**
        * @return string[]
        */
	private function findTemplatesInDirectory( string $dir, string $namespace ): array {
```

For future-proof support of higher PHPStan levels

```

gbirke (approved):
Code works and looks good.

While looking at the console commands I realized that Symfony has its own cache warming infrastructure and that we're building parallel structures. I'd be OK with the PR, since Twig for HTML templates is on its way out (and we probably won't get cache stampede for mail templates), so we can delete this script in the future.

However, if you want to make this really shine, change the class from a Symfony command to a cache warmer implementation. You can lift most of the implementation from https://github.com/symfony/twig-bundle/blob/d386aaa46d1afe5afb51b39675fc2ab206159206/CacheWarmer/TemplateCacheWarmer.php We'd keep our custom Twig initialization via FunFunFactory and getting the template iterator. This might introduce problems with the unset locale, but maybe we can overcome them.
After that, you need to tag this class as a cache warmer in the services file: https://symfony.com/doc/current/reference/dic_tags.html#kernel-cache-warmer

gbirke (approved):
Thank you so much for your changes!

gbirke (approved):

PR #2352, opened 2021-11-25
Add address option for C21_WMDE_Test_15

Abban:
Ticket: https://phabricator.wikimedia.org/T296390

### DIFF
--- a/app/config/campaigns.yml
+++ b/app/config/campaigns.yml
@@ -70,6 +70,7 @@ campaigns:
         - "direct"
         - "multistep"
         - "multistep_var"
+        - "multistep_checkmarks"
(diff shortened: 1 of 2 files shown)


gbirke (approved):
This is getting confusing and I'm wondering if we're not better off making this two separate campaigns with their own URL key. that way you can end one campaign independently from another and don't have to wonder which one is running and which is not. I know that this will cause some repetition of `<feature-toggle>` on the frontend, but the increased clarity and separation might be worth it.

Abban:
It'll mean nested `feature-toggles` which I'm not sure will work. A better solution might be be to remove the unused ones (if they are unused)

PR #2357, opened 2021-12-01
Add BannerSubmissionTracker

Abban:
We need to log campaign/keyword statistics
when a user submits a banner and is sent to
donations/add.

Ticket: https://phabricator.wikimedia.org/T296627

### DIFF
--- a/app/EventHandlers/TrackBannerDonationRedirects.php
+++ b/app/EventHandlers/TrackBannerDonationRedirects.php
@@ -1,0 +1,107 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace WMDE\Fundraising\Frontend\App\EventHandlers;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * When a user makes an anonymous donation from a banner the data gets posted directly to
+ * add donation which forwards them to the external payment processor. They're then
+ * returned to the confirmation page on a successful payment. That causes the
+ * campaign tracking information to be dropped from the URL and messes up our statistics.
+ * This event handler adds temporary session values that are restored to the url and
+ * deleted from the session upon a successful payment.
+ */
+class TrackBannerDonationRedirects implements EventSubscriberInterface {
+
+	public const PIWIK_CAMPAIGN = 'piwik_campaign';
+	public const PIWIK_KWD = 'piwik_kwd';
+
+	private SessionInterface $session;
+	private string $submissionRoute;
+	private string $confirmationRoute;
+	private string $bannerSubmissionUrlParameter;
+
+	public static function getSubscribedEvents() {
+		return [
+			KernelEvents::REQUEST => 'onKernelRequest',
+			KernelEvents::RESPONSE => 'onKernelResponse'
+		];
+	}
+
+	public function __construct( SessionInterface $session, string $submissionRoute, string $confirmationRoute, string $bannerSubmissionUrlParameter ) {
+		$this->session = $session;
+		$this->submissionRoute = $submissionRoute;
+		$this->confirmationRoute = $confirmationRoute;
+		$this->bannerSubmissionUrlParameter = $bannerSubmissionUrlParameter;
+	}
+
+	public function onKernelRequest( RequestEvent $event ): void {
+		$request = $event->getRequest();
+
+		switch ( $request->attributes->get( '_route' ) ) {
+			case $this->submissionRoute:
+				$this->storeCampaignParameters( $request );
+				break;
+			case $this->confirmationRoute:
+				$this->restoreCampaignParameters( $event, $request );
+				break;
+		}
+	}
+
+	public function onKernelResponse( ResponseEvent $event ): void {
+		$request = $event->getRequest();
+
+		if ( $request->attributes->get( '_route' ) !== $this->submissionRoute ) {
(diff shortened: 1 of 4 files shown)


src/Infrastructure/MatomoBannerSubmissionTracker.php
```diff
@@ -0,0 +1,19 @@
...
+class MatomoBannerSubmissionTracker implements BannerSubmissionTracker {
+
+	private \MatomoTracker $matomoTracker;
+
+	public function __construct( \MatomoTracker $matomoTracker ) {
+		$this->matomoTracker = $matomoTracker;
+	}
+
+	public function track( string $campaign, string $keyword ): void {
+		$this->matomoTracker->setAttributionInfo( json_encode( [ $campaign, $keyword ] ) )
```
Abban, line 16: Thinking about it, I think the Matomo call should go into a try/catch and we try log if it fails or something. Eg if the tracker goes down and this code crashes the user won't be redirected to the payment service.

app/EventHandlers/TrackBannerDonationRedirects.php
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	public function onKernelRequest( RequestEvent $event ): void {
+		$request = $event->getRequest();
+
+		if ( !in_array( $request->attributes->get( '_route' ), $this->submissionRoutes ) ) {
+			return;
+		}
+
+		if ( !$request->get( $this->urlParameter ) ) {
```
gbirke, line 37: This `if` statement will fail more often than the route check (since most banners and donors don't skip the address page). Could you put this at the top and the route check below it? That might save us some nanoseconds and micrograms of CO2.

gbirke (approved):

app/EventHandlers/TrackBannerDonationRedirects.php
```diff
@@ -0,0 +1,88 @@
...
+			KernelEvents::REQUEST => 'onKernelRequest',
+			KernelEvents::RESPONSE => 'onKernelResponse'
+		];
+	}
+
+	public function __construct( SessionInterface $session, string $submissionRoute, string $confirmationRoute, string $bannerSubmissionUrlParameter ) {
+		$this->session = $session;
+		$this->submissionRoute = $submissionRoute;
+		$this->confirmationRoute = $confirmationRoute;
+		$this->bannerSubmissionUrlParameter = $bannerSubmissionUrlParameter;
```
gbirke, line 36:
```suggestion
		$this->isBannerSubmissionUrlParameter = $bannerSubmissionUrlParameter;
```
Just a suggestion for naming this thing better - it's an an URL parameter that indicates that the submission comes from a banner.

app/EventHandlers/TrackBannerDonationRedirects.php
```diff
@@ -0,0 +1,88 @@
...
+		$request = $event->getRequest();
+
+		if ( $request->attributes->get( '_route' ) !== $this->submissionRoute ) {
+			$this->session->remove( self::PIWIK_CAMPAIGN );
+			$this->session->remove( self::PIWIK_KWD );
+		}
+	}
+
+	private function storeCampaignParameters( Request $request ): void {
+		if ( !$request->get( $this->bannerSubmissionUrlParameter ) ) {
```
gbirke, line 62: The methods are all aptly named and make this handler very readable. Maybe we can turn it up a notch.

```suggestion
		if ( $this->requestWasSubmittedFromBanner($request) ) {
```

app/EventHandlers/TrackBannerDonationRedirects.php
```diff
@@ -0,0 +1,88 @@
...
+		if ( $this->hasQueryTracking( $request ) ) {
+			return;
+		}
+
+		$campaignKey = self::PIWIK_CAMPAIGN;
+		$keywordKey = self::PIWIK_KWD;
+		$campaign = $this->session->get( $campaignKey );
+		$keyword = $this->session->get( $keywordKey );
+
+		$url = $request->getRequestUri() . "&{$campaignKey}={$campaign}&{$keywordKey}={$keyword}";
```
gbirke, line 80: We know that the request URI will always have params. If we wanted to be extra sure and guard against future changes, we could do

```suggestion
		$separator = $request->getQueryString() ? '&' : '?';
		$url = $request->getRequestUri() . $separator . "{$campaignKey}={$campaign}&{$keywordKey}={$keyword}";
```

But that would also mean another test, so just pointing it out, not requesting a change

config/services.yaml
```diff
@@ -51,14 +51,24 @@ services:
             - ["handle_paypal_payment_notification", "handle_paypal_membership_fee_payments"]
             - "@logger"
 
-        # Inject the log path and route into the PayPal IPN Logger
+    # Inject the log path and route into the PayPal IPN Logger
```
gbirke, line 54: This comment is lying btw. (copypaste error) No need to change it in this PR, but we should change that at some point.

tests/EdgeToEdge/TrackBannerDonationRedirectsTest.php
```diff
@@ -0,0 +1,202 @@
...
+	private const URL_PARAMETER = 'banner_submission';
+	private const TEST_CAMPAIGN = 'test_campaign';
+	private const TEST_KEYWORD = 'test_keyword';
+
+	private SessionInterface $session;
+	private KernelBrowser $client;
+	private Donation $donation;
+
+	public function setUp(): void {
+		$this->session = new SessionSpy();
```
gbirke, line 35: This could spare you the lengthy re-implementation of the session interface:

```suggestion
		$this->session = new Session(new MockArraySessionStorage());
```

gbirke (approved):
Haven't user-tested it, but code looks really good. I like how it reads and expresses intent.

Maybe add a docblock to the event handler that explains what it does and why it exists.

config/services.yaml
Abban: Probably don't event need the comment

app/EventHandlers/TrackBannerDonationRedirects.php
Abban: That variable contains the name of the parameter though, rather than the value

gbirke (approved):

PR #2392, opened 2022-01-18
Update to Symfony 6.0

gbirke:
Update most Symfony dependencies to 6.0


gbirke:
Config and filesystem are currently held back by `phpmd`. They're preparing a new release that is compatible. But since phpmd is not part of our CI, we could also remove it. Fixing all the messages it drops would be too much effort IMHO

PR #2413, opened 2022-02-07
Fix-double encoding of city lookup

gbirke:
Our geodata table for finding cities is already UTF-8 encoded, so we
can pass thru the strings we get from the database, without re-encoding
them.

The change to the integration test is of dubious value: If our database
had the wrong encoding, the storage round trip of special characters
would probably come out fine. I've added it none the less, so show
*potential* behaviour.

This fixes https://phabricator.wikimedia.org/T300132


moiikana (approved):

Abban (approved):
I think this is a leftover from when we were loading the cities from a JSON file in the prototype.

PR #2472, opened 2022-05-17
Use MariaDB in dev environment

gbirke:
This brings the dev environment in sync with the production environment.

Tested with the MariaDB database instead of sqlite, as described in the
README

Update the README, replacing MySQL with MariaDB and fixing out-of-date
information.


Abban (approved):
We could add a command to the makefile (`make up-app-phpmyadmin`) that starts the environment with an extra compose override (`docker-compose-phpmyadmin.yml`) that starts a pre-configured PhpMyAdmin service connected to the database one. It'd be an easy way for a dev to browse the data in their browser.

gbirke:
@moiikana I've added some more changes related to the database for the README. Please have a look if it's helpful or if you have more questions

PR #2489, opened 2022-06-09
Add fallback for donation id to PP IPN

Abban:
If the item_number parameter is not set the
code will check for it in the custom parameter
instead.

Ticket: https://phabricator.wikimedia.org/T310117


app/Controllers/Payment/PaypalNotificationController.php
```diff
@@ -108,6 +108,15 @@ private function newUseCaseRequestFromPost( ParameterBag $postRequest ): PayPalP
 			->setPaymentType( $postRequest->get( 'payment_type', '' ) );
 	}
 
+	private function getDonationId( ParameterBag $postRequest ): int {
+		$itemId = (int)$postRequest->get( 'item_number', 0 );
```
gbirke, line 112:
```suggestion
		$itemId = $postRequest->getInt( 'item_number' );
```

gbirke (approved):

PR #2516, opened 2022-07-07
Fix "make clean" command

gbirke:
Shell commands with wildcards must be passed as a command string to a
shell, otherwise the wildcard may be interpreted in a wrong context.


moiikana (approved):

Abban:
👍

PR #2517, opened 2022-07-07
Integrate moderation into Fundraising App

moiikana:
- upgrade environment to php 8.1
- adapt mailing to new moderation reason handling

https://phabricator.wikimedia.org/T306703


moiikana:
do I have to comment out the phpstan analysis lines in the .travis settings file for now? like in the other refactoring repos, because of php8.1

gbirke:
> do I have to comment out the phpstan analysis lines in the .travis settings file for now? like in the other refactoring repos, because of php8.1

You should not need to. In the fundraising app we're using our custom-built PHPStan Docker image. 
As far as I can see, the current error comes from the unadapted `docker-compose.debug.yml` file that's used for coverage. I've updated our dev images to PHP 8.1 yesterday, so it should work on the next Travis run

Abban (approved):

PR #2521, opened 2022-07-14
Update Donation Routes

Abban:
Updates the cc/sofort/paypal IPN handlers to use the new payment domain.

Ticket: https://phabricator.wikimedia.org/T307892


gbirke (changes requested):
Looking at the coverage of the controllers (using `docker-compose -f docker-compose.yml -f docker-compose.debug.yml run --rm --no-deps -e XDEBUG_MODE=coverage app_debug php -d memory_limit=1G vendor/bin/phpunit --configuration=phpunit.xml.dist --coverage-html coverage`), I found two things:

* The tests for the credit card and sofort notification do not test the all the unhappy paths. 
* The PayPal test does not cover the logging half of `logResponseIfNeeded`. You can drop that method - errors are now logged by the various `if`/`catch` blocks.

I'd like to pair with you to see if we can refactor the controller methods to make them shorter and more readable. I have some ideas on this, but would only implement them if you like them and agree that they make the code more readable.

You can delete the `SofortNotificationRequestTest` and the PHPStan exception for it, as [discussed in the chat ](https://mattermost.wikimedia.de/swe/pl/biidz7s4tfbddqts6cge9i5u4c)

gbirke (approved):

PR #2527, opened 2022-07-26
Update confirmation mail intro

Abban:
Pass in PaymentType to allow it to show a different message per type.

Ticket: https://phabricator.wikimedia.org/T312075

Content commit is here: https://github.com/wmde/fundraising-frontend-content/commit/f63339a61d7b71c5f1e44ecdcea498870dfbdd4a

### DIFF
--- a/app/MailTemplates.php
+++ b/app/MailTemplates.php
@@ -230,6 +230,13 @@ class MailTemplates {
 						'moderationFlags' => [
 							MembershipModerationIdentifier::MANUALLY_FLAGGED_BY_ADMIN->name => true
 						]
+					],
+					'payment_type_UEB' => [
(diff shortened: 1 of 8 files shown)


gbirke (changes requested):
Please move the `if` statements from content to the main template. In all mailers we follow the pattern that we put all the branching statements and logic in the application template and the content should contain the pure snippets (and maybe some variable replacement). You might want to add new content files to the content repo. The reason for this pattern is that we can test the permutations of the logic in the application (using the `MailTemplates` class) and to keep logic and syntax of templates away from the non-dev content editors.



Please also add test cases to the `MailTemplates` class. You might also remove the Paypal test cases there.

gbirke (approved):

PR #2528, opened 2022-07-27
Add paymenttype.json translations to membership confirmation page

moiikana:
The frontend code needs the paymentTypes.json translation file to translate
the payment type that should get shown on the membership confirmation page

belongs to ticket: https://phabricator.wikimedia.org/T312072


gbirke (approved):
Could you also add the payment types to the `ApplyForMembershipController`? Abban will need those for his ticket.

gbirke (approved):

PR #2529, opened 2022-07-27
Update payment context

gbirke:


gbirke:
We're ready to deploy to test, but don't want to merge yet

PR #2551, opened 2022-08-03
Update all Symfony dependencies to 6.1

gbirke:


gbirke:
Will now be done in #2565

PR #2553, opened 2022-08-11
[DNM] Improve fee validation

gbirke:
Our controller validates more than just the fee - it's a combination of
member type, interval and amount (to calculate upper bounds). It also
checks for the allowed payment type.

Previously, this controller was fee-related and the frontend assumed
that the validation was for a wrong fee. Now we need to pass back more
detailed field that has validation errors, so we can mark the right
fields as invalid in the client-side code.

This is for https://phabricator.wikimedia.org/T312071


moiikana (approved):

Abban (approved):

moiikana:
did a rebase with main

PR #2601, opened 2022-10-19
Convert encoding for PayPal requests

gbirke:
Convert data that PayPal sends us from ISO-8859-1 to UTF-8. This is not the best solution - we should have some indication of what encoding PayPal actually sends us.


gbirke:
Closing this PR, as our approach of re-encoding string broke the validation. Salvaged the logger improvements in #2604

PR #2603, opened 2022-10-21
Fix PayPal encoding errors without automated re-encoding

gbirke:


gbirke:
Deployed branch as a "Quick-Fix-For-The-Quick-Fix" to production.

PR #2604, opened 2022-10-21
Improve PayPal logging

gbirke:
Send errors and Warnings to our Errbit logger

### DIFF
--- a/src/Factories/ErrbitLoggerFactory.php
+++ b/src/Factories/ErrbitLoggerFactory.php
@@ -10,6 +10,31 @@ use Monolog\Logger;
 use Psr\Log\LogLevel;
 
 class ErrbitLoggerFactory {
+
+	/**
+	 * Data from PayPal IPN we want to have in ErrBit for faster debugging.
+	 *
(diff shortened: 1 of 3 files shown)


src/Factories/ErrbitLoggerFactory.php
```diff
@@ -10,6 +10,32 @@
...
 
 class ErrbitLoggerFactory {
+
+	/**
+	 * Data from PayPal IPN we don't want to have in ErrBit
+	 *
+	 * Should be in sync with PayPalBookingTransformer,
+	 * but since that is an internal class, we repeat the values here.
+	 */
+	private const FILTERED_PPL_PARAMS = [
```
Abban, line 20: I think this should be an allow list so we don't accidentally log personal data if PayPal add a new field or change a field name. Maybe overkill, but I think it's better to be safe than sorry.

Abban (changes requested):

src/Factories/ErrbitLoggerFactory.php
gbirke: Changed, please re-review

Abban (approved):

PR #2617, opened 2022-11-03
Add variable values to the intro_start template

moiikana:
the template call was missing the variables to fill the values in the fields

solves: https://phabricator.wikimedia.org/T322185


gbirke (approved):
User-Tested - works

PR #2619, opened 2022-11-05
Clean up old A/B testing campaigns

gbirke:
Move old campaigns to archived_campaigns.yml

Prepare address_type_steps for new test (see https://phabricator.wikimedia.org/T322415)

Ticket: https://phabricator.wikimedia.org/T298318

### DIFF
--- a/app/Controllers/Donation/NewDonationController.php
+++ b/app/Controllers/Donation/NewDonationController.php
@@ -38,7 +38,7 @@ class NewDonationController {
 				$interval,
 				$validationResult->isSuccessful(),
 				$trackingInfo,
-				$request->get( 'addressType', $ffFactory->getAddressType() ),
+				$request->get( 'addressType' ),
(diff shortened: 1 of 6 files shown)


Abban (approved):

app/Controllers/Donation/NewDonationController.php
```diff
@@ -38,7 +38,7 @@ public function index( FunFunFactory $ffFactory, Request $request ): Response {
 				$interval,
 				$validationResult->isSuccessful(),
 				$trackingInfo,
-				$request->get( 'addressType', $ffFactory->getAddressType() ),
+				$request->get( 'addressType', 'person' ),
```
moiikana, line 41: shouldn't this be null?

moiikana (changes requested):

app/Controllers/Donation/NewDonationController.php
moiikana: fixed by abban's commit

moiikana (approved):

PR #2620, opened 2022-11-11
Improve city lookup

gbirke:


src/Autocomplete/UseCases/FindCitiesUseCase.php
```diff
@@ -20,6 +20,7 @@ public function __construct( LocationRepository $locationRepository ) {
 	 * @return string[]
 	 */
 	public function getCitiesForPostcode( string $postcode ): array {
-		return $this->locationRepository->getCitiesForPostcode( $postcode );
+		$alphanumericPostCode = preg_replace( '/[^0-9a-zA-Z ]/', '', trim( $postcode ) );
```
moiikana, line 23: if you're cleaning this for our usecase, you could also clean away all the letters and only keep numbers, 
because we only look up cities for Germany (postcodes only consist of numbers)

moiikana (approved):

PR #2622, opened 2022-11-23
Avoid PayPal error-retry loops for booked payments

gbirke:
Check booking response if the failure reason is that the payment was
already booked. We still log the error (as a warning), but no longer
return an error response, because PayPal would try to book the
already-booked payment again several times.

Check success response for empty body in the tests.

Ticket:  https://phabricator.wikimedia.org/T321346

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
 		"wmde/email-address": "~1.0",
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
-		"wmde/fundraising-donations": "~11.0",
+		"wmde/fundraising-donations": "dev-special-booking-failure-message",
 		"wmde/fundraising-memberships": "~10.0",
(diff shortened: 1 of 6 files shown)


gbirke:
We need to update `composer.json` and run `make update-php COMPOSER_FLAGS="-W wmde/fundraising-donations` when https://github.com/wmde/fundraising-donations/pull/204 has been merged and we have a new (minor) donation bounded context release

Abban (approved):

composer.json
```diff
@@ -16,7 +16,7 @@
 		"wmde/email-address": "~1.0",
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
-		"wmde/fundraising-donations": "~11.0",
+		"wmde/fundraising-donations": "dev-special-booking-failure-message",
```
Abban, line 19: Don't forget to fix this before merging

PR #2624, opened 2022-12-01
Check if anonymous users are forbidden to use bank transfer

gbirke:
Ticket: https://phabricator.wikimedia.org/T325436


gbirke:
When https://github.com/wmde/fundraising-donations/pull/205 is merged, create a now minor Donation release and update the dependency in this branch. that should fix the failing integration test.

gbirke:
When https://github.com/wmde/fundraising-donations/pull/207 is merged, create a new release of the bounded context and use it in the `composer.json` instead of the `dev` branch. Then run `make update-php COMPOSER_FLAGS="wmde/fundraising-donations"` to update composer.lock.

Abban (approved):

PR #2625, opened 2022-12-01
Update the micropayment service client

gbirke:
The vendor has pulled the 1.25 binary, so we need to use the 1.26 version when installing dependencies.


gbirke:
We no longer use this library, I'll create a different PR

gbirke:
Replaced by #2626

PR #2632, opened 2022-12-27
Use docker compose command

gbirke:
New docker versions support a standalone "compose" subcommand instead of using the Python docker-compose script.


gbirke:
The docker version used by Travis does not support `docker compose` yet

PR #2636, opened 2023-01-03
Allow donation address to be updated

Abban:
Also removes deprecated newsletter and receipt opt-ins.

Ticket (address update): https://phabricator.wikimedia.org/T316632
Ticket (is exported): https://phabricator.wikimedia.org/T326246


moiikana (approved):

gbirke (changes requested):
We are striving to remove the test templates (a leftover from the old PHP-webapp-centered days), see https://phabricator.wikimedia.org/T163452

Please take the opportunity to change the tests and delete the template files.

moiikana:
there is also a test missing here that needs to be fixed, related to that:
UpdateDonorRouteTest::testWhenDonationAlreadyHasAddress_requestIsDenied
the assertion needs to assume false

Abban:
> there is also a test missing here that needs to be fixed

Huh, I deleted that already and it came back.

Abban:
Oh I see, it is deleted. We no longer need that test though, as we don't have that functionality there's no branch to test anymore.

gbirke (approved):

skins/test/templates/Donation_Confirmation.html.twig
```diff
@@ -22,7 +22,7 @@
 	</p>
 {% endif %}
 
-{% if donation.optsIntoNewsletter == true %}
+{% if donation.newsletter == true %}
```
gbirke, line 25: You can delete this file now as well

PR #2637, opened 2023-01-05
Create new AB test for reduced membership intervals

moiikana:
https://phabricator.wikimedia.org/T326300


src/Factories/ChoiceFactory.php
```diff
@@ -15,14 +15,19 @@ class ChoiceFactory {
 
 	/**
 	 * @var FeatureToggle
```
Abban, line 17: This entire docbloc can be deleted

Abban (approved):

PR #2639, opened 2023-01-09
Add new mailer for updated donations

Abban:
This uses the new donations template with a new
subject line for updated donation mails.

As donations can now be updated multiple times we
no longer want to send admins a moderation email
each time one is changed.

Ticket: https://phabricator.wikimedia.org/T317641


gbirke (approved):

gbirke:
This will work when rebased on current main

PR #2647, opened 2023-01-10
Add donation update API endpoint

Abban:


tests/EdgeToEdge/APIRoutes/UpdateDonorControllerTest.php
```diff
@@ -0,0 +1,222 @@
...
+			$client,
+			$this->newPrivateDonorData(),
+			$donation->getId(),
+			self::CORRECT_UPDATE_TOKEN,
+			self::INVALID_UPDATE_TOKEN
+		);
+
+		$response = $client->getResponse();
+
+		$this->assertSame( Response::HTTP_BAD_REQUEST, $response->getStatusCode() );
```
gbirke, line 109: Technically, this should be a "403 - Access Denied". But we don't gain much value from checking this specific error result from the use case, so I'm inclined to approve this.

gbirke (approved):

tests/EdgeToEdge/APIRoutes/UpdateDonorControllerTest.php
Abban: Also, the front end probably won't ever care about the response other than if they're 200 or not. I think maybe 400 as a general code to cover all failure outcomes is acceptable?

gbirke (approved):

PR #2754, opened 2023-05-31
Remove WMFG

Abban:
* Replace sub organization email templates with main organization
* Remove sub organization mailer from FunFunFactory
* Update tests
* Change contact-info config item

Ticket: https://phabricator.wikimedia.org/T332140

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -119,7 +119,7 @@
 
 		"wmde/psr-log-test-doubles": "~3.0",
 		"wmde/fundraising-phpcs": "~8.0",
-		"wmde/fundraising-frontend-content": "dev-test",
+		"wmde/fundraising-frontend-content": "dev-remove-wmfg",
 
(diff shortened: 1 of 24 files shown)


Abban:
The content update is here https://github.com/wmde/fundraising-frontend-content/pull/222 and needs to be merged, and the dependency updated in thie PR before merging it.

The config item in the infrastructure needs to be updated before deploying.

moiikana (approved):

composer.json
```diff
@@ -111,15 +111,15 @@
 	"require-dev": {
 		"ext-pdo_sqlite": "*",
 
-		"matthiasnoback/symfony-config-test": "dev-phpunit-10",
+		"matthiasnoback/symfony-config-test": "~4.3.0",
```
gbirke, line 114: Before merging this, please change to 5.0 because the 4.3.x release is not compatible with PHPUnit 10

PR #2780, opened 2023-06-29
Pin the Doctrine ORM version

Abban:
The current version of the Doctrine ORM 2.15
is throwing an error while generating Doctrine
classes.

This is a temporary fix until a new patched
version is released.


Abban:
This didn't work so closing

PR #2799, opened 2023-07-27
Allow null intervals in initialFormValues

Abban:
Interval was previously always being converted
with intval before being output to the front
end. Because it always had a value it caused
the interval in local storage on the form to
be overwritten on page refresh.

This change allows it to be null.

Ticket: https://phabricator.wikimedia.org/T342547

### DIFF
--- a/app/Controllers/Donation/NewDonationController.php
+++ b/app/Controllers/Donation/NewDonationController.php
@@ -9,6 +9,7 @@ use Symfony\Component\HttpFoundation\Response;
 use WMDE\Fundraising\Frontend\App\Routes;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
 use WMDE\Fundraising\Frontend\Presentation\Presenters\DonationFormPresenter\ImpressionCounts;
+use WMDE\Fundraising\PaymentContext\Domain\Model\PaymentInterval;
 
 class NewDonationController {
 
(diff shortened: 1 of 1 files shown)


app/Controllers/Donation/NewDonationController.php
```diff
@@ -19,12 +19,9 @@ public function index( FunFunFactory $ffFactory, Request $request ): Response {
...
 		$amount = intval( $request->get( 'amount', 0 ) );
 		$paymentType = (string)$request->get( 'paymentType', '' );
-		$interval = $request->get( 'interval', 0 );
-		if ( $interval !== null ) {
-			$interval = intval( $interval );
-		}
+		$interval = $this->getIntervalFromRequest( $request );
 
-		$validationResult = $ffFactory->newPaymentValidator()->validatePaymentData( $amount, $interval, $paymentType, $ffFactory->newDonationPaymentValidator() );
+		$validationResult = $ffFactory->newPaymentValidator()->validatePaymentData( $amount, $interval ?? 0, $paymentType, $ffFactory->newDonationPaymentValidator() );
```
moiikana, line 24: while reviewing I was wondering what that 0 here stands for as a fallback value.
Do we express with this that we want "one time payment" as a fallback user selection?

If that's correct, maybe we could give that 0 a name to not have it in there as a magic number.

gbirke (approved):
Thank you for writing an explanation in the commit message!

PR #2825, opened 2023-08-23
Add Authentication classes and fix broken feature branch

gbirke:
This commit implements the authentication interfaces for the bounded contexts, with persistence. I called the implementation `OldStyleTokens` to leave the option to have a different implementation that uses only tokens.

This also removes the reliance on the Doctrine event system for generating tokens. All the code in the Context Factories in the bounded contexts can be removed.

I accidentally force-pushed to the bounded context branches, which led to errors during the rebase of the `payment-with-ppl` branch. I removed the composer.lock during the rebase and I'm adding it back here. This commit fixes the broken `payment-with-ppl` branch.

Tickets:
https://phabricator.wikimedia.org/T341795
https://phabricator.wikimedia.org/T344271


src/Authentication/OldStyleTokens/PersistentAuthorizer.php
```diff
@@ -0,0 +1,60 @@
...
+	public function authorizeMembershipAccess( int $membershipId ): URLAuthenticator {
+		$token = $this->repository->getTokenById( $membershipId, AuthenticationBoundedContext::Membership );
+		if ( $token instanceof NullToken ) {
+			$token = $this->createNewToken( $membershipId, AuthenticationBoundedContext::Membership );
+		}
+
+		return new AccessTokenUrlAuthenticator( $token );
+	}
+
+	private function createNewToken( int $id, AuthenticationBoundedContext $context ): AuthenticationToken {
```
gbirke, line 46: should be `createAndStore`

app/Controllers/Donation/AddDonationController.php
```diff
@@ -49,13 +46,11 @@ public function index( FunFunFactory $ffFactory, Request $request ): Response {
...
-				[
-					'id' => $responseModel->getDonation()->getId(),
-					'accessToken' => $responseModel->getAccessToken()
-				]
-			) );
+			$authenticationLoader = $this->ffFactory->getUrlAuthenticationLoader();
+			$authenticator = $authenticationLoader->getDonationUrlAuthenticator( $responseModel->getDonation()->getId() );
+			$url = $this->ffFactory->getUrlGenerator()->generateAbsoluteUrl( Routes::SHOW_DONATION_CONFIRMATION );
+			$url = $authenticator->addAuthenticationTokensToApplicationUrl( $url );
+			return new RedirectResponse( $url );
```
gbirke, line 53: This logic and branch should be moved into the Payment bounded context so we'll always redirect.

PR #2826, opened 2023-08-23
Handle Notifications from PayPal API

gbirke:
https://phabricator.wikimedia.org/T329162


gbirke:
Closing this, because it contains just some TODOs for the next steps and then we realized we need to do all the authentication stuff.

PR #2840, opened 2023-09-12
Donation form test 01

Abban:


composer.json
```diff
@@ -16,7 +16,7 @@
 		"wmde/email-address": "~1.0",
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
-		"wmde/fundraising-donations": "~14.0",
+		"wmde/fundraising-donations": "dev-test",
```
gbirke, line 19: Interesting, shouldn't that be `dev-company-with-contact`?

app/Controllers/Donation/AddDonationController.php
```diff
@@ -63,6 +63,10 @@ private function newHttpResponse( AddDonationResponse $responseModel ): Response
 	private function createDonationRequest( Request $request ): AddDonationRequest {
 		$donationRequest = new AddDonationRequest();
 
+		if ( $request->get( 'addressType', '' ) === 'company_with_contact' ) {
+			$request->query->set( 'addressType', AddressType::LEGACY_COMPANY );
+		}
+
```
gbirke, line 69: This hack should go into `getSafeDonorType` IMHO, with a comment or maybe even with a restructuring of that code so we know what's going on.

app/Controllers/Validation/ValidateAddressController.php
```diff
@@ -24,11 +24,16 @@ class ValidateAddressController {
...
+		$isCompanyWithContact = $this->handleCompanyWithContact( $request );
+
 		$addressType = $this->getAddressType( $request );
 		if ( $addressType === AddressType::PERSON ) {
 			$nameViolations = $this->getPersonViolations( $request );
 		} elseif ( $addressType === AddressType::COMPANY ) {
 			$nameViolations = $this->getCompanyViolations( $request );
+			if ( $isCompanyWithContact ) {
+				$nameViolations = array_merge( $nameViolations, $this->getPersonViolations( $request ) );
+			}
```
gbirke, line 36: This check should go into `getCompanyViolations` to avoid 2 levels of nesting here

gbirke (changes requested):
If we quickly need to deploy this to make the test work, you can consider this as approved. But we then should address my comments as a follow-up

gbirke (approved):

moiikana:
needs to point to fundraising-donations v14.1

moiikana (approved):

PR #2869, opened 2023-10-25
Make the new design default

Abban:


moiikana:
before this is merged:
check if all current & incoming banners explicitly state which design they should use and not rely on the default here

moiikana (approved):
all the current banners have the design explicitly specified :white_check_mark:

PR #2871, opened 2023-10-26
Fix PayPal Adapter feature flag

gbirke:
Make sure we don't contact the PayPal API at all when the feature flag for using the legacy adapter is set.

Rename feature flag for more clarity.

Add proper canary check to test


gbirke:
This is ready for review & works, I just put it as draft because we should create a new patch release in payment before merging this PR

moiikana (approved):

PR #2879, opened 2023-11-03
Improve Sofort notification behavior

gbirke:
Don't return a "400 - Bad Request" message when a payment has already been booked. Instead, just log the incident. Otherwise, Sofort will retry 5 times.

Improve SofortPaymentNotificationRouteTest by removing the deprecated createEnvironment method. This leads to a lot of indentation changes. The only *real* change in the test is the test case `testGivenAlreadyConfirmedPayment_requestDataIsLogged`, where we now expect a "Ok" reponse instead of an error response.

### DIFF
--- a/app/Controllers/Payment/SofortNotificationController.php
+++ b/app/Controllers/Payment/SofortNotificationController.php
@@ -46,6 +46,10 @@ class SofortNotificationController {
 			return new Response( 'Ok', Response::HTTP_OK );
 		}
 
+		if ( $response->paymentWasAlreadyCompleted() ) {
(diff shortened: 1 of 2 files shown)


Abban (approved):
Should we add a test to check that an already completed payment will return 200/Ok?

gbirke:
> Should we add a test to check that an already completed payment will return 200/Ok?

That's already tested in `testGivenAlreadyConfirmedPayment_requestDataIsLogged()`, which contains the only *real* change in the test file. I looked at the "arrange" and "act" sections  of the tests and decided to test the return value in that test. I should probably change the test description

PR #2881, opened 2023-11-07
Update membership context

moiikana:
v13.0 -> v13.1
belongs to https://phabricator.wikimedia.org/T350378, still requires some frontend adjustments

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
 		"wmde/fundraising-donations": "~15.0",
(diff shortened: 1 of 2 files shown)


gbirke (approved):
Keep in mind that running `make update-php` will update *all* packages. You could have done `make update-php COMPOSER_FLAGS=wmde/fundraising-memberships` ton update memberships only. But in this case (during the campaign where our dev dependencies are not updated by dependabot) I think it's good you did a full update

Abban (approved):

gbirke:
I've created another membership fix in https://github.com/wmde/fundraising-memberships/pull/139
That could be released as v13.1.1

PR #2882, opened 2023-11-07
[DNM] Add command for fixing failed Sofort payments

gbirke:
This is temporary and not for merging, just for deploying it once.


gbirke:
Closing this, Sofort payments have been fixed with the command. We don't need the command any more.

PR #2890, opened 2023-11-28
Update fun-validators to v4.1.0

moiikana:
contains the changes for https://phabricator.wikimedia.org/T350676


gbirke:
This fails because the fun-validators version restriction is too tight in two repos. To make this PR pass, we need to merge the following two PRs and create patch releases in the repos, then update the dependencies in this PR again
- https://github.com/wmde/fundraising-memberships/pull/140
- https://github.com/wmde/fundraising-payments/pull/150

PR #2891, opened 2023-11-30
Add campaign for new membership banner visibility

gbirke:
Add an "A/B test" campaign" (really a feature toggle) that triggers
different behavior of suppressing fundraising banners on the wikipedia
projects during the fundraising and thank-you campaigns

Ticket: https://phabricator.wikimedia.org/T349192


gbirke:
Closing this, as the underlying cookie mechanism doesn't work with modern browsers' privacy protection and the ticket has to be reworked

PR #2892, opened 2023-12-05
Improve naming in application configuration

gbirke:
While working on https://phabricator.wikimedia.org/T350595 we discovered
some configuration keys that don't adhere to our language standards.
This commit makes it possible to transition existing configurations to a
better language, while keeping the configuration schema backwards compatible

Ticket: https://phabricator.wikimedia.org/T352788

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -881,17 +881,26 @@ class FunFunFactory implements LoggerAwareInterface {
 			new SimpleFileFetcher(),
 			$this->getLogger()
 		);
-		$textPolicyConfig = $this->config['text-policies'][$policyName];
 		return new TextPolicyValidator(
 			new WordListFileReader(
 				$fetcher,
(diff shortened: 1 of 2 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -2161,4 +2170,18 @@ private function newMembershipAuthorizer(): MembershipAuthorizer {
...
 
+	/**
+	 * @return string[]
+	 */
+	private function getEmailAddressBlockList(): array {
+		// Backwards compatibility check for legacy config key
+		// Remove this check (and maybe inline this method) once
+		// we have changed all configuration files (also in the infrastructure repo)
+		// see https://phabricator.wikimedia.org/T352788
+		if ( isset( $this->config['email-address-blacklist'] ) ) {
```
Abban, line 2181: Should this be the other way around? Eg we check for the blocklist and if it doesn't exist fall back to the old one.

Abban (approved):

PR #2894, opened 2023-12-18
Improve base template

gbirke:
Display a message when the frontend assets come from a server. This
improves the developer experience, so we can distinguish between "empty
page when frontend code has errors" and "page when frontend code server
is not running"


moiikana (approved):

gbirke:
I've noticed the text "flashing" when switching pages and the assets are loaded (e.g. submitting a donation), which is a bit unnerving. I'd rather add a bit of styling to the message that makes it transparent and shows it after 1 second

PR #2901, opened 2024-01-10
Pass campaign params to Frontend

gbirke:
Rename "Renderer" to "PropertyExtractor"

Ticket: https://phabricator.wikimedia.org/T350935

### DIFF
--- a/src/Presentation/CampaignPropertyExtractor.php
+++ b/src/Presentation/CampaignPropertyExtractor.php
@@ -1,0 +1,22 @@
+<?php
+declare( strict_types=1 );
+
+namespace WMDE\Fundraising\Frontend\Presentation;
+
+use WMDE\Fundraising\Frontend\BucketTesting\Domain\Model\Campaign;
+
+class CampaignPropertyExtractor {
+
+	/**
+	 * @return string[]
+	 */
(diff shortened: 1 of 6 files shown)


src/Presentation/CampaignPropertyExtractor.php
```diff
@@ -0,0 +1,22 @@
...
+
+class CampaignPropertyExtractor {
+
+	/**
+	 * @return string[]
+	 */
+	public static function listURLKeys( Campaign ...$campaigns ): array {
+		return array_map(
+			static function ( Campaign $c ) {
+				return $c->getUrlKey();
```
Abban, line 16: Should we be filtering these by `$c->isOnlyActiveWithUrlKey()`?

Abban (approved):

src/Presentation/CampaignPropertyExtractor.php
gbirke: That would make the feature more "correct", yes. But let's save the CPU cycles of another `if` statement for now, since for the last 3 years we haven't used any A/B test campaigns without requiring an URL key.

Co and me also started with an additional check if the campaign is active at all, but then decided against it, with the reasoning that it's nice for developers to see the URL parameters of all the campaigns in the URL.

PR #2902, opened 2024-01-10
Use docker compose 2.0

gbirke:
Change Travis file and Makefile to use `docker compose` instead of `docker-compose` (the latter is a python script, the former is a Docker subcommand).


gbirke:
Closing for now, will try again when we've migrated away from Travis

PR #2906, opened 2024-01-22
Handle mailing list Subscription on donor update

Sperling-0:
- Pass mailing list parameter from API request to use case
- Delete old non-API use case

Ticket: https://phabricator.wikimedia.org/T344425


Abban (approved):

composer.json
```diff
@@ -16,7 +16,7 @@
 		"wmde/email-address": "~1.0",
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
-		"wmde/fundraising-donations": "~15.0",
+		"wmde/fundraising-donations": "dev-preselect-notification-opt-in",
```
gbirke, line 19: You can now set this to `~15.2`, I've merged our changes in the donation repo. Then You can regenerate the composer.lock file. If CI goes through, you can then merge this, Abban has accepted this PR

PR #2918, opened 2024-02-05
Clean up translations passed to frontend

gbirke:
The contents of the files `membershipTypes.json` and `validations.json`
are not used on the frontend, so we can remove them.

Ticket: https://phabricator.wikimedia.org/T355124


moiikana (changes requested):
at least membershipTypes are used on the confirmation summary

moiikana:
while checking if those keys are maybe used in the mail templates I noticed that they are used on the confirmation page
Bildschirmfoto vom 2024-02-05 19-20-26

moiikana:
closing because this needs another investigation on what can be cleaned up and what not

PR #2923, opened 2024-02-15
Update DonorType usage

Abban:
DonorType is now a proper php enum so use that.

Ticket: https://phabricator.wikimedia.org/T339080

### DIFF
--- a/app/Controllers/Validation/ValidateAddressController.php
+++ b/app/Controllers/Validation/ValidateAddressController.php
@@ -7,6 +7,7 @@ namespace WMDE\Fundraising\Frontend\App\Controllers\Validation;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use WMDE\Fundraising\DonationContext\Domain\Model\DonorType;
 use WMDE\Fundraising\Frontend\Factories\FunFunFactory;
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
 		"wmde/email-address": "~1.0",
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
-		"wmde/fundraising-donations": "~16.0",
+		"wmde/fundraising-donations": "dev-donortype-enum",
```
moiikana, line 19: needs new release of `fundraising-donations` when [the `fundraising-donations](https://github.com/wmde/fundraising-donations/pull/233)  is merged

app/Controllers/Validation/ValidateAddressController.php
```diff
@@ -23,31 +24,38 @@ class ValidateAddressController {
...
+				$nameViolations = $this->getPersonViolations( $request );
+				break;
+			case DonorType::COMPANY:
+				$nameViolations = $this->getCompanyViolations( $request );
+				if ( $isCompanyWithContact ) {
+					$nameViolations = array_merge( $nameViolations, $this->getPersonViolations( $request ) );
+				}
+				break;
+			case DonorType::ANONYMOUS:
+				return $this->newSuccessResponse();
```
moiikana, line 56: as there is no default case here, I would suggest adding a `break` here as well, to be future proof against missing it accidentally, when adding new cases.

moiikana (approved):

PR #2925, opened 2024-02-22
Upgrade phpstan level

Sperling-0:
Upgrade phpstan level to 9


composer.json
```diff
@@ -115,7 +115,7 @@
 		"matthiasnoback/symfony-config-test": "~5.0",
 		"mikey179/vfsstream": "~1.6",
 		"phpmd/phpmd": "~2.10",
-		"phpstan/phpstan": "^1.8",
+		"phpstan/phpstan": "^1.10",
```
gbirke, line 118:
```suggestion
		"phpstan/phpstan": "^1.11",
```

gbirke (changes requested):
Part of this PR should be making `strict-phpstan.neon` the new default `phpstan.neon`. Then the Drone CI errors should go away.

composer.json
Sperling-0: @gbirke This error was not there earlier. I think it occured because of `"phpstan/phpstan": "^1.11",` change.
What to do in this case?

Screenshot from 2024-06-18 08-42-47

gbirke: You introduced a change in the Makefile that causes this - have a look at the `stan` task and make sure that you only change the level an nothing else (except deleting the `strict-stan` task). Your current change is running the phpstan binary not from the composer-installed `vendor/bin/phpstan` but from an outdated Docker image.

I first suspected that the `phpstan-doctrine` dependency might be responsible (seeing that `DoctrineProxyForbiddenClassNamesExtension` class). I've tested our current version (1.3.69) with the correct docker invocation and found that that's not the case. Instead, looking at open and closed issues of the extension helped me to take another look at the Makefile, when I saw https://github.com/phpstan/phpstan-doctrine/issues/558

Nevertheless we should take the opportunity of this PR and update ouur 2 phpstan extension dependencies as well: changing the `composer.json` version requirement for `phpstan-doctrine` from the overly specific `~1.3.62` to a more generic `~1.4` and for `phpstan-phpunit` as well.

Makefile
```diff
@@ -131,10 +131,7 @@ fix-cs:
 	-docker-compose run --rm --no-deps app ./vendor/bin/phpcbf
 
 stan:
-	docker run --rm -it --volume $(BUILD_DIR):/app -w /app $(DOCKER_IMAGE):latest php -d memory_limit=1G vendor/bin/phpstan analyse --level=5 --no-progress cli/ src/ tests/
```
gbirke, line 134: This line should stay almost as-is, only with a changed level

gbirke (changes requested):

gbirke (approved):

PR #2926, opened 2024-02-28
Fix integration errors caused by new enums

moiikana:
- the introduction of AddressType and DonorType as native php enums cause some issues


Abban:
There's already a PR for the DonorType integration https://github.com/wmde/fundraising-application/pull/2923

PR #2927, opened 2024-03-01
Rebuild Drone cache before running tests

Abban:
Our CI fails randomly, with an error that files are missing. This looks like it's coming from a phpStan cleanup that's happening simultaneously with the cache rebuild step.

This commit moves the rebuild cache step between the build and test steps so it will rebuild the cache before Stan runs.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,12 +26,6 @@ steps:
       - echo -e "machine github.com\n  login $GITHUB_TOKEN" > ~/.netrc
       - composer config -g github-oauth.github.com "$GITHUB_TOKEN"
       - composer install --prefer-dist --no-progress --no-interaction
-      # CI
-      - vendor/bin/phpcs
-      - php -d xdebug.mode=coverage vendor/bin/phpunit --coverage-clover coverage.xml
-      - php -d memory_limit=1G vendor/bin/phpstan analyse --level=5 --no-progress src/ tests/
-      # Upload coverage to Scrutinizer
-      - ocular code-coverage:upload --no-interaction --format=php-clover coverage.xml
 
   - name: rebuild-cache
     image: drillster/drone-volume-cache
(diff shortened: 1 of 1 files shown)


gbirke (changes requested):
I have investigated this further in #2928 and came to a different conclusion for the source of this: I think it's not some leftover PHPStan files but rather 2 cache processes running in parallel: For new PRs, GitHub creates 2 CI runs - one for the PR and one for the branch. These run in parallel on our Drone server. That means the "cache rebuild" steps also run in parallel. The caching script uses `rsync` for copying the files: https://github.com/Drillster/drone-volume-cache/blob/master/cacher.sh#L49
The cache script runs `rsync` with the `--delete` flag to delete  files from the cache that are no longer in the source. `rsync` also creates temporary files while copying. Now if two rsync processes are running parallel, they will interfere with each other, because each one will delete the temporary files from the other one, which is IMO the reason why the `rebuild-cache` steps fails.

Originally I put in the cache step to preserve the files in the `vendor` directory, so composer only has to install them if the `composer.json` or `composer.lock` file is different between CI runs. I've checked the times for a uncached `composer install` for the Fundraising Application - it takes about 8 seconds. Which is less than the "restore cache" and "rebuild cache" steps take, so the vendor caching does not give us a speedup. I recommend we remove those steps from our CI pipelines. We **must** keep the `cache` volume and mounting the composer cache volume in the CI step. This will ensure that composer will not have to download the dependencies from packagist or GitHub for every CI run.


Another thing I discovered while investigating this: It's purely accidental that the `drillster/drone-volume-cache` works for caching vendor for a repo. You can see that the cache key is also the current Drone job number: https://github.com/Drillster/drone-volume-cache/blob/master/cacher.sh#L14 But `DRONE_JOB_NUMBER` does not exist, only `DRONE_BUILD_NUMBER`, see https://docs.drone.io/pipeline/environment/reference/
That means by default the build caches should be isolated, but it also means that sharing a cache across builds with the plugin can be dangerous. Which in my opinion is another good argument for removing the `vendor` caching and relying solely on a shared composer cache.

Abban:
Closing this, I'll open a new one that removes the caching altogether

PR #2929, opened 2024-03-07
Implement new Donation and Membership objects with removed FVO

Abban:
This updates the creation and access of the Donation and Membership objects that had the Freezable Value Object trait removed from them.

Ticket: https://phabricator.wikimedia.org/T359090


gbirke:
To use memberships version 15 (with removed FVO and improved types), you need to cherry-pick the commits from #2936 into this

PR #2932, opened 2024-03-12
Take care of potential `false` values

Sperling-0:
    
### DIFF
--- a/src/BucketTesting/Validation/CampaignUtilizationValidator.php
+++ b/src/BucketTesting/Validation/CampaignUtilizationValidator.php
@@ -94,17 +94,29 @@ class CampaignUtilizationValidator {
 		$campaignBuckets = [];
 		foreach ( array_count_values( $this->featureToggles ) as $featureToggle => $featureToggleCount ) {
 			$campaign = $this->getCampaignNameFromFeatureToggle( $featureToggle );
+			if ( $campaign === '' ) {
+				$this->errorLogger->addError(
+					'Feature toggle name did not contain a campaign name. The campaign name should be the last ' .
+					'item in a dot-separated string (with at least one "." character). Feature toggle name was: ' .
+					$featureToggle
+				);
+				continue;
+			}
 			if ( isset( $campaignBuckets[$campaign] ) && $campaignBuckets[$campaign] !== $featureToggleCount ) {
 				$this->errorLogger->addError(
 					'Campaign buckets for "' . $featureToggle . '" have not been implemented consistently.'
 				);
 				continue;
 			}
-			$campaignBuckets[$this->getCampaignNameFromFeatureToggle( $featureToggle )] = $featureToggleCount;
+			$campaignBuckets[$campaign] = $featureToggleCount;
 		}
 	}
 
 	private function getCampaignNameFromFeatureToggle( string $featureToggle ): string {
-		return substr( $featureToggle, 0, strrpos( $featureToggle, '.' ) );
+		$lastSeparatorPosition = strrpos( $featureToggle, '.' );
+		if ( $lastSeparatorPosition === false ) {
+			return '';
+		}
+		return substr( $featureToggle, 0, $lastSeparatorPosition );
 	}
 }
--- a/src/Factories/WebTemplatingFactory.php
+++ b/src/Factories/WebTemplatingFactory.php
@@ -39,7 +39,8 @@ class WebTemplatingFactory extends TwigFactory {
 			new TwigFunction(
 				'translations',
 				function (): string {
-					return json_encode( $this->translations );
+					// "We know the translations come from JSON so we can let the JSON Exception be unchecked because that should never happen
+					return json_encode( $this->translations, JSON_THROW_ON_ERROR );
 				},
 				[ 'is_safe' => [ 'html' ] ]
 			),
--- a/src/Infrastructure/Mail/MailTemplateFilenameTraversable.php
+++ b/src/Infrastructure/Mail/MailTemplateFilenameTraversable.php
@@ -14,7 +14,18 @@ class MailTemplateFilenameTraversable implements \IteratorAggregate {
 	}
 
 	public function getIterator(): \Iterator {
-		foreach ( glob( $this->mailTemplatePath . '/*\.twig' ) as $fileName ) {
+		$glob = glob( $this->mailTemplatePath . '/*\.twig' );
+
+		// we can't reliably trigger glob() returning `false` on Linux systems
+		// @codeCoverageIgnoreStart
+		if ( $glob === false ) {
+			throw new \RuntimeException( sprintf( "Failed to find path name: %s",
+				var_export( $this->mailTemplatePath . '/*\.twig', true ) ) );
+		}
+
+		// @codeCoverageIgnoreEnd
+
+		foreach ( $glob as $fileName ) {
 			yield basename( $fileName );
 		}
 	}
--- a/src/Infrastructure/UserDataKeyGenerator.php
+++ b/src/Infrastructure/UserDataKeyGenerator.php
@@ -8,7 +8,7 @@ use WMDE\Clock\Clock;
 
 class UserDataKeyGenerator {
 
-	private const START_TIME = '2020-01-01 0:00:00Z';
+	private const SECONDS_PER_DAY = 86400;
 
 	private string $masterKey;
 	private Clock $time;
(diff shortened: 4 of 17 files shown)


src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -105,6 +105,11 @@ private function findInconsistentFeatureToggleOccurrences(): void {
 	}
 
 	private function getCampaignNameFromFeatureToggle( string $featureToggle ): string {
-		return substr( $featureToggle, 0, strrpos( $featureToggle, '.' ) );
+		$strpos = strrpos( $featureToggle, '.' );
+		if ( $strpos === false ) {
+			throw new \RuntimeException( 'Failed to find the position of the last occurrence of "." 
+			in the string: ' . $featureToggle );
```
gbirke, line 111:
```suggestion
			throw new \RuntimeException( 'Feature toggle name did not contain a campaign name. The campaign name should be the last item in a dot-separated string (with at least one "." character). Feature toggle name was: ' . $featureToggle );
```

src/Factories/WebTemplatingFactory.php
```diff
@@ -39,7 +39,12 @@ function ( string $name, array $context = [] ): string {
 			new TwigFunction(
 				'translations',
 				function (): string {
-					return json_encode( $this->translations );
+					$result = json_encode( $this->translations );
+					if ( $result === false ) {
+						throw new \RuntimeException( sprintf( "Failed to get JSON representation of: %s",
+							var_export( $this->translations, true ) ) );
+					}
+					return $result;
```
gbirke, line 47: This is a good example of an instance where using arrays is hurting us: `$translations` should not be an array (of `array<string,string>` type by the way), but an object that can take of its (de-)serialization. Because these strings come from actual JSON files and if those can be decoded, they can be encoded again. Testing this Twig function have a lot of setup overhead. I recommend we have a look at this together and see how we can refactor translation collection and serialization. 

If we want to merge this PR quickly without extending the scope (refactoring translations into a class), we could do one of two things:

1)  add a "TODO" in here, e.g. "TODO: Convert $ranslations array to a class that can serialize itself to JSON. then we don't have to check for false for PHPSTan's sake".
2) Use `json_encode($this->translations, JSON_THROW_ON_ERROR)` with the comment "We know the translations come from JSON so we can let the JSON Exception be unchecked because that should never happen"

gbirke: I've now done a followup-PR instead in #2933

src/Infrastructure/UserDataKeyGenerator.php
```diff
@@ -20,7 +20,12 @@ public function __construct( string $masterKey, Clock $time ) {
 
 	public function getDailyKey(): string {
 		$now = $this->time->now();
-		$daysSince2020 = abs( $now->diff( new \DateTimeImmutable( self::START_TIME ) )->days );
+		$days = $now->diff( new \DateTimeImmutable( self::START_TIME ) )->days;
+		if ( $days === false ) {
+			throw new \RuntimeException( sprintf( "Failed to get the total number of days in the interval span: %s",
```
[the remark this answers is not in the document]
Sperling-0: @gbirke My IDE and phpstan stays `diff` doesn't always return a difference! I have attached the screenshots for your reference. Let me know what to do in this case.
       
Screenshot from 2024-03-13 20-27-42

      

Screenshot from 2024-03-13 20-31-55

tests/EdgeToEdge/Routes/CreditCardPaymentNotificationRouteTest.php
```diff
@@ -46,8 +46,13 @@ public function testGivenInvalidRequest_applicationIndicatesError(): void {
 			);
 
 			$this->assertSame( 1, $logger->getLogCalls()->count() );
-			$this->assertStringContainsString( "status=error\n", $client->getResponse()->getContent() );
-			$this->assertStringContainsString( 'msg=', $client->getResponse()->getContent() );
+			$content = $client->getResponse()->getContent();
+			if ( $content === false ) {
+				throw new \RuntimeException( sprintf( "Failed to get the content of: %s",
```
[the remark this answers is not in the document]
Sperling-0: @gbirke I have removed the `if` statement. And the initial errors came back. (I have attached a screenshot for your reference.) But now I have added `Parameter #2 $haystack of method PHPUnit\Framework\Assert::assertStringContainsString() expects string, string|false given.` error message in the `phpstan-baseline.neon` file. So, now the errors are skipped successfully.

       
       
Screenshot from 2024-03-13 21-15-41

tests/Unit/Infrastructure/ConfigReaderTest.php
```diff
@@ -38,10 +38,15 @@ public function setUp(): void {
 	}
 
 	private function getDistConfigContents(): string {
-		return json_encode(
+		$result = json_encode(
 			$this->getDistConfig(),
 			JSON_PRETTY_PRINT
 		);
+		if ( $result === false ) {
+			throw new \RuntimeException( sprintf( "Failed to get JSON representation of: %s",
```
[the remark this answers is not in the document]
Sperling-0: @gbirke I added a baseline rule for this error as well. But doing that skipped another _unrelated_ error in this file!

tests/EdgeToEdge/Routes/CreditCardPaymentNotificationRouteTest.php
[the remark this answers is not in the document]
Sperling-0: Look for the correct match in the generated source neon file.

src/Factories/WebTemplatingFactory.php
Sperling-0: Done.

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -105,6 +105,12 @@ private function findInconsistentFeatureToggleOccurrences(): void {
 	}
 
 	private function getCampaignNameFromFeatureToggle( string $featureToggle ): string {
-		return substr( $featureToggle, 0, strrpos( $featureToggle, '.' ) );
+		$strpos = strrpos( $featureToggle, '.' );
+		if ( $strpos === false ) {
```
Sperling-0, line 109: @gbirke Should I re-throw the Exception as we are doing in the case of JsonException?

src/Infrastructure/Mail/MailTemplateFilenameTraversable.php
```diff
@@ -14,7 +14,12 @@ public function __construct( private readonly string $mailTemplatePath ) {
 	}
 
 	public function getIterator(): \Iterator {
-		foreach ( glob( $this->mailTemplatePath . '/*\.twig' ) as $fileName ) {
+		$glob = glob( $this->mailTemplatePath . '/*\.twig' );
+		if ( $glob === false ) {
```
Sperling-0, line 18: @gbirke Should I re-throw the Exception as we are doing in the case of JsonException?

src/BucketTesting/Validation/CampaignUtilizationValidator.php
gbirke: `strrpos` does not throw an exception like `json_encode`: https://www.php.net/manual/en/function.strrpos
Therefore *have* to check for false instead of catching and re-throwing.

src/BucketTesting/Validation/CampaignUtilizationValidator.php
```diff
@@ -105,6 +105,12 @@ private function findInconsistentFeatureToggleOccurrences(): void {
 	}
 
 	private function getCampaignNameFromFeatureToggle( string $featureToggle ): string {
-		return substr( $featureToggle, 0, strrpos( $featureToggle, '.' ) );
+		$strpos = strrpos( $featureToggle, '.' );
```
gbirke, line 108: Naming the variable `strpos` when the function is `strrpos` confused me for a moment (there is also a `strpos` function). Let's give this variable a more semantic name like `$indexOfRightmostDot` or `$lastSeparatorPosition`

src/Infrastructure/Mail/MailTemplateFilenameTraversable.php
gbirke: Unlike the JSON functions, `glob` does not have a flag that changes its behavior from returning an error result vs throwing an exception: https://www.php.net/manual/en/function.glob
So we have to check for false here.

gbirke (commented):
Don't forget to write test cases for the `if` statements added to the classes in `src/`. And for the JsonBucketLogger.

gbirke (approved):

PR #2936, opened 2024-03-14
Adapt to new membership mailer

gbirke:
The new confirmation mailer of the membership bounded context now passes
class-based parameters, breaking the "unified" `sendMail` interface that
exists across bounded contexts. This change splits the mailer into two
parts - the TemplateMailerInterface that all Twig-based mailers and
their decorators implement and an adapter, that converts the membership
mailer into the TemplateMailerInterface.

This is a "template" for possible changes to the mailers in all bounded
context - they will probably also switch from arrays to value objects,
getting their own adapters.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~1.0",
 		"wmde/fundraising-donations": "~17.0",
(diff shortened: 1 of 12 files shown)


gbirke:
Closing, needed to merge this with the changes in #2929

PR #2953, opened 2024-04-16
Add page titles to the messages

Abban:
This will allow static pages a way to access them in order to set the correct html title on page load.

Also removes unneeded privacy protection template.

Ticket: https://phabricator.wikimedia.org/T362631


gbirke (approved):
The more individual JSON files we include for messages, the more potential for unwanted overrides and duplicate keys we have. But that's something that needs to be solved on the content side not in this PR.
The `pageTitles.json` contains some very short keys, I would have wished them to have prefixes so they're more easily recognizable as page titles. But that would trigger a lot of changes, both in the backend and the frontend, so I'm fine with this

PR #2954, opened 2024-04-23
Add individual page titles to the templates

Abban:
Adds the page titles into the page templates to remove the title content flash that was happening on page load.

It does this by adding a page title function to Twig.

Ticket: https://phabricator.wikimedia.org/T362630

### DIFF
--- a/skins/laika/templates/page_layouts/base.html.twig
+++ b/skins/laika/templates/page_layouts/base.html.twig
@@ -3,6 +3,8 @@
 {% set title = site_metadata.page_titles[page_id] %}
 {% set content = web_content("pages/#{page_id}") %}
 
+{% block title %}{$ page_title( page_id ) $}{% endblock %}
+
 {% block main %}
 	<div id="app">
 		<noscript>
--- a/skins/laika/templates/page_layouts/supporters.html.twig
+++ b/skins/laika/templates/page_layouts/supporters.html.twig
@@ -1,6 +1,8 @@
 {% extends 'Base_Layout.html.twig' %}
 {% set title = site_metadata.page_titles[page_id] %}
 
+{% block title %}{$ page_title( page_id ) $}{% endblock %}
+
 {% block scripts %}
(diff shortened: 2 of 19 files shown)


gbirke (approved):

skins/laika/templates/page_layouts/supporters.html.twig
```diff
@@ -1,6 +1,8 @@
 {% extends 'Base_Layout.html.twig' %}
 {% set title = site_metadata.page_titles[page_id] %}
 
+{% block title %}{$ page_title( page_id ) $}{% endblock %}
```
Sperling-0, line 4: Maybe it should be?
```javascript
{% block title %}{$ page_title( supporters_page_title ) $}{% endblock %}
```

skins/laika/templates/page_layouts/base.html.twig
```diff
@@ -3,6 +3,8 @@
 {% set title = site_metadata.page_titles[page_id] %}
 {% set content = web_content("pages/#{page_id}") %}
 
+{% block title %}{$ page_title( page_id ) $}{% endblock %}
```
Sperling-0, line 6:
```javascript
{% block title %}{$ page_title( page_id ) $}{% endblock %}
```

In this line the translation key is `page_id` because the file is `base.html.twig`?

Abban: Base is used for the static pages, and page_id is the key that the individual page translation is stored under in the content repo.

If you look at the change I made in the `WebTemplatingFactory.php` file you can see I defined a new function to pull out the title based on that key.

skins/laika/templates/page_layouts/supporters.html.twig
Abban: I think the language key is `hall-of-fame`, but it works fine as it is, and it's consistent with the other static pages so lets leave it as is.

PR #2957, opened 2024-05-17
Add modal target element to base layout

Abban:
Modals need to be displayed outside the main vue app
element, so that when opened the app can be marked as
inert to prevent the donor tabbing to it.

Ticket: https://phabricator.wikimedia.org/T361818


Abban:
Don't need this now modals are native HTML Dialog elements

PR #2958, opened 2024-05-23
Update to Symfony 7.0

gbirke:
Update dependencies

Fix BC-breaking changes:
- Cache Warmer interface
- Different behavior of ParameterBag::getInt when key is present but
  empty. This now triggers an exception. Changed the test because it
  does not test faulty, but missing parameters.


src/Infrastructure/SkinCacheWarmer.php
```diff
@@ -18,12 +18,7 @@ public function isOptional(): bool {
 		return false;
 	}
 
-	/**
-	 * @param string $cacheDir
-	 *
-	 * @return string[]
-	 */
-	public function warmUp( string $cacheDir ): array {
+	public function warmUp( string $cacheDir, ?string $buildDir = null ): array {
```
moiikana, line 21: should have a string[] return type in the docs again

moiikana (approved):

PR #2967, opened 2024-06-07
Take care of `cannot access offset on mixed` errors

Sperling-0:

### DIFF
--- a/src/BucketTesting/CampaignConfigurationLoader.php
+++ b/src/BucketTesting/CampaignConfigurationLoader.php
@@ -26,8 +26,12 @@ class CampaignConfigurationLoader implements CampaignConfigurationLoaderInterfac
 	public function loadCampaignConfiguration( string ...$configFiles ): array {
 		$cacheKey = $this->getCacheKey( ...$configFiles );
 		if ( $cacheKey !== '' && $this->cache->has( $cacheKey ) ) {
-			return $this->cache->get( $cacheKey )['campaigns'];
+			$cache = $this->cache->get( $cacheKey, [] );
+			if ( is_array( $cache ) && isset( $cache[ 'campaigns' ] ) ) {
+				return $cache[ 'campaigns' ];
+			}
 		}
+
 		$configs = $this->loadFiles( ...$configFiles );
 
 		if ( count( $configs ) === 0 ) {
--- a/src/Factories/TwigFactory.php
+++ b/src/Factories/TwigFactory.php
@@ -31,7 +31,10 @@ abstract class TwigFactory {
 	}
 
 	private function getLoader(): LoaderInterface {
-		if ( !empty( $this->config['loaders']['filesystem'] ) ) {
+		if (
+			is_array( $this->config['loaders'] ) &&
+			isset( $this->config['loaders']['filesystem'] )
+		) {
 			return new FilesystemLoader( $this->config['loaders']['filesystem'] );
 		}
 		throw new \UnexpectedValueException( 'Invalid Twig loader configuration - missing filesystem' );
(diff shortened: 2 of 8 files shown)


src/BucketTesting/CampaignConfigurationLoader.php
```diff
@@ -25,7 +25,7 @@ public function __construct(
 	 */
 	public function loadCampaignConfiguration( string ...$configFiles ): array {
 		$cacheKey = $this->getCacheKey( ...$configFiles );
-		if ( $cacheKey !== '' && $this->cache->has( $cacheKey ) ) {
+		if ( $cacheKey !== '' && $this->cache->has( $cacheKey ) && is_array( $this->cache->get( $cacheKey ) ) ) {
```
Abban, line 28: I think this would work and be a little more readable:

```php
$cache = $this->cache->get( $cacheKey, [] );
if ( isset( $cache[ 'campaigns' ] ) ) {
	return $cache[ 'campaigns' ];
}
```

src/Factories/TwigFactory.php
```diff
@@ -31,7 +31,10 @@ public function __construct(
 	}
 
 	private function getLoader(): LoaderInterface {
-		if ( !empty( $this->config['loaders']['filesystem'] ) ) {
+		if (
+			is_array( $this->config['loaders'] ) &&
+			isset( $this->config['loaders']['filesystem'] )
+		) {
```
Abban, line 37: You should be able to evaluate using only isset here as it'll fail if any of the levels are missing.

```php
if ( isset( $this->config[ 'loaders' ][ 'filesystem' ] ) ) {
```

src/Infrastructure/Mail/DonationConfirmationMailSubjectRenderer.php
```diff
@@ -20,8 +20,12 @@ public function __construct(
 	 * @param array<string, mixed> $templateArguments
 	 */
 	public function render( array $templateArguments = [] ): string {
-		if ( $templateArguments['donation']['paymentType'] === PaymentType::BankTransfer->value ) {
-			return $this->translator->trans( $this->bankTransferSubjectKey );
+		if (
+			is_array( $templateArguments['donation'] ) &&
```
Abban, line 24: You can probably also drop the first check from here

Abban (approved):

src/BucketTesting/CampaignConfigurationLoader.php
Sperling-0: @Abban I did the changes that you suggested but then the initial error came back:

  Line   src/BucketTesting/CampaignConfigurationLoader.php  
 ------ --------------------------------------------------- 
  30     Cannot access offset 'campaigns' on mixed.         
         🪪  offsetAccess.nonOffsetAccessible  
         
         
What should be the next step then?

src/Factories/TwigFactory.php
Sperling-0: If I do as you suggest then the initial error comes back:

 ------ --------------------------------------------- 
  Line   src/Factories/TwigFactory.php                
 ------ --------------------------------------------- 
  35     Cannot access offset 'filesystem' on mixed.  
         🪪  offsetAccess.nonOffsetAccessible         
 ------ --------------------------------------------- 
 
 What should be the next step then?

src/Infrastructure/Mail/DonationConfirmationMailSubjectRenderer.php
Sperling-0: @Abban If I do as you suggest then the initial error comes back:

 ------ --------------------------------------------------------------------- 
  Line   src/Infrastructure/Mail/DonationConfirmationMailSubjectRenderer.php  
 ------ --------------------------------------------------------------------- 
  24     Cannot access offset 'paymentType' on mixed.                         
         🪪  offsetAccess.nonOffsetAccessible                                 
 ------ ---------------------------------------------------------------------
 
 What should be the next step then?

gbirke (approved):

PR #2969, opened 2024-06-11
Take care of `offset` errors

Sperling-0:

### DIFF
--- a/src/Factories/FunFunFactory.php
+++ b/src/Factories/FunFunFactory.php
@@ -453,6 +453,11 @@ class FunFunFactory implements LoggerAwareInterface {
 		return ( new SimpleFileFetcher() )->fetchFile( $this->getI18nDirectory() . '/data/supporters.json' );
 	}
 
+	private function getLanguageFromLocale(): string {
+		$locale = Locale::getPrimaryLanguage( $this->getLocale() );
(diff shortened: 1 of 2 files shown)


src/Factories/FunFunFactory.php
```diff
@@ -466,6 +466,10 @@ public function getSkinTwig(): Environment {
 				$packageFactory->newAssetPackages()
 			);
 			$locale = Locale::parseLocale( $this->getLocale() );
+			if ( $locale === null ) {
+				$locale = [ 'language' => 'de' ];
+			}
+
```
gbirke, line 472: You could DRY this code up a bit by putting the lookup into a `getLanguage` function that encapsulates the check and replace the `$locale['language']` access with `$this->getLanguageFromLocale()`.

Inside the new method you could also use  `Locale::getPrimayLanguage()` instead of `Locale::parseLocale()`, see https://www.php.net/manual/en/locale.getprimarylanguage.php

gbirke (approved):

src/Factories/FunFunFactory.php
Sperling-0: I did the changes. I hope that is what you meant.

gbirke (approved):
Perfect, thanks!

PR #2971, opened 2024-06-12
Take care of 'expectation' issues

Sperling-0:


src/Authentication/OldStyleTokens/AccessTokenUrlAuthenticator.php
```diff
@@ -26,7 +26,7 @@ public function addAuthenticationTokensToApplicationUrl( string $url ): string {
 	 * @param class-string $urlGeneratorClass
 	 * @param string[] $requestedParameters
 	 *
-	 * @return array<string, int|string>
+	 * @return array<string, bool|int|string>
```
gbirke, line 29: The boolean comes from the `json_encode` and is confusing/misleading. Please add `JSON_THROW_ON_ERROR` instead to ensure `custom` is a always string

src/Infrastructure/Translation/TranslatorInterface.php
```diff
@@ -8,7 +8,7 @@ interface TranslatorInterface {
 
 	/**
 	 * @param string $messageKey
-	 * @param array<string, scalar> $parameters
+	 * @param array<string, string|float|null> $parameters
```
gbirke, line 11:
```suggestion
	 * @param array<string, scalar|null> $parameters
```

If it was scalar before, I'd rather not risk an error when passing an integer or boolean.

src/Validation/FieldTextPolicyValidator.php
```diff
@@ -15,7 +15,7 @@ class FieldTextPolicyValidator {
 	public function __construct( private readonly TextPolicyValidator $textPolicyValidator ) {
 	}
 
-	public function validate( $value ): ValidationResult {	// @codingStandardsIgnoreLine
+	public function validate( string $value ): ValidationResult {	// @codingStandardsIgnoreLine
```
gbirke, line 18: This class is entirely unused outside of its test. Please delete it and its test.

gbirke (approved):

src/Infrastructure/Translation/TranslatorInterface.php
Sperling-0: @gbirke PHPStan was giving me an error on this line, to fix it I made the changes.

If I change it back to `scalar` the error will come back, what to do in that case??

Sperling-0: I have made the changes suggested by you.

PR #2974, opened 2024-06-18
Make updateAddress URL point to API route

Abban:
Now address changes will be made through an API call we
need to pass that route to the fronted.


gbirke (approved):
We could start prefixing the API routes with API_ to distinguish them from web application routes. But that’s entirely optional

PR #2976, opened 2024-06-27
Add third address form test option

Abban:
We'll actually have 3 variants of the address page running
at the same time for this test as the mobile banner will
use the legacy form, and the desktop test will test 2 variants
of the new form.

Ticket: https://phabricator.wikimedia.org/T366576


Abban:
Closing this because it turns out we actually only need 2 variants

PR #2992, opened 2024-09-12
Add API controller for donation comments

gbirke:
The new frontend code now sends a JSON body, so we can take the
opportunity to modernize the API.

Use `#[MapRequestPayload]` attribute for JSON serialization and type
checking.

Reused most of the controller and test code of the old controller, which
can be deleted when the frontend is merged.

Updated all dependencies when adding new dependency
`symfony/serializer-pack` .

Ticket: https://phabricator.wikimedia.org/T367387

### DIFF
--- a/app/Controllers/API/Donation/AddCommentController.php
+++ b/app/Controllers/API/Donation/AddCommentController.php
@@ -1,0 +1,36 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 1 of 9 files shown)


Abban (approved):
Looks good, we might try adding the mapping to the other API endpoints too at some point. In those we do `$request->toArray()` which I guess handles getting the data out of the json body.

PR #2994, opened 2024-10-07
Document Database Setup and Migrations

gbirke:
How to integrate a database change from a bounded context in the Fundraising Application

### DIFF
--- a/doc/Database_Setup_and_Migrations.md
+++ b/doc/Database_Setup_and_Migrations.md
@@ -7,7 +7,7 @@ The Fundraising Application has two different database environments:
 
 The two environments have different strategies for initial setup and migration of existing database structure and data:
 
-- The local environment is a "throwaway" environment - When one developer on the team changes the database structure, they notify the rest of the team that they should throw away their docker container and docker volumes and restart the development environment. The mechanism we use is putting **SQL files** with structure and data into the directory `.docker/database/`. The `docker-compose.yml` file defines a database volume that we can throw away. When the docker container starts and has no volume from previous runs, MariaDB will execute the files (in alphabetical order).
+- The local environment is a "throwaway" environment - When one developer on the team changes the database structure, they notify the rest of the team that they should throw away their docker container and docker volumes and restart the development environment. The mechanism we use is putting **SQL files** with structure and data into the directory `.docker/database/`. The `docker-compose.yml` file defines a database volume that we can throw away. When the docker container starts and has no volume from previous runs, MariaDB will execute the SQL files (in alphabetical order).
 - The server environment needs to preserve the existing data (all donations since 2011, etc.), so we can't throw the database away when we want to make changes to the structure. Therefore we use **Database Migration Scripts** that modify the database structure (and copy or modify the existing data if it needs adaptation to the new structure).
 
 
--- a/doc/Database_Setup_and_Migrations.md
+++ b/doc/Database_Setup_and_Migrations.md
@@ -61,7 +61,7 @@ You should see a list of migrations, none them applied. In the local environment
 To test / run the new migration, run the following command:
 
 ```shell
-docker-compose exec app vendor/bin/doctrine-migrations migrations:migrations:execute 'Your\\Migration\\Name'
+docker compose exec app vendor/bin/doctrine-migrations migrations:migrations:execute 'Your\\Migration\\Name'
 ```
 
 Replace the placeholder `Your\\Migration\\Name` with the namespace name of the actual migration, e.g. `'WMDE\\Fundraising\\MembershipContext\\DataAccess\\Migrations\\Version20240528122513'`. 
--- a/doc/Database_Setup_and_Migrations.md
+++ b/doc/Database_Setup_and_Migrations.md
@@ -1,0 +1,92 @@
+# How the database setup and migration works
+
+The Fundraising Application has two different database environments:
+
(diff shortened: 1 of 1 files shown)


doc/Database_Setup_and_Migrations.md
```diff
@@ -0,0 +1,92 @@
...
+- The **server environment** has a dedicated database server with two databases (test and production) that our test and production web servers connect to. The Fundraising App and the Fundraising Operation Center both share the same production or testing database.
+
+The two environments have different strategies for initial setup and migration of existing database structure and data:
+
+- The local environment is a "throwaway" environment - When one developer on the team changes the database structure, they notify the rest of the team that they should throw away their docker container and docker volumes and restart the development environment. The mechanism we use is putting **SQL files** with structure and data into the directory `.docker/database/`. The `docker-compose.yml` file defines a database volume that we can throw away. When the docker container starts and has no volume from previous runs, MariaDB will execute the files (in alphabetical order).
```
moiikana, line 10:
```suggestion
- The local environment is a "throwaway" environment - When one developer on the team changes the database structure, they notify the rest of the team that they should throw away their docker container and docker volumes and restart the development environment. The mechanism we use is putting **SQL files** with structure and data into the directory `.docker/database/`. The `docker-compose.yml` file defines a database volume that we can throw away. When the docker container starts and has no volume from previous runs, MariaDB will execute the SQL files (in alphabetical order).
```
I think adding "SQL" at the end might help understanding it while reading.

doc/Database_Setup_and_Migrations.md
```diff
@@ -0,0 +1,92 @@
+# How the database setup and migration works
+
+The Fundraising Application has two different database environments:
+
+- The **local development environment**. We're running a MariaDB database in a Docker container for local user and acceptance testing. We are using an in-memory SQLite database for running unit tests. This makes running the tests fast and allows for a clear separation between interactive testing and automated testing. 
+- The **server environment** has a dedicated database server with two databases (test and production) that our test and production web servers connect to. The Fundraising App and the Fundraising Operation Center both share the same production or testing database.
```
moiikana, line 6:
```suggestion
- The **server environment** is running on a dedicated database server with two databases (test and production) that our test and production web servers connect to. The Fundraising App and the Fundraising Operation Center both share the same production or testing database.
```

doc/Database_Setup_and_Migrations.md
```diff
@@ -0,0 +1,92 @@
...
+```shell
+docker compose exec app php bin/doctrine migrations:list
+```
+
+You should see a list of migrations, none them applied. In the local environment this is fine, because we throw away the Doctrine migration metadata table whenever we regenerate the local environment. If the migration does not show up, check the migrations directory in `vendor/wmde/your-bounded-context` and the configuration file `app/config/migrations.php`.
+
+To test / run the new migration, run the following command:
+
+```shell
+docker-compose exec app vendor/bin/doctrine-migrations migrations:migrations:execute 'Your\\Migration\\Name'
```
moiikana, line 64:
```suggestion
docker compose exec app vendor/bin/doctrine-migrations migrations:migrations:execute 'Your\\Migration\\Name'
```

doc/Database_Setup_and_Migrations.md
```diff
@@ -0,0 +1,92 @@
...
+Run the command `docker compose down -v`. This will stop all containers and will also clear out their data volumes. 
+
+Afterwards, run `docker compose up -d` to start the environment again.
+
+### Troubleshooting
+
+If you're getting errors from the application after the containers are up, check if the database tables exist by running the following command:
+
+```shell
+docker compose exec database mysql -u fundraising -p"INSECURE PASSWORD" fundraising -e "SHOW TABLES"
```
moiikana, line 25:
```suggestion
docker compose exec database mysql -u fundraising -p "INSECURE PASSWORD" fundraising -e "SHOW TABLES"
```

doc/Database_Setup_and_Migrations.md
```diff
@@ -0,0 +1,92 @@
...
+
+## How to integrate a database change in the local environment
+
+To integrate a database change from a bounded context you need to do the following steps:
+
+1. Create an integration branch in in the application repository, pointing the version of the bounded context in `composer.json` to the `-dev` branch of the bounded context.
+2. Run CI and fix failing tests.
+3. Run the database migration on the *old* version of the database schema
+4. User-test the changed database (to check if the migrated database is correct)
+5. Re-generate SQL for the local environment.
```
moiikana, line 39: does this point refer to `### Regenerating the local SQL` ? Maybe we could link there or add the command here as well

moiikana (changes requested):
see comments above for suggestions/change requests

doc/Database_Setup_and_Migrations.md
gbirke: Nope, this is one of the pecularities of the MySQL client: if you specify `-p something` it will interpret that as "ask for a password and use parameter 'something'". If its `-psomething` it will interpret that as "Use password 'something'". And sometimes will give a warning about that practice because that command can then end up in the bash history, revealing the password.

doc/Database_Setup_and_Migrations.md
gbirke: Hmm, maybe "consists of" or just "uses"? Pinging @Abban for a comment

Abban: "Consists of" is good

Abban (approved):

PR #3008, opened 2024-12-05
Use new Use of Funds content

Abban:
Ticket: https://phabricator.wikimedia.org/T381411


gbirke (approved):
This PR contains a lot of dependency updates - Symfony from 7.1 to 7.1, a major update of PHPStan to 2.0

If everybody did proper semver, nothing will break, but maybe you could do `make update-content` instead of `make update-php`

PR #3009, opened 2024-12-18
Pass receipt option to front end

Abban:
Banners can now pass a receipt option that need to be handled in the front end, this passes it through from the request.

Ticket: https://phabricator.wikimedia.org/T382304


gbirke (approved):
The long parameter list of `DonationFormPresenter` is quite "smelly" even before the addition of yet another parameter, but I don't see good ways to make it better quickly, so let's continue making it bad until it becomes unbearable ...

PR #3014, opened 2025-01-06
Re-enable dependabot 2025

Sperling-0:
- Update dependencies before re-enabling dependabot


gbirke:
I had a look at the PHPStan issues - they can be solved by providing the type of `TRequest` and `TResponse` for the `AbstractBrowser` class.

Start with the `WebRouteTestCase` and change the comment like this:

```
* @param AbstractBrowser<Request,Response> $client
```

Make sure to (auto)import the right request and response classes, both should be in the `Symfony\Component\HttpFoundation\` namespace. If you forget to import (`use`) those classes or use the wrong ones, PHPStan will complain (with a different error message).

The remaining errors are all in the APIRoute tests where we alias `AbstractBrowser` to `Client`, so the comment would look like this:

```
* @param Client<Request,Response> $client
```

If you want to read more about the underlying mechanism (docblock-level generics), see https://phpstan.org/blog/generics-in-php-using-phpdocs and https://arnaud.le-blanc.net/post/phpstan-generics.html

gbirke (approved):

PR #3026, opened 2025-01-31
Extend the address_pages AB test end date

moiikana:
- the `test_02` bucket (ap=1) is still used in the last banners from 2024, so we still want to be able to navigate to the different buckets with an URL parameter. For that the end date had to be adjusted.

https://phabricator.wikimedia.org/T385562

### DIFF
--- a/app/config/campaigns.yml
+++ b/app/config/campaigns.yml
@@ -26,7 +26,7 @@ campaigns:
     address_type_choice:
       description: Test offering an "anonymous" choice (coming from a banner where the user donated less than 10 EUR)
       reference: "https://phabricator.wikimedia.org/T380462"
-      start: "2024-11-25" 
+      start: "2024-11-25"
       end: "2024-12-31"
       buckets:
         - "default"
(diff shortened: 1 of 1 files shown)


app/config/campaigns.yml
```diff
@@ -53,7 +53,7 @@ campaigns:
       description: Test different iterations of address pages
       reference: "https://phabricator.wikimedia.org/T366576"
       start: "2024-07-01"
-      end: "2024-12-31"
+      end: "2030-12-31"
```
gbirke, line 56: Giving it another 5 years makes this more of a feature than an A/B test. I'd like to have a discussion with the PM if there are really "tests" going on around this, that might conclude at some point (where we will remove the `<FeatureToggle>` tags and the duplicated address forms from the frontend) or if they consider the 2 address forms now a "feature" of the fundraising application. In the latter case we'd need a bit of refactoring to make the fundraising application act on the `ap` parameter without the A/B test campaigns. This is to avoid "feature toggle creep" and give the feature explicitly named code for better readability and visibility instead of having to guess and look up what the feature toogle ID means and getting pointed to 2023/2024 banner tickets

gbirke (changes requested):

moiikana:
> I'd like to have a discussion with the PM 

I created a ticket: ['d like to have a discussion with the PM ](https://phabricator.wikimedia.org/T385562#10520469)

PR #3028, opened 2025-01-31
Remove scrutinizer config

gbirke:
We don't want to run Scrutinizer any more


moiikana:
there now is a ticket for removing scrutinizer https://phabricator.wikimedia.org/T385554

moiikana (approved):

PR #3046, opened 2025-03-07
Use the new memberships version

Abban:
Ticket: https://phabricator.wikimedia.org/T316201


Sperling-0 (approved):

moiikana:
released https://github.com/wmde/fundraising-memberships/releases/tag/v17.1.0

gbirke (approved):
I'd be inclined to initialize the anonymizers inside the commands instead of the factory, because they're so infrequently used and the factory is loaded on every request. But that's just a suggestion.

PR #3077, opened 2025-04-23
Add donation number test to contact validation

Abban:
The donation number is optional but if passed is
required to be numeric.

This adds a test to ensure that functionality exists.

Ticket: https://phabricator.wikimedia.org/T392447


gbirke (changes requested):
Thanks for adding this test! You could add another one that proves (and documents) that an empty donation number is valid.

gbirke (approved):

PR #3096, opened 2025-05-12
Clean up confirmation mail template

moiikana:
- confirmation mails should only be sent when there is no moderation status set for the donation --> that's why the else section can be removed
- Since no e-mail should be sent for high donation amounts, the code that changes the sender name and address in the e-mail to the donor for confirming high amount donations can be removed

https://phabricator.wikimedia.org/T390840


gbirke (approved):

moiikana:
this PR requires https://github.com/wmde/fundraising-infrastructure/pull/452 to be merged

PR #3116, opened 2025-06-24
Remove Sofort from config

Abban:
We will no longer be taking Sofort payments so this removes
it as an option on the front end.

Ticket: https://phabricator.wikimedia.org/T397707


gbirke:
You can delete the failing test case

PR #3134, opened 2025-08-31
Remove Matomo fallback tracking

gbirke:
Remove link to "tracking image" that gets loaded when user has disabled
JavaScript.

This change is a shot in the dark to try and make an PCI ASV test pass. If the
test still fails (because we load the tracker with JavaScript), revert
this change.

This change is an "emergency fix" to meet a deadline as fast as
possible. It's not coordinated with the campaigns team and should be
reverted if they object to it.


gbirke:
We found a way to allow Matomo tracking as "External Link" in the ASV scan

PR #3227, opened 2026-06-02
Add controller and factory wiring in application context

Sperling-0:
Add controller and factory wiring in application context

- Add controller in - app/Controllers/API/Membership folder
- Add a new route in - config/api-routes.yaml

Ticket: https://phabricator.wikimedia.org/T427636

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~2.0",
 		"wmde/fundraising-donations": "~29.0",
-		"wmde/fundraising-memberships": "~21.0",
+		"wmde/fundraising-memberships": "~21.1",
 		"wmde/fundraising-payments": "~8.4",
 		"wmde/fundraising-subscriptions": "6.0",
 		"wmde/fundraising-content-provider": "~6.4",
--- a/app/Controllers/API/Membership/UpdateMembershipApplicationController.php
+++ b/app/Controllers/API/Membership/UpdateMembershipApplicationController.php
@@ -1,0 +1,64 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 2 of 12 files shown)


gbirke (changes requested):
Code looks good, but route needs and Edge2Edge test

Abban (changes requested):
~~You need to hook up the `MembershipUpdatedEvent` in the `CreateAddressChangeHandler` so it can update the Address Change attached to the membership. You can see how it does it for the`DonationUpdatedEvent`.~~

Edit: We're deleting the Address Change now so you don't need to worry about this.

Abban (changes requested):
The Edge to Edge test is still missing. I think you may have overwritten it in a force push. I found it [here](https://github.com/wmde/fundraising-application/compare/08074d42a884e3e6e7640eeeee451fbb874da419...d25ce0a225d5086be536302cd52f72eeae026d9d). I think you need to go through everything and make sure you haven't lost other work. [Here's all of your force pushes](https://github.com/wmde/fundraising-application/activity?ref=update-address)

Also just to note, We [have a rule](https://github.com/wmde/fundraising-cheatsheets/blob/main/Cheatsheets/Git.md) that the subject of a commit shouldn't be longer than 50 characters or so, because we want to make sure that it doesn't get truncated when displayed on the PR page in GitHub.

Abban (approved):

composer.json
```diff
@@ -17,7 +17,7 @@
 		"wmde/euro": "~1.0",
 		"wmde/clock": "~2.0",
 		"wmde/fundraising-donations": "~29.0",
-		"wmde/fundraising-memberships": "~21.0",
+		"wmde/fundraising-memberships": "dev-main",
```
moiikana: please create a new release of the main branch in the fundraising-membership repo and use that version here

PR #3241, opened 2026-06-26
Add subscriptions bounded context to migrations tooling

moiikana:
- in order to generate doctrine migrations we need to add the subscriptions dependency as a context here
- we need to add migrations to properly change or recreate the database structure


Abban (approved):

gbirke (approved):

moiikana:
contains a new migration! 
needs to be run when deploying

